PR #557, opened 2021-06-03
Stepan/tg-138

stepan662:


webapp/src/component/navigation/NavigationWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Box
+      style={{
+        borderBottom: `1px solid ${grey[200]}`,
+        backgroundColor: grey[50],
+      }}
+      pl={4}
+      pr={4}
+    >
+      <div className={classes.container}>{children}</div>
```
JanCizmar, line 27: Do we really need Box and div?

webapp/src/component/navigation/NavigationWrapper.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import grey from '@material-ui/core/colors/grey';
+import { Box, makeStyles } from '@material-ui/core';
+
+const useStyles = makeStyles({
+  container: {
```
JanCizmar, line 6: display, alignItems could be provided as props of Box component

webapp/src/component/navigation/NavigationWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+    height: 48,
+    display: 'flex',
+    alignItems: 'center',
+    '& > * + *': {
+      marginLeft: 10,
```
JanCizmar, line 11: Do you really need exact 10px? Component Box has "ml" or "marginLeft" prop or you can use theme.spacing function.

webapp/src/component/layout/DashboardPage.tsx
```diff
@@ -1,90 +1,40 @@
...
 
 export const DashboardPage: FunctionComponent<DashboardPageProps> = ({
   children,
-  sideMenuItems,
   ...props
 }) => {
   const classes = useStyles({});
 
   return (
     <div className={classes.root}>
```
JanCizmar, line 34: Using <Box> you don't have to use that many style props.

webapp/src/component/navigation/NavigationWrapper.tsx
stepan662: Fixed

webapp/src/component/navigation/NavigationWrapper.tsx
stepan662: Fixed

PR #584, opened 2021-06-21
feat(MenuIcons): change to custom

stepan662:


JanCizmar (changes requested):
Make every Icon as a separate component

JanCizmar (approved):

PR #597, opened 2021-06-24
feat(InvitationLink): fix code remembering and add tests

stepan662:


webapp/package.json
```diff
@@ -38,7 +38,7 @@
   "scripts": {
     "start": "DISABLE_ESLINT_PLUGIN=true craco start",
     "startConcurrently": "concurrently \"craco start\"",
-    "build": "craco build",
+    "build": "DISABLE_ESLINT_PLUGIN=true craco build",
```
JanCizmar, line 41: Is it possible to do so in .env? This wouldn't work on windows. Maybe we should add Husky to run eslint as precommit hook.

JanCizmar (changes requested):

webapp/package.json
stepan662: Ok, I've put it to .env.production

PR #746, opened 2021-08-20
Stepangranat/tg 201 optimize translations scrolling

stepan662:


e2e/cypress/common/translations.ts
```diff
@@ -10,28 +10,15 @@ import { ProjectDTO } from '../../../webapp/src/service/response.types';
 import Chainable = Cypress.Chainable;
 
 export function getCellEditButton(content: string) {
-  return cy
-    .contains(content)
-    .xpath(
-      "./ancestor::*[@data-cy='translations-table-cell']//*[@data-cy='translations-cell-edit-button']"
-    )
-    .invoke('show');
+  return cy.contains(content);
```
JanCizmar, line 13: I don't think, this would work for all possible cell contents.

webapp/src/service/http/ApiHttpService.tsx
```diff
@@ -78,7 +78,10 @@ export class ApiHttpService {
             );
             throw new Error('Error status code from server');
           }
-          if (r.status == 403) {
+          if (
+            r.status == 403 &&
+            (init?.method === undefined || init?.method === 'get')
```
JanCizmar, line 83: Why so?

webapp/src/service/http/ApiHttpService.tsx
```diff
@@ -87,7 +90,11 @@ export class ApiHttpService {
             );
             return;
           }
-          if (r.status == 404 && !options.disableNotFoundHandling) {
+          if (
+            r.status == 404 &&
+            !options.disableNotFoundHandling &&
+            (init?.method === undefined || init?.method === 'get')
```
JanCizmar, line 96: Why so?

JanCizmar (changes requested):

e2e/cypress/common/translations.ts
stepan662: Changed

webapp/src/service/http/ApiHttpService.tsx
stepan662: Changed

webapp/src/service/http/ApiHttpService.tsx
stepan662: Changed

PR #755, opened 2021-08-27
feat: Translations - comments TG-186

stepan662:


e2e/cypress/integration/translations/comments.ts
```diff
@@ -0,0 +1,104 @@
...
+} from '../../common/apiCalls';
+import {
+  commentsButton,
+  createComment,
+  deleteComment,
+} from '../../common/comments';
+import { enterProject } from '../../common/projects';
+import { waitForGlobalLoading } from '../../common/loading';
+
+describe('Translation states', () => {
```
JanCizmar, line 14: Translation states -> translation comments

webapp/src/views/projects/translations/comments/Comments.tsx
```diff
@@ -0,0 +1,306 @@
...
+    },
+    reverseScroller: {
+      display: 'flex',
+      flexDirection: 'column-reverse',
+      overflowY: 'auto',
+      overflowX: 'hidden',
+      flexBasis: 100,
+      flexGrow: 1,
+      overscrollBehavior: 'contain',
+      // scrollBehavior: 'smooth',
```
JanCizmar, line 47: Remove this line.

webapp/src/views/projects/translations/comments/Comments.tsx
```diff
@@ -0,0 +1,306 @@
...
+    },
+    loadMore: {
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'flex-end',
+      minHeight: 50,
+    },
+    bottomPanel: {
+      padding: '3px 5px 3px 12px',
```
JanCizmar, line 56: We should use global theming constants for padding and margin `theme.spacing()`. If it doesn't satisfy our needs, maybe we should consider different  value for the global spacing unit, which is now 8px.

JanCizmar (changes requested):

webapp/src/views/projects/translations/comments/Comments.tsx
stepan662: It ain't that simple.

stepan662: We need quite detailed spacing e.g. here: https://tolgee.myjetbrains.com/youtrack/issue/TGX-16

stepan662: Maybe we could set it to 4. But I'm afraid that it would require a lot of changes.

JanCizmar: In this specific case, it's file to fine-tune it. I just would like to be sure, we are following material-ui standards to avoid future problems.

JanCizmar (approved):

webapp/src/views/projects/translations/comments/Comments.tsx
stepan662: Ok

PR #771, opened 2021-09-09
feat: Translations counter and UP button TG-269

stepan662:


webapp/src/views/projects/translations/TranslationsHeader.tsx
```diff
@@ -93,72 +106,89 @@ export const TranslationsHeader = () => {
...
-        )}
+          {projectPermissions.satisfiesPermission(
+            ProjectPermissionType.EDIT
+          ) && (
+            <Button
+              startIcon={<Add />}
+              color="primary"
+              size="small"
+              variant="contained"
+              aria-label="add"
```
JanCizmar, line 170: Aria label should be translated.

webapp/src/views/projects/translations/TranslationsToolbar.tsx
```diff
@@ -0,0 +1,111 @@
...
+      })}
+    >
+      <div className={classes.index}>
+        <span data-cy="translations-toolbar-counter">{counterContent}</span>
+        {/* stretch content by monospace font, so it's not jumping */}
+        <div className={classes.stretcher}>{counterContent}</div>
+      </div>
+      <div className={classes.divider} />
+      <IconButton
+        data-cy="translations-toolbar-to-top"
```
JanCizmar, line 102: Aria-label and tooltip would be cool here.

JanCizmar (changes requested):

PR #788, opened 2021-09-22
fix: Hiding menu jumping TG-312

stepan662:


webapp/src/views/projects/translations/ColumnResizer.tsx
```diff
@@ -8,7 +8,7 @@ const useStyles = makeStyles({
     zIndex: 1,
     position: 'absolute',
     top: 0,
-    background: 'transparent',
+    // background: 'transparent',
```
JanCizmar, line 11: remove this line

JanCizmar (approved):

PR #792, opened 2021-09-23
feat: Concept of the help for keyboard shortcuts TG-297

stepan662:


webapp/src/views/projects/translations/ShortcutsHint.tsx
```diff
@@ -0,0 +1,143 @@
...
+  },
+  hoverHidden: {
+    opacity: 0,
+    transition: 'opacity 300ms ease-in-out',
+  },
+}));
+
+type Shortcut = {
+  name: React.ReactNode;
+  meta?: string;
```
JanCizmar, line 81: Maybe we can be more specific here like meta?: 'CTRL' | 'CMD'

webapp/src/views/projects/translations/TranslationOpened.tsx
```diff
@@ -166,6 +179,24 @@ export const TranslationOpened: React.FC<Props> = ({
               onCancel={onCancel}
               onStateChange={onStateChange}
             />
+            <ShortcutsHint
+              items={[
+                {
+                  name: <T>{translationStates[nextState].translationKey}</T>,
```
JanCizmar, line 185: `TypeError: Cannot read properties of undefined (reading 'translationKey')` is thrown when I try to edit untranslated value. Please add test for this.

JanCizmar (changes requested):
It looks nice, but It doesn't make the process much faster. 

I am still not able to review the translations with hitting single shortcut without moving between translations with mouse or hitting tab many times. I think, that combination of moving with arrows and allowing users to to focus the key items like key or specific translations without opening editor would be the best option. In the focused item or with hover (which can do both the same thing, actually) we can show the shortcuts the way Zuzka designed.

e2e/cypress/common/shortcuts.ts
```diff
@@ -0,0 +1,51 @@
+import { waitForGlobalLoading } from './loading';
+
+export const IS_MAC = Boolean(navigator.userAgent.includes('Mac OS X'));
```
JanCizmar, line 3: What about older Macs? ❓ 🥇

stepan662: Ok, switched to Mac

PR #900, opened 2022-01-28
Jancizmar/tg 492 update spring and other deps cleanup

JanCizmar:


docker/base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:14-jdk-alpine
+FROM openjdk:11-jdk-alpine
```
stepan662, line 1: Is this correct? Shouldn't we go up? 😄

stepan662 (approved):

docker/base/Dockerfile
JanCizmar: Right! This is mistake.

PR #924, opened 2022-02-10
fix: Sanitize api keys for OpenReplay, fix user identification

JanCizmar:


webapp/src/component/App.tsx
```diff
@@ -85,17 +89,22 @@ const MandatoryDataProvider = (props: any) => {
...
+    setOpenReplayTracker(window.openReplayTracker);
+  }, [config?.clientSentryDsn, config?.openReplayApiKey]);
 
   useEffect(() => {
-    if (userData && window.openReplayTracker) {
-      window.openReplayTracker.setUserID(userData.username);
+    if (userData && openReplayTracker) {
+      openReplayTracker.setUserID(userData.username);
+      setTimeout(() => {
+        openReplayTracker?.setUserID(userData.username);
```
stepan662, line 104: This is to make sure? 😄

stepan662 (approved):

webapp/src/component/App.tsx
JanCizmar: He said, I should do it twice. 😄  But I don't think that calling it 2 times right after should be enough so I did it this way. He said, they'll fix it.

PR #944, opened 2022-02-25
feat: Add comments resolution

stepan662:


JanCizmar (changes requested):
The button looks more like some kind of tag/inforamation. The numbers are inconsistent, there is different number in the tab name and different number in the icon. Perhaps the header should contain both numbers (reviewed/all). I also don't like the positioning of the "no comments dot" in the icon as well as the resolve button. This should be reviewed by Zuzka.

JanCizmar (approved):

PR #946, opened 2022-02-28
feat: Configurable export

stepan662:


stepan662:
Add e2e tests

JanCizmar (changes requested):
- In the State filter, there should be "All" option, selecting all states and there shouldn't be the "untranslated" option. "All" should be default.
- When there is nothing to export the backend throws 400 error message, which currently leads to infinite loading, which should never happen and it was covered before the migration to React Query.

JanCizmar (approved):

PR #985, opened 2022-03-15
feat: Translation history - Backend

### LINKED ISSUE #913: Translation history/activity
Enable users to see history of translations

JanCizmar:


stepan662 (approved):

backend/app/src/main/kotlin/io/tolgee/api/v2/hateoas/translations/TranslationHistoryModel.kt
```diff
@@ -0,0 +1,33 @@
...
+@Relation(collectionRelation = "revisions", itemRelation = "revision")
+open class TranslationHistoryModel(
+  @Schema(description = "Translation text")
+  val text: String?,
+
+  @Schema(description = "State of translation")
+  val state: TranslationState?,
+
+  @Schema(description = "Was translated using Translation Memory or Machine translation service?")
+  val auto: Boolean?,
```
stepan662, line 21: Isn't this too complicated, I thought that we will store history purely as just text changes.

JanCizmar: I think that information about state changes or auto translation dismissal would be also something users should know. Also from backend point of view, it's simple to return everything then just particular changes.

PR #995, opened 2022-03-19
Add german translation

eweren:
I've created the german translation automatically but went through every single key an corrected where necessary.


eweren:
Closed because of false branch naming

Issue #919, opened 2022-02-03
Translate permissions for specific language


JanCizmar:
Currently, we cannot simply enable users to translate specific language. So we need to provide special configuration option to specify languages when user has translate permission.

PR #996, opened 2022-03-20
feat: Add german translation

eweren:
I automatically translated all keys to german. I went through all of them and edited them where necessary.


JanCizmar:
Thanks a lot for this! ❤️  I added the german translation to our project in Tolgee.

It seems like we have few issues with our workflow. Since we are replacing the translations in the project with the translations from Tolgee while the project is built with GitHub actions, your data would be replaced. That's why it makes no sense to merge this PR.

Since we currently don't have #919 done. I cannot provide you access to our project in Tolgee. So I had to import it myself. 
It seems like there are some translations missing. Would you be so kind and translate the missing as well and provide the result? Then I will update the available languages myself and close this issue.

Hopefully, we will be done with the security fast enough and would be able to provide access for community translators like you to our project in Tolgee.

[Tolgee itself (17).zip](https://github.com/tolgee/server/files/8311637/Tolgee.itself.17.zip)

eweren:
> Would you be so kind and translate the missing as well and provide the result? Then I will update the available languages myself and close this issue.

Sure thing. There you go! Somehow the export had a completely different structure.

eweren:
For the workflow of yours: I set up a project of a client on friday with a script similar to the following. It grabs all the translation files of a tolgee project, moves them into the correct directory and if there were changes in translations, these changes will be commited and pushed. Since it's not that smart to do this on all branches everytime a commit happens, we came up with the idea to trigger this when there are commits on main. This ensures that there are no merge-conflicts. The locale files are hidden in our IDE, so nobody tries to edit those manually. Maybe this would be something four this project too.

Next great thing in this topic: Webhooks that gets triggered when a translation is "published" on tolgee. But I guess this is kind of bigger thing since there would need to be some  publish/unpublish logic as well as the webhook thin build from scratch from what I saw as well as more changes to permission-management. Maybe you already have something similar planned. If not: As soon as I get the setup to run correctly I will try to dive into spring and concept the webhooks topic.

```bash
#!/bin/bash

curl "$TOLGEEAPI_URL/api/project/export/jsonZip?ak=$TOLGEE_API_KEY" \
--output data.zip
unzip data.zip -o ./i18n
rm data.zip

if [[ `git status --porcelain` ]]; then
  echo "New translations found"
  git commit -am ":recycle: Update translations"
  git push
fi

```

JanCizmar:
@eweren Hey! We released the language based permissions, so now I can give you access to our project. Would you be open to check whether the translations are correct and add the missing ones, so we can release the german you provided?

eweren:
> Would you be open to check whether the translations are correct and add the missing ones, so we can release the german you provided?

@JanCizmar I am done going through all the german translations and added the missing ones. So german is done :)

JanCizmar:
Great! 🥇 Thanks a lot! Gonna release it tomorrow! 😍

JanCizmar:
Closing in favour of https://github.com/tolgee/tolgee-platform/pull/1061 

@eweren Do you have a twitter account? We would like to thank you in a tweet.

eweren:
@JanCizmar No I don't have a Twitter account. But I'm happy I could help 😊

PR #998, opened 2022-03-20
fix: Clean up dependencies

eweren:
I saw some dependencies that were located in the dependencies instead of the devDependencies.
Moving them saves just 6kB in build size, but they shouldn't have to be packed in production build.


stepan662:
Hey, thanks for the cleanup, it's definitely an improvement. I just want to point out that separation to "devDependencies" and "dependencies" is purely for esthetical reasons here as all the libraries will get installed anyway.
https://jsramblings.com/do-dependencies-devdependencies-matter-when-using-webpack/

stepan662 (approved):

PR #1003, opened 2022-03-21
feat: Upgrade material-ui -> mui

eweren:
I saw that this project used the material-ui lib. Since that one is deprecated for a while now, I upgraded to mui - it's successor. This results in some changes of how the components look like - mainly those inputs that were previously underlined instead of outlined. Feel free to reject or request changes so that this matches the old design.


JanCizmar:
Hi there, thanks for creating this pull request. 

This is actually a big change, we actually planned to do in the future. 

We cannot merge it now, since it is breaking some components and changing the layout. This is something, which has to be reviewed by our designer.

Also, I will provide some guidelines for creating PRs, since it looks like you didn't run it, because it couldn't be built without fixing with prettier.

eweren:
> Also, I will provide some guidelines for creating PRs, since it looks like you didn't run it, because it couldn't be built without fixing with prettier.

Yeah a guideline would be good. I ran it and I built it and since both went fine I thought everything was good. Of course I would have come back and check if the pipeline was successful and if not fix those issues

eweren:
Closed because currently out of scope

Issue #983, opened 2022-03-14
Danger zone - frame issue


mbiesiad:
Hi. First of all, really great project!

Repro steps:
1. As a logged in user, go to the Project settings:
`https://app.tolgee.io/projects/<number>/manage/edit`
2. Change browser window size, make a window width smaller
3. Take a look at the Danger zone (out of the frame)
(important note! these steps with **expanded** menu from the left side)

Screenshots.
Normal view (correct):
![image](https://user-images.githubusercontent.com/18367606/158232748-1d0808c7-d13d-46d6-a309-332fa8c68077.png)

Smaller view (still fine):
![image](https://user-images.githubusercontent.com/18367606/158232872-3ec8ad6b-3fd8-434e-ad11-a35d47196a01.png)

Mentioned issue (incorrect):
![image](https://user-images.githubusercontent.com/18367606/158232978-948abaff-32b8-4f20-b230-92b7980ce818.png)

Best wishes,

JanCizmar:
There are issues similar to this, but thanks a lot! 👍

eweren:
@JanCizmar should be resolved because of #1005

JanCizmar:
Thanks @eweren

PR #1005, opened 2022-03-21
#983 fix: Danger zone alignment on small screens

eweren:


stepan662 (approved):
Thanks for the fix, it looks much better now

PR #1006, opened 2022-03-22
Stepangranat/language permissions fe

stepan662:


webapp/src/views/projects/members/component/InviteDialog.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const project = useProject();
+  const invite = useApiMutation({
+    url: '/v2/projects/{projectId}/invite',
+    method: 'put',
+    invalidatePrefix: '/v2/projects/{projectId}/invitations',
+  });
+
+  const yupSchema = useMemo(
+    () =>
+      Yup.object({
```
JanCizmar, line 67: There is a file storing all Yup schemas, we should probably be consistent in this. It's there to make it easy to edit constraints for the fields globally.

webapp/src/views/projects/members/component/InviteDialog.tsx
```diff
@@ -0,0 +1,222 @@
...
+        permission: Yup.string(),
+        permissionLanguages: Yup.array(Yup.string()),
+        type: Yup.string(),
+        text: Yup.string().when('type', (val: string) =>
```
JanCizmar, line 71: Maybe it would be better to split it to two different fields. It's not very readable :(

stepan662: Hmm, I think it's more elegant like this as the field will keep the state when you switch between 'link' and 'email'.

JanCizmar: ouk

Issue #915, opened 2022-02-03
Add project dashboard page


JanCizmar:
Project dashboard should provide statistics for whole project and for each language as well as links to some basic actions.

## Information about project
- [ ] ID
- [ ] Members (Users)
- [ ] Project Owner (Organization/User)
- [ ] Total key count
- [ ] Total word/strings count ([total number of stings] = [number of keys] * [number of languages])
- [ ] Total translated percent + words/strings (switchable?)
- [ ] Total number of reviewed percent + words/strings (switchable?)

## Individual language info
 - [ ] Name / Original name
 - [ ] Tag
 - [ ] Is base
 - [ ] Progress bar with percentage
 - [ ] link to edition
 - [ ] link to translations
 - [ ] other actions: Delete, Auto Translation, Export
 - [ ] Collapsed stats for each lang
   - [ ] Word / String count in each state, percent
   - [ ] Total words/strings

PR #1085, opened 2022-04-21
Jancizmar/add project dashboard page #915

JanCizmar:


webapp/src/views/projects/dashboard/DashboardPreview.tsx
```diff
@@ -0,0 +1,53 @@
...
+            </Grid>
+          </StyledLanguagesAndActivityGridContainer>
+          <DailyActivityChart />
+        </StyledContainer>
+      )}
+    </BaseView>
+  );
+};
+
+const StyledContainer = styled(Box)`
```
stepan662, line 45: Could we have convention that we define this before the main component?

webapp/src/views/projects/dashboard/ProjectTotals.tsx
```diff
@@ -0,0 +1,175 @@
...
+              {t('project_dashboard_tag_count', 'Tags')}
+            </TileDescription>
+          </TileDataItem>
+        </Tile>
+      </Tiles>
+    </>
+  );
+};
+
+const Tiles = styled(Box)`
```
stepan662, line 118: Missing `Styled` prefix. I don't persist on that, but it's already everywhere, so I would do it like that.

webapp/src/views/projects/dashboard/ProjectTotals.tsx
```diff
@@ -0,0 +1,175 @@
...
+const TileDescription = styled(Typography)`
+  font-size: 18px;
+`;
+
+const TileSubDescription = styled(Typography)`
+  font-size: 13px;
+  color: #808080;
+`;
+
+const OwnerAvatar = (props: {
```
stepan662, line 160: I think it's nicer to define Prop types separately:
```
type Props = {
  ...
}

const OwnerAvatar: React.FC<Props> = (...) => {}
```

stepan662 (approved):

webapp/src/views/projects/dashboard/DashboardPreview.tsx
JanCizmar: Is really the first thing I want to read in react component the style definitions? The convention should be more defining it after, so reader could read less and directly get to the important part.

webapp/src/views/projects/dashboard/ProjectTotals.tsx
JanCizmar: Right.

webapp/src/views/projects/dashboard/ProjectTotals.tsx
JanCizmar: Its more verbose, but OK

PR #1125, opened 2022-05-16
feat: Implement Google oauth login

VerHext:
Especially for companies, the integration of Google as an authorization service is very attractive. This PR extends the oauth login for Google. There is also an option `tolgee.authentication.google.workspace-domain: "domain.com"` to restrict authentication to members of the workspace only, 

However, there are some adjustments that need to be made to the following components:
- The frontend handling, as it would be too complex if two _(or in the future more Oauth providers)_ are handled in this component. Outsourcing the functions to `OAuthService.tsx` seems to make the most sense to me here, am open to other suggestions.
- The transmission of the code to the backend was changed from `PathVariable` to `RequestParameter`, because the Google code contains a slash, that would be difficult to encode this correctly. In addition, the parameter `redirect_uri` is now a part of the API request by default, because Google needs this to verify the request.

So please give me feedback and your opinions about this PR. Thanks for the project


JanCizmar:
Thanks a lot for contributing. I am going to test it right now. What I see from the test results, you haven't run the `./gradlew ktlintformat` to format the code, so the check fails.

JanCizmar (approved):

JanCizmar:
So I tested it and it works just fine. There is just missing documentation for this in documentation section so let me know if you would like do it yourself or not. 

https://github.com/tolgee/documentation

We would also probably change style of the login buttons, but we have to consult this with our designer, so we can do it later.

JanCizmar:
Some E2E tests are still failing :(. Do you want to fix it yourself? Otherwise, we can merge it into some other branch in Tolgee repo, then I can fix it and then we can merge it to main.

VerHext:
I am not according to your speed 😄.  I looked at this yesterday and have yet to find a "nice" solution to the failed e2e test. It seems to me that cy can't match URLs with wildcard if it contains a URL as query parameter. 
If you have a solution ready, feel free to adopt it, otherwise I'd have to look into it in more detail today.

JanCizmar:
Hi, the last failing test is related to missing translations, because this is external PR and there is Tolgee API key missing, this will be resolved when merged to main. :)

Unfortunately, Stepan merged some new changes into main, so there is a conflict. 😆

JanCizmar:
Ou. I tried to resolve the conflict, but it sees I did a poor job, because now eslint is failing. Never ming. I am going to fix it in our main. Thanks a lot for this great contribution.

JanCizmar:
OK, so I resolved all and created new PR: #1133

PR #1141, opened 2022-05-23
feat: Dashboard loading placeholder and fix user avatar

stepan662:


webapp/src/views/projects/dashboard/DashboardView.tsx
```diff
@@ -98,27 +147,27 @@ export const DashboardView = () => {
...
-          )}
-          <Box gridArea="activityList">
-            <ActivityList />
-          </Box>
-          <Box gridArea="activityChart">
-            <DailyActivityChart />
-          </Box>
-        </StyledContainer>
+        {anythingLoading ? (
+          <EmptyListMessage loading={true} />
```
JanCizmar, line 151: Base view has also loading props, doesn't it?

JanCizmar (approved):
Its fix, not a feature.

webapp/src/views/projects/dashboard/DashboardView.tsx
stepan662: Yes, but that does loading at the top

JanCizmar: Oki doke.

PR #1174, opened 2022-06-07
fix: Use new delete endpoint for keys delete

stepan662:


JanCizmar (approved):
I think you meant "fix: Use new delete endpoint for keys delete"

PR #1254, opened 2022-07-18
Stepangranat/billing fe improvements

stepan662:


webapp/src/component/billing/CircularBillingProgress.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type Props = {
+  percent: number;
+  size?: number;
+};
+
+export const CircularBillingProgress = ({ percent, size = 28 }: Props) => {
+  const normalized = percent > 100 ? 100 : percent < 0 ? 0 : percent;
+  const critical = normalized < BILLING_CRITICAL_PERCENT;
+  const strokeDashoffset = CIRCUIT - (normalized / 100) * CIRCUIT;
```
JanCizmar, line 36: Offset

webapp/src/component/security/UserMenu/UserMenu.tsx
```diff
@@ -173,6 +171,33 @@ export const UserMenu: React.FC = () => {
                 {item.label}
               </MenuItem>
             ))}
+
+            {showBilling && (
```
JanCizmar, line 175: Separate component?

webapp/src/component/security/AcceptInvitationHandler.tsx
```diff
@@ -27,7 +27,7 @@ const AcceptInvitationHandler: FunctionComponent<AcceptInvitationHandlerProps> =
     const match = useRouteMatch();
 
     const code = match.params[PARAMS.INVITATION_CODE];
-    const initialDataDispatch = useInitialDataDispatch();
+    const initialDataDispatch = useGlobalDispatch();
```
JanCizmar, line 30: I guess the const  should be renamed as well.

webapp/src/globalContext/globalActions.ts
```diff
@@ -0,0 +1,22 @@
+import { components } from 'tg.service/apiSchema.generated';
+
+type OrganizationModel = components['schemas']['OrganizationModel'];
+type UsageModel = components['schemas']['UsageModel'];
+
+export type GlobalActionType =
```
JanCizmar, line 6: Maybe it would be better to avoid this redux style actions and use just functions? Is there any benefit of this approach I can't see?

webapp/src/globalContext/GlobalContext.tsx
```diff
@@ -0,0 +1,40 @@
...
+    globalDispatchRef.current = dispatch;
+
+    const contextData = {
+      ...initialData.data!,
+      isFetching: initialData.isFetching,
+      isLoading: initialData.isLoading,
+      organizationUsage: organizationUsage.data,
+    };
+
+    return [contextData, dispatch];
```
JanCizmar, line 39: Why can't we just export the methods and data in an object without the Redux approach? Then you woudn't have to define the type separately. Also also it would imporve the readibility of the code. In my opinion. Every time I see. ```something.dispatch({type: 'DO_SOMETHING''})``` It takes me like 30 seconds to understand what happening. ```something.doSomething()``` is pretty clear right away.

webapp/src/globalContext/helpers.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const useUser = () => useGlobalContext((v) => v.userInfo);
+
+export const usePreferredOrganization = () => {
+  const initialDataDispatch = useGlobalDispatch();
+  const preferredOrganization = useGlobalContext(
+    (v) => v.preferredOrganization!
+  );
+  const updatePreferredOrganization = (org: number | OrganizationModel) =>
+    initialDataDispatch({ type: 'UPDATE_ORGANIZATION', payload: org });
```
JanCizmar, line 17: so initial or global data?

webapp/src/views/organizations/useLeaveOrganization.tsx
```diff
@@ -17,7 +17,7 @@ export const useLeaveOrganization = () => {
     method: 'put',
   });
   const history = useHistory();
-  const initialDataDispatch = useInitialDataDispatch();
+  const initialDataDispatch = useGlobalDispatch();
```
JanCizmar, line 20: ...

webapp/src/globalContext/helpers.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return { preferredOrganization, updatePreferredOrganization };
+};
+
+export const useOrganizationUsage = () => {
+  return useGlobalContext((v) => v.organizationUsage!);
+};
+
+export const useOrganizationUsageMethods = () => {
```
JanCizmar, line 25: So if there were the methods exposed right away, than you wouldn't need to create this helper method.

JanCizmar (changes requested):
The user and langiage menus should have box shadow instead of border. Also there are double bottom borders in light mode.

webapp/src/component/billing/CircularBillingProgress.tsx
stepan662: It's a name of the property in css: 'stroke-dashoffset' -> 'strokeDashoffset'

JanCizmar: They should fix it then.

webapp/src/globalContext/globalActions.ts
stepan662: With separate functions it's difficult to make them stable (not causing re-renders).

webapp/src/globalContext/GlobalContext.tsx
stepan662: ⬆️

webapp/src/globalContext/helpers.tsx
stepan662: global

webapp/src/views/organizations/useLeaveOrganization.tsx
stepan662: 👍

webapp/src/component/security/UserMenu/UserMenu.tsx
stepan662: 👍

JanCizmar (approved):

PR #1346, opened 2022-09-04
feat: Improve the German translation to make it more natural and consistent

waigel:


JanCizmar:
Hello, thanks a lot. I've imported the translations to our Tolgee project. So it will be released with next release.

PR #1347, opened 2022-09-05
feat: Websockets improvements & listening on frontend

JanCizmar:


backend/app/src/main/kotlin/io/tolgee/websocket/ActivityWebsocketListener.kt
```diff
@@ -0,0 +1,90 @@
...
+    return userId?.let {
+      val user = userAccountService.find(userId).orElse(null) ?: return@let null
+      ActorInfo(
+        type = ActorType.USER,
+        data = userAccountModelAssembler.toModel(user)
+      )
+    } ?: ActorInfo(type = ActorType.UNKNOWN, data = null)
+  }
+
+  private fun onTranslationDAtaModified(
```
cyyynthia, line 50: `Data` is improperly cased

webapp/src/views/projects/WebsocketPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+        authentication: { jwtToken: jwtToken },
+        serverUrl: process.env.REACT_APP_API_URL,
+      });
+      client.subscribe(
+        `/projects/${project.id}/translation_data_modified`,
+        (data) => addMessage(JSON.stringify(data, undefined, 2))
+      );
+      return () => client.disconnect();
+    }
+  }, [config, project, jwtToken]);
```
cyyynthia, line 28: I'm unsure about how this hook's logic is handled.
The project only affects the client *subscription*, and shouldn't trigger a full socket reconnect.

I'd have the socket connection be handled in a separate effect hook, which depends on the `jwtToken` (config seem an unused dependency as far as I can see); and then have a hook managing subscription which depends on the `client` and the `project`. Performance nitpick here.

webapp/src/views/projects/translations/context/services/useWebsocketListener.ts
```diff
@@ -0,0 +1,67 @@
...
+                      name: 'keyName',
+                    }),
+                  },
+          }));
+          translationService.updateTranslationKeys(keyUpdates);
+        }
+      );
+      return () => client.disconnect();
+    }
+  }, [config, project, jwtToken]);
```
cyyynthia, line 55: See my previous comment on websocket connection

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/WebsocketProperties.kt
```diff
@@ -0,0 +1,8 @@
+package io.tolgee.configuration.tolgee
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties(prefix = "tolgee.websocket")
+class WebsocketProperties(
+  var useRedis: Boolean = false,
```
cyyynthia, line 7: Is keeping this configurable really meaningful? If a Redis server is configured, I think there is nothing wrong with automatically enabling Redis routing, and falling back to standard routing when Redis is not configured. I'm not so sure about the way things are handled internally, but I'd assume this has not much impact on the dispatch performance?

backend/app/src/main/kotlin/io/tolgee/websocket/Types.kt
```diff
@@ -0,0 +1,7 @@
+package io.tolgee.websocket
+
+enum class Types() {
+  TRANSLATION_DATA_MODIFIED;
+
+  val typeName get() = name.lowercase()
```
cyyynthia, line 6: Nitpick here, but considering event names take the form of URI paths, I'd consider having the name dash-cased would be more consistent overall. Very minor though.

backend/app/src/main/kotlin/io/tolgee/websocket/WebSocketConfig.kt
```diff
@@ -0,0 +1,71 @@
...
+
+          if (projectId != null) {
+            val user = (accessor.user as? UsernamePasswordAuthenticationToken)?.principal as? UserAccountDto
+              ?: throw MessagingException("Unauthorized")
+
+            try {
+              securityService.getProjectPermission(projectId = projectId, user.id)
+            } catch (e: Exception) {
+              null
+            } ?: throw MessagingException("Unauthorized")
```
cyyynthia, line 59: Another nitpick, if we want to follow HTTP-like semantics the error should be "Forbidden" for this case rather than Unauthorized.

cyyynthia (changes requested):

webapp/src/views/projects/WebsocketPreview.tsx
JanCizmar: You're right. But we currently need the connection only on the translation page, so It makes sense for now, to disconnect it when project changes, otherwise we would have to share the connection in some global context and so keep the connection alive also when it's not needed. In the future, when we add more endpoints I will fix this.

webapp/src/views/projects/translations/context/services/useWebsocketListener.ts
JanCizmar: See my previous comment.

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/WebsocketProperties.kt
JanCizmar: I would keep it configurable. The Redis configuration depends on this prop as well. 

See: kotlin/io/tolgee/configuration/ConditionalRedissonAutoConfiguration.kt
```
@Configuration
@ConditionalOnClass(Redisson::class, RedisOperations::class)
@AutoConfigureBefore(RedisAutoConfiguration::class)
@EnableConfigurationProperties(RedissonProperties::class, RedisProperties::class)
@ConditionalOnExpression(
  "\${tolgee.websocket.use-redis:false} or " +
    "(\${tolgee.cache.use-redis:false} and \${tolgee.cache.enabled:false})"
)
class ConditionalRedissonAutoConfiguration : RedissonAutoConfiguration()
```

To fix this I would have to create some global prop like "tolgee.redis.enabled" to enable Redis. But maybe user would like to use redis only for caching, since they have only single instance and not for websocket pub/sub.

stepan662 (approved):

webapp/src/constants/links.tsx
```diff
@@ -210,6 +210,11 @@ export class LINKS {
     'socket_io_preview'
```
cyyynthia, line 210: Shouldn't this be removed as I assume it has gone unused?

cyyynthia (approved):

JanCizmar:
@all-contributors please add @JanCizmar for All

JanCizmar:
@all-contributors please add @JanCizmar for code

PR #1363, opened 2022-09-09
feat: Support for right-to-left languages

JanCizmar:


webapp/src/views/projects/translations/TranslationVisual.tsx
```diff
@@ -112,3 +119,13 @@ export const TranslationVisual: React.FC<Props> = ({
     );
   }
 };
+
+const DirectionLocaleWrapper: FC<{ languageTag: string }> = (props) => (
```
stepan662, line 123: Maybe separate file?

stepan662 (approved):

webapp/src/views/projects/translations/TranslationVisual.tsx
JanCizmar: hueeeh, ok.

PR #1365, opened 2022-09-09
feat: account security enhancements

cyyynthia:
This PR features a bunch of enhancements and new features to users account security.
 - Added Two-Factor Authentication support
 - Changing email requires the current password
 - Changing password requires the current password
 - Password change form has been moved to a dedicated "Account Security" section
 - Sessions are invalidated upon changing password or enabling/disabling 2FA
 - LDAP accounts can no longer change their password or their login
 - Accounts created via a third-party login must go through a password reset flow before they can manage their account security settings

Notable internal changes
 - `POST /v2/user` has been deprecated in favor of `PUT /v2/user`
 - The frontend now uses `/v2/user` instead of `/api/user`
 - Added a `PasswordEncoder` bean for use throughout the backend instead of creating a new object each time
 - Maximum password size has been lowered to 50 from 100, to reflect the limitations of the BCrypt password encoder

Might also include random small changes to fix minor issues I've encountered I forgot to list here


webapp/src/views/userSettings/accountSecurity/EnableMfaDialog.tsx
```diff
@@ -0,0 +1,172 @@
...
+            >
+              <TextField
+                inputProps={{
+                  'data-cy': 'mfa-enable-dialog-otp-input',
+                }}
+                name="otp"
+                placeholder="000000"
+                label={t('account-security-mfa-otp-code')}
+              />
+              <TextField
```
stepan662, line 158: Could we maybe switch the order of the inputs? Since you need to be fast to type correct mfa to not to timeout, it's better to insert password first and mfa second.

webapp/src/views/userSettings/accountSecurity/MfaSettings.tsx
```diff
@@ -0,0 +1,84 @@
...
+const StyledEnabled = styled('span')`
+  color: ${({ theme }) => theme.palette.success.main};
+`;
+
+const StyledDisabled = styled('span')`
+  color: ${({ theme }) => theme.palette.error.main};
+`;
+
+const StyledEnabledParam = (value) => <StyledEnabled>{value}</StyledEnabled>;
+const StyledDisabledParam = (value) => <StyledDisabled>{value}</StyledDisabled>;
```
stepan662, line 17: You don't need these, it's possible to do it this way:
```
<T parameters={{ b: <StyledEnabled/> }}>
```
value will be injected as child automatically

webapp/src/views/userSettings/userProfile/UserProfileView.tsx
```diff
@@ -58,37 +63,53 @@ export const UserProfileView: FunctionComponent = () => {
...
+            <TextField
+              variant="standard"
+              name="name"
+              label={<T>User settings - Full name</T>}
+            />
+            <TextField
+              variant="standard"
+              name="email"
+              disabled={isManaged}
+              helperText={isManaged ? t('managed-account-field-hint') : void 0}
```
stepan662, line 80: Why not `: undefined`? This is a bit unusual.

webapp/src/component/security/LoginView.tsx
```diff
@@ -26,16 +33,35 @@ import {
...
+type LoginViewCredentialsProps = {
+  credentialsRef: RefObject<Credentials>;
+  onMfaEnabled: () => void;
+};
+type LoginViewTotpProps = {
+  credentialsRef: RefObject<Credentials>;
+  onMfaCancel: () => void;
+};
+
+export function LoginViewCredentials(props: LoginViewCredentialsProps) {
```
stepan662, line 51: Plz make these separate files, it's easier to understand when it's one file - one component

stepan662 (changes requested):

webapp/src/views/userSettings/userProfile/UserProfileView.tsx
cyyynthia: `void 0` is something I'm used to write as it's shorter, and because of how JavaScript works; tldr it's to avoid shadowing risk and because it skips the lookup step in the V8 engine. That being said, I can write it as `undefined` if it better suits the code style

backend/data/src/main/java/io/tolgee/security/JwtToken.java
```diff
@@ -1,31 +1,37 @@
...
     private final String value;
-    private final Key key;
+
+    private final Jws<Claims> parsed;
 
     public JwtToken(String value, Key key) {
         this.value = value;
-        this.key = key;
+        this.parsed = Jwts.parserBuilder().setSigningKey(key).build().parseClaimsJws(value);
     }
```
JanCizmar, line 19: When I have expired token it throws error 500 now, since the token is validated here, which didn't happen before. Please add some test that expired token throws only 401.

JanCizmar: Maybe it would be better to not parse the key here at all. And let it for the io.tolgee.security.JwtTokenProviderImpl.validateToken

backend/data/src/main/kotlin/io/tolgee/service/MfaService.kt
```diff
@@ -0,0 +1,131 @@
...
+    if (loginRequest.otp.length == 6) {
+      if (!validateTotpCode(user, loginRequest.otp)) {
+        throw AuthenticationException(Message.INVALID_OTP_CODE)
+      }
+    } else {
+      if (!user.mfaRecoveryCodes.contains(loginRequest.otp)) {
+        throw AuthenticationException(Message.INVALID_OTP_CODE)
+      }
+      userAccountService.consumeMfaRecoveryCode(user, loginRequest.otp)
+    }
```
JanCizmar, line 108: This whole logic can be moved to consumeMfaRecovery.

backend/data/src/main/kotlin/io/tolgee/service/UserAccountService.kt
```diff
@@ -202,35 +244,51 @@ class UserAccountService(
...
+    userAccount.password = passwordEncoder.encode(dto.password)
+    return userAccountRepository.save(userAccount)
   }
 
   private fun updateUserEmail(
     userAccount: UserAccount,
     dto: UserUpdateRequestDto
   ) {
     if (userAccount.username != dto.email) {
+      if (!Regex("[^@]+@[^@]+").matches(dto.email)) {
```
JanCizmar, line 287: Maybe we can only make email nullable in the request and when not provided, we can leave it, so we can keep the email check there. This is not a good way how to check e-mail anyway...

backend/data/src/main/kotlin/io/tolgee/service/UserCredentialsService.kt
```diff
@@ -0,0 +1,69 @@
...
+  @set:Autowired
+  lateinit var userAccountService: UserAccountService
+
+  @set:Autowired
+  lateinit var authenticationManager: AuthenticationManager
+
+  fun checkUserCredentials(username: String, password: String): UserAccount {
+    if (configuration.authentication.ldap.enabled) {
+      val details = checkLdapUserCredentials(username, password)
+      return userAccountService.findOptional(details.username).orElseGet {
```
JanCizmar, line 29: The found user should be LDAP user, otherwise this should throw something.

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -2113,4 +2113,31 @@
...
+    </changeSet>
+    <changeSet author="cynthia (generated)" id="1661345119881-1">
+        <addColumn tableName="user_account">
+            <column name="account_type" type="varchar(255)"/>
+        </addColumn>
+    </changeSet>
+    <changeSet author="cynthia" id="1661345119881-2">
+        <sql dbms="postgresql">
+            UPDATE user_account SET account_type = 'LOCAL';
+            UPDATE user_account SET account_type = 'THIRD_PARTY' WHERE password IS NULL;
```
JanCizmar, line 2134: Not really. There are users in the production DB who are registered via third party and have passwords.

backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/UserMfaController.kt
```diff
@@ -0,0 +1,50 @@
...
+  private val mfaService: MfaService,
+  private val jwtTokenProvider: JwtTokenProvider
+) {
+  @PutMapping("/totp")
+  @Operation(summary = "Enables TOTP-based two-factor authentication. Invalidates all previous sessions upon success.")
+  @DenyPatAccess
+  fun enableMfa(@RequestBody @Valid dto: UserTotpEnableRequestDto): JwtAuthenticationResponse {
+    mfaService.enableTotpFor(authenticationFacade.userAccountEntity, dto)
+    return JwtAuthenticationResponse(
+      jwtTokenProvider.generateToken(authenticationFacade.userAccountEntity.id).toString()
```
JanCizmar, line 30: Why new token?

backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/UserMfaController.kt
```diff
@@ -0,0 +1,50 @@
...
+    )
+  }
+
+  @DeleteMapping("/totp")
+  @Operation(summary = "Disables TOTP-based two-factor authentication. Invalidates all previous sessions upon success.")
+  @DenyPatAccess
+  fun disableMfa(@RequestBody @Valid dto: UserTotpDisableRequestDto): JwtAuthenticationResponse {
+    mfaService.disableTotpFor(authenticationFacade.userAccountEntity, dto)
+    return JwtAuthenticationResponse(
+      jwtTokenProvider.generateToken(authenticationFacade.userAccountEntity.id).toString()
```
JanCizmar, line 40: Why new token?

backend/app/src/main/kotlin/io/tolgee/api/v2/hateoas/user_account/UserAccountModel.kt
```diff
@@ -11,6 +11,8 @@ data class UserAccountModel(
   val username: String,
   var name: String?,
   var emailAwaitingVerification: String?,
+  var mfaEnabled: Boolean,
```
JanCizmar, line 14: This model is used also publicaly. Pease create separate model. We don't want to provide this information to everybody.

backend/app/src/main/kotlin/io/tolgee/controllers/PublicController.kt
```diff
@@ -94,14 +87,25 @@ class PublicController(
...
+      message.subject = "Password reset"
+      message.text = """Hello!
+To reset your password click this link: 
 $url
 
- If you have not requested password reset, please just ignore this e-mail."""
+If you have not requested password reset, please just ignore this e-mail."""
+    }
+
     message.from = properties.smtp.from
```
JanCizmar, line 109: Please do it like here: io.tolgee.component.email.EmailVerificationSender.sendEmailVerification

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/UserMfaControllerTest.kt
```diff
@@ -0,0 +1,152 @@
...
+    enableMfa()
+    val requestDto = UserMfaRecoveryRequestDto(
+      password = initialPassword
+    )
+
+    var fromDb = userAccountService.findOptional(initialUsername)
+    Assertions.assertThat(fromDb).isNotEmpty
+    Assertions.assertThat(fromDb.get().mfaRecoveryCodes).isEmpty()
+
+    performAuthPut("/v2/user/mfa/recovery", requestDto).andExpect(MockMvcResultMatchers.status().isOk)
```
JanCizmar, line 68: .andExpect(MockMvcResultMatchers.status().isOk) - > .andIsOk

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/UserMfaControllerTest.kt
```diff
@@ -0,0 +1,152 @@
...
+  fun `it requires valid TOTP code for activation`() {
+    val requestDto = UserTotpEnableRequestDto(
+      totpKey = TOTP_KEY,
+      otp = (mfaService.generateCode(encodedKey) + 1).toString(),
+      password = initialPassword
+    )
+
+    val res = performAuthPut("/v2/user/mfa/totp", requestDto)
+      .andExpect(MockMvcResultMatchers.status().isBadRequest)
+      .andReturn()
```
JanCizmar, line 84: -> andIsBadRequest

JanCizmar: And the same way everywhere else

JanCizmar (changes requested):
Good job! Thanks!

backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/UserMfaController.kt
cyyynthia: > Sessions are invalidated upon changing password or enabling/disabling 2FA

Enabling or disabling MFA causes sessions to be invalidated, so a new token is returned to keep the current session alive

backend/data/src/main/resources/db/changelog/schema.xml
cyyynthia: They still do have local authentication available, and accounts registered via third party are migrated to "local" status upon a password reset flow. The `THIRD_PARTY` flag is really for accounts which only have the OAuth flow as an authentication mean, and is used by the frontend to show or not account security settings

backend/data/src/main/kotlin/io/tolgee/service/UserAccountService.kt
cyyynthia: I used the bare minimum regex to check emails, the common rationale around this being that it's useless to check for a stricy regex which is heavier to process and will still let invalid emails through (or for some regexes, reject actually valid emails because they make wrong assumptions). There are a lot of articles about it online

I can use a stricter regex, or look into manually calling hibernate validator if that's too much of a concern

backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/UserMfaController.kt
JanCizmar: I don't understand. Where it causes it?

backend/data/src/main/resources/db/changelog/schema.xml
cyyynthia: Ideally, the OAuth account used could be moved to a separate table of linked accounts which could be used later on for 3rd party integrations, so that it becomes an alternative login flow and also does allow for more than 1 account to be linked.

backend/data/src/main/kotlin/io/tolgee/service/UserAccountService.kt
JanCizmar: Call the validator. This can cause another exceptions thrown while sending e-mails to invalid e-mails.

backend/data/src/main/resources/db/changelog/schema.xml
JanCizmar: OK

backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/UserMfaController.kt
cyyynthia: https://github.com/tolgee/tolgee-platform/blob/73983eba9c7eaf60a9e969c123d69f464ea6f7c1/backend/data/src/main/kotlin/io/tolgee/service/UserAccountService.kt#L167

It's an intentional behavior I've seen from multiple platforms in my experience, as changing 2FA settings is considered credentials change
it's easy to drop this behavior if it's something you'd prefer not to happen

JanCizmar: Aha, OK, I was only curious, how you can invalidate JWT token. But this way seems cool!

JanCizmar:
Would you please squash the commits when all is resolved and add a single `feat: ` commit?

cyyynthia:
Erf, I hate force pushes from the bottom of my heart but I can do it; there's also the "Squash & merge" feature of GitHub that'd do a single commit to the main branch

JanCizmar:
> Erf, I hate force pushes from the bottom of my heart but I can do it; there's also the "Squash & merge" feature of GitHub that'd do a single commit to the main branch

If you hate it so much, lets try the squash and merge :)

backend/data/src/main/kotlin/io/tolgee/security/JwtToken.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import io.jsonwebtoken.Claims
+import io.jsonwebtoken.Jws
+import io.jsonwebtoken.Jwts
+import java.security.Key
+import java.util.*
+
+class JwtToken(private val value: String, private val key: Key?) {
+  private lateinit var _parsed: Jws<Claims>
+  private val parsed: Jws<Claims>
```
JanCizmar, line 11: Check this out.
https://kotlinlang.org/docs/delegated-properties.html#lazy-properties

JanCizmar (approved):

PR #1366, opened 2022-09-09
 feat: Filter by multiple key names, fixed lazy loading

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/dtos/request/translation/TranslationFilters.kt
```diff
@@ -31,7 +31,7 @@ To add multiple languages, repeat this param (eg. ?languages=en&languages=de)"""
   var search: String? = null
 
   @field:Parameter(description = "Selects only one key with provided name")
```
cyyynthia, line 33: Description should be changed as multiple keys can be fetched

backend/data/src/main/kotlin/io/tolgee/dtos/request/translation/TranslationFilters.kt
```diff
@@ -31,7 +31,7 @@ To add multiple languages, repeat this param (eg. ?languages=en&languages=de)"""
...
-  var filterKeyName: String? = null
+  var filterKeyName: List<String>? = null
 
   @field:Parameter(description = "Selects only one key with provided id")
```
cyyynthia, line 36: Description should be changed as multiple keys can be fetched

cyyynthia (approved):

PR #1371, opened 2022-09-13
fix: Fix MFA tests

JanCizmar:


cyyynthia (approved):
Good catch! I did check if codes starting with 0 were working on the frontend, but completely forgot it could be an issue here.

PR #1376, opened 2022-09-15
feat: limit number of selected languages

stepan662:


JanCizmar (approved):

vlad-roid:
@stepan662 @JanCizmar can this please be fixed so there is no hard limit? I want to be able to see and review ALL languages at once and don't care if it's slower to load. A "(de)select all" would be immensely helpful as well.
Right now reviewing 30 languages means clicking 10 check boxes one by one, then deselecting them, clicking the next 10 check boxes, deselecting them, clicking the last 10 check boxes. This is such a ridiculous waste of time.

PR #1383, opened 2022-09-28
fix: consistent sorting when sort is unspecified

### LINKED ISSUE #1345: Cursor-based pagination skips results when no sort is specified
The UI always specifies `keyName` sort, but if you leave it `null` (even though it's an optional API parameter) the back end returns inconsistent results and skips some keys when using the cursor to fetch all data.

cyyynthia:
When no sorting was specified, Tolgee defaulted to sorting by `keyId` and `keyName`. However, the returned cursor only specified `keyId` and this caused subsequent requests to not use the same sorting method. This patch fixes it by only using `keyId` when no sort is specified.

Regression test included.

closes #1345

### DIFF
--- a/backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/translation/V2TranslationsController.kt
+++ b/backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/translation/V2TranslationsController.kt
@@ -35,6 +35,7 @@ import io.tolgee.model.enums.ApiScope
 import io.tolgee.model.enums.TranslationState
 import io.tolgee.model.key.Key
 import io.tolgee.model.translation.Translation
+import io.tolgee.model.views.KeyWithTranslationsView
 import io.tolgee.security.AuthenticationFacade
 import io.tolgee.security.apiKeyAuth.AccessWithApiKey
 import io.tolgee.security.project_auth.AccessWithAnyProjectPermission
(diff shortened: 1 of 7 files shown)


JanCizmar (approved):

stepan662 (approved):

JanCizmar:
Hmm, this way, we are actually changing the public API. We should probably sort by name by default.

cyyynthia:
> Hmm, this way, we are actually changing the public API

Well, the behavior when no sorting is applied is undocumented, so I figured the cursor data was reflective of the expected behavior.

If we want to sort by name by default, this should also be specified in the docs, as the default value for `sort`

JanCizmar:
Hmm, I see. This is also fix for this 
https://github.com/tolgee/tolgee-platform/issues/1345 then. Anyway, I don't think this is proper way how to fix it. I would fix it by removing the default sort related stuff from both cursorUtil and the TranslationViewBuilder and add @SortDefault annotation to the controller method. That way we don't have to keep it in sync.

JanCizmar:
See io.tolgee.api.v2.controllers.organization.OrganizationController#getAll for example.

cyyynthia:
I see. Should I also make the change to sort by name like it previously attempted to do?

JanCizmar:
Let's keep it by id. It's faster.

cyyynthia:
Hm, well in that case I'm unsure if the `SortDefault` would do the trick here: my understanding is that `id` is always used for sorting to make sure there is always a unique field used. So, whenever a user sorts by translation text for example, which is not a unique field, the `id` ensures no data is skipped over.

As such, applying the sort by `id` should always be a thing regardless of user choice; by having it specified as a default only we lose the guarantee sorting can be performed without skipping/duplicating data even when the sorting is applied to non-unique fields.

JanCizmar:
then creating of `getSafeSortPageable()` in V2TranslationController with this body should do the trick
```
  val sortContainsId = pageable.sort.none { it.property == "id" }
    var sort = pageable.sort
    if (!sortContainsId) {
      sort = sort.and(Sort.by(Sort.Direction.ASC, "id"))
    }
    pageable = PageRequest(pageable.pageNumber, pageable.pageSize, sort)

```

PR #1414, opened 2022-11-03
fix: Orphan removal of Translations & Comments, Retry intermittent E2…

JanCizmar:
…E tests

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/UserMfaControllerTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/UserMfaControllerTest.kt
@@ -6,6 +6,7 @@ import io.tolgee.dtos.request.UserTotpEnableRequestDto
 import io.tolgee.fixtures.andIsBadRequest
 import io.tolgee.fixtures.andIsForbidden
 import io.tolgee.fixtures.andIsOk
+import io.tolgee.fixtures.retry
 import io.tolgee.testing.AuthorizedControllerTest
 import io.tolgee.testing.assertions.Assertions.assertThat
 import org.apache.commons.codec.binary.Base32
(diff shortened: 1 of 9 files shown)


cyyynthia (approved):

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/UserMfaControllerTest.kt
```diff
@@ -117,28 +118,30 @@ class UserMfaControllerTest : AuthorizedControllerTest() {
...
-    loginAsAdminIfNotLogged()
-    Thread.sleep(2000)
-
-    val disableRequestDto = UserTotpDisableRequestDto(
-      password = initialPassword
-    )
-    performAuthDelete("/v2/user/mfa/totp", disableRequestDto).andIsOk
-    refreshUser()
-    performAuthGet("/v2/user").andExpect(MockMvcResultMatchers.status().isUnauthorized)
+    retry {
```
cyyynthia, line 121: All the tests should be wrapped in a retry block, due to the dependency to timing stuff. Ideally, we should look into cleaning up these timing-sensitive tests and inject a `Clock` object, so we can easily mock time and simulate huge time shifts (or "disable" time).

An internal endpoint to set the time would also be handy for the Cypress tests, so they can also manipulate the time, and avoid flaky tests.

This would be cleaner and also applies to token expiry tests. 🤔

JanCizmar: It's already possible to mock time in integrations tests. There is `io.tolgee.component.CurrentDateProvider` which can be used to mock the time. Using `@MockBean` I just didn't want to dig too deep here. Feel free, to fix it right, but for now I would just merge it as it is. 😄

cyyynthia: I'll open an issue about it 👍🏻 

I'm unsure the mfa service would respond to mocking that, it'll need to be checked

PR #1427, opened 2022-11-14
Add namespaces on FE

stepan662:


webapp/src/views/projects/translations/Namespace/NamespaceContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    const handleClose = () => {
+      setOpen(undefined);
+    };
+
+    return (
+      <>
+        <StyledNamespace
+          ref={ref}
+          onClick={toggle}
```
JanCizmar, line 68: The onclick shouldn't filter it, but open the dropdown. User can choose it there. There should be some hover as there is hover for tags.

webapp/src/views/projects/translations/Namespace/NamespaceBanner.tsx
```diff
@@ -0,0 +1,58 @@
+import { useEffect, useRef } from 'react';
+import { styled } from '@mui/material';
+import { useHeaderNsDispatch } from '../context/HeaderNsContext';
+import { NamespaceContent } from './NamespaceContent';
+import { NsBannerRecord } from '../context/useNsBanners';
+
+const StyledNsRow = styled('div')`
+  display: flex;
+  background: ${({ theme }) => theme.palette.emphasis[50]};
+  height: 17px;
```
JanCizmar, line 10: According the design, it should have height 1px. Last in ns and last in ns should be identified and should have larger padding-bottom/top to create space for the banner.

JanCizmar (changes requested):
- I forgot to add namespace change to activities.
- We really need the namespace selector
- When adding new language via select in import view, it refreshed the table. Which is OK, but it should show loading over the input instead and keep the data while new language is created
- We should reconsider removing the top banner once again. It behaves pretty weird now, whe user scrolls up with the namespaces and showing top bar.

PR #1437, opened 2022-11-18
fix: Empty namespace filter

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/service/query_builders/translationViewBuilder/QueryGlobalFiltering.kt
```diff
@@ -98,6 +98,14 @@ class QueryGlobalFiltering(
     }
   }
 
+  private fun getFilterNamespace(): List<String>? {
+    val filterNamespace = params.filterNamespace
+    if (filterNamespace != null && filterNamespace.isEmpty()) {
+      return listOf("")
```
stepan662, line 104: Will it work also in this case? `?filterNamespace=&filterNamespace=test`

JanCizmar: Yes.

stepan662 (approved):

PR #1463, opened 2022-12-19
feat: Caching of translations endpoint & configurable structure delim…

JanCizmar:
…iter


backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/translation/V2TranslationsController.kt
```diff
@@ -116,9 +123,37 @@ class V2TranslationsController(
...
-    ns: String? = ""
-  ): Map<String, Any> {
-    return translationService.getTranslations(languages, ns, projectHolder.project.id)
+    ns: String? = "",
+    @Parameter(
+      description = """Delimiter to structure response content. 
+
+e.g. For key "home.header.title" would result in {"home": {"header": "title": {"Hello"}}} structure.
+
+When null, resulting file won't be structured.
```
cyyynthia, line 132:
```suggestion
      description = """Delimiter to structure response content. 

e.g. For key "home.header.title" would result in {"home": {"header": {"title": "Hello"}}} structure.

When null, resulting file will be a flat key-value object.
```

cyyynthia (approved):

PR #1468, opened 2022-12-21
fix: Default structure delimiter in translations endpoint

JanCizmar:


backend/app/src/main/kotlin/io/tolgee/api/v2/controllers/translation/V2TranslationsController.kt
```diff
@@ -153,6 +155,17 @@ When null, resulting file will be a flat key-value object.
...
 
+  /**
+   * It has to be handled manually since spring returns default value even when empty value provided
+   */
+  private fun WebRequest.getStructureDelimiter(): Char? {
+    val structureDelimiterParam = this.parameterMap["structureDelimiter"]?.first() ?: return '.'
+    if (structureDelimiterParam == "") {
+      return null
+    }
+    return structureDelimiterParam.toCharArray().first()
```
cyyynthia, line 166: Unsure about this. What if people want to split with multi-char patterns, like `##` or `//` for <whatever reason>?

JanCizmar: We are not ready for this.

cyyynthia: I guess it's a limitation of the current implementation that takes a Char specifically, but I think multichar delimiters might be a nice-to-have

cyyynthia (approved):

PR #1480, opened 2023-01-03
fix: Startup import sorted & refactored

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/service/StartupImportService.kt
```diff
@@ -34,29 +34,38 @@ class StartupImportService(
...
+    SecurityContextHolder.getContext().authentication = authenticationProvider.getAuthentication(userAccount)
+  }
+
+  private fun isProjectExisting(projectName: String, organization: Organization) =
+    projectService.findAllByNameAndOrganizationOwner(projectName, organization).isNotEmpty()
+
+  private fun getDirsToImport(): List<File>? {
+    val dir = properties.import.dir
+    if (dir !== null && File(dir).exists() && File(dir).isDirectory) {
+      return File(dir).listFiles()?.filter { it.isDirectory }?.sortedBy { it.name }
```
cyyynthia, line 64: Instantiating 3 times the File object doesn't look super pretty. Maybe a `.let` block can do the trick on the dir declaration instead?

cyyynthia (commented):
It'd be nice to add logging to validate what this is doing at runtime. During my troubleshooting yesterday I couldn't easily validate the import was running due to the lack of logging information.

It would also have saved me a few hours yesterday by making the problem more obvious by the log being in the wrong order as well. 😅

cyyynthia (approved):

PR #1500, opened 2023-01-12
feat: Import improvements

JanCizmar:


backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2ImportController/V2ImportControllerAddFilesTest.kt
```diff
@@ -135,6 +142,34 @@ class V2ImportControllerAddFilesTest : AuthorizedControllerTest() {
       }
   }
 
+  @Test
+  fun `it imports empty keys`() {
```
cyyynthia, line 146: Are we sure we want to keep this considering the UI issues it causes?

JanCizmar: Importing only empty keys is edge case. So I would keep it. In most cases user imports also some translations.

Issue #1147, opened 2022-05-27
Explicit "edited" (or "outdated") state


cyyynthia:
It would be a nice QoL feature to have all translated strings be flagged as outdated when the source string is updated. This would make it easier for translators to spot outdated translations, and to exclude them if necessary when bundling translations into an app.

This could also be used during review, or as a marker for translators who tell themselves "I'll deal with this one later", only to forget about it 😅

JanCizmar:
Right! "Outdated" is nice term for such a thing. Thanks! :) We were also thinking about setting the status of the translation reviewed -> translated, but adding another field is probably better idea. :)

JanCizmar:
@Flamefork I think that's a different story. We actually don't store "translation versions", but we store log of changes to selected entities. That's how translation history works. 

What Cynthia describes is a basically a problem of modified base translation. (In most cases in english), which is then translated to other languages. Let's say to these things happen:
1. Developer creates a string in english "Hello World!"
2. Translator translates it to german as "Hallo Welt!"
3. Reviewer marks "Hallo Welt!" as reviewed. 
4. Developer changes the english string to "Hello Universe!"

So now we are in a state when "Halo Welt!" is reviewed, but the base translation means different thing. However, "Hallo Welt" is still correctly spelled and its grammatically correct.
What we can do is setting the state of "Hallo Welt" back to `Translated`, but for many user's who don't use `Reviewed` state at all, this wouldn't be enough. 

So the only way how to correctly solve this is adding a special indicator, which says "The base translation was changed after last modification of this value".

Flamefork:
Thanks for the blazingly fast response! And yep, that's a totally different case, sorry for sidetracking it.

cyyynthia:
I don't think there is an issue opened yet about it, but it seems to me what you're referring to falls under the branching thingy listed on the [roadmap](https://tolgee.io/roadmap), with each branch having a separate state than the base branch. With that, you could have both "Hello World" be reviewed on the base branch, and an updated branch with "Hello Universe", with a different translation and be reviewed too.

ZuzanaOdstrcilova:
Design of behavior "Source change indication", "Auto translation indication" and changing state to "Reviewed": https://www.figma.com/file/UCvcqHZe6owmJV5QolBMoj/tolgee-tasks?node-id=1236%3A1433&t=nIZJBNVUQutAMawU-0

PR #1503, opened 2023-01-13
feat: Source change indication

JanCizmar:


cyyynthia:
Is my understanding that this fixes #1147 correct? :eyes:

cyyynthia (approved):

JanCizmar:
Huh! Yesterday we were brainstorming, how it should be named! Outdated sounds perfect. Ill rename it. And yes it solves that when the frontend is done.

stepan662 (approved):

PR #1511, opened 2023-01-20
feat: Fixtures for Figma plugin

JanCizmar:


JanCizmar:
Since there is large change in DB structure I have to test it with realistic data.

stepan662 (approved):

PR #1512, opened 2023-01-20
fix: Startup import base language

JanCizmar:


cyyynthia:
I guess this works considering it's a feature meant for debugging only rather than production. That being said I'm unsure if allowing `baseLanguage` to go unset is a good idea 🤔

To me it seems that a project without a `baseLanguage` should be an impossible state and therefore shouldn't be possible to represent and create. Unless I'm missing a case where it'd make sense...

JanCizmar:
> To me it seems that a project without a baseLanguage should be an impossible
It is impossible. This is just a fix to make it deterministic.

cyyynthia:
I see. I think it still might be a good idea to clean these things up some time in the future as these cases where non-null is not properly asserted in the type declaration bleed into the OpenAPI specification, and some fields are considered nullable despite this being an impossible state.

cyyynthia (approved):

JanCizmar:
Right. I basically allowed null baseLanguage to skip migrating user data (searching for the base language for all). But yep. This should be handled properly.

PR #1518, opened 2023-01-24
fix: Import deletion throws 500

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/service/dataImport/ImportService.kt
```diff
@@ -154,6 +154,10 @@ class ImportService(
     return this.find(projectId, authorId) ?: throw NotFoundException()
   }
 
+  fun get(id: Long): Import {
+    return importRepository.findById(id).orElse(null) ?: throw NotFoundException()
```
cyyynthia, line 158: Could've been written with `orElseThrow`

cyyynthia (approved):

backend/data/src/main/kotlin/io/tolgee/service/dataImport/ImportService.kt
JanCizmar: Yep.

PR #1525, opened 2023-01-30
Fix import failure when multiple keys are present for a new namespace

llambeau:
Using the import API (`POST /v2/projects/keys/import`) fails with the following error if the keys list includes more than one key for an unexisting namespace:

The error:

```json
{
    "code": "unexpected_error_occurred",
    "params": [
        "org.springframework.dao.DataIntegrityViolationException"
    ]
}
```

The input payload:

```json
{
  "keys": [
    {
      "name": "key1",
      "namespace": "new-namespace",
      "translations": {
        "en": "key one"
      }
    },
    {
      "name": "key2",
      "namespace": "new-namespace",
      "translations": {
        "en": "key two"
      }
    }
  ]
}
```

I think I know where the bug is. The error is in the [KeyService#importKeys method](https://github.com/tolgee/tolgee-platform/blob/main/backend/data/src/main/kotlin/io/tolgee/service/key/KeyService.kt#L180). The code gets a Map of the list of namespaces that already exist, but then loops on all translation keys that need to be created and tries to create a new namespace if it can't be found in the previously computed Map. (Which works on the first occurence of a new key for the new namespace but then fails for all others). Given it's done in a transaction the whole transaction fails and rollbacks applying no changes at all.

This PR's changes were tested in production and the original bug is fixed. That being said I'm not at all a kotlin/java developer and I'm not sure my changes are really idiomatic kotlin/java code.


JanCizmar:
Thanks a lot! :) Nice giutar btw.

PR #1531, opened 2023-02-02
fix: Check for empty namespace in complex-update endpoint

### LINKED ISSUE #1520: Cannot move to empty namespace when key exists
It doesn't check correctly if the key exists in default namespace and fails with 500 on constraint violation

Martin005:
This PR fixes #1520 by implementing `setJsonNamespace` function on the `ComplexEditKeyDto` (the same as in `EditKeyDto` and `CreateKeyDto`).

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/V2KeyInNamespaceControllerTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/V2KeyInNamespaceControllerTest.kt
@@ -2,7 +2,6 @@ package io.tolgee.api.v2.controllers.v2KeyController
 
 import io.tolgee.controllers.ProjectAuthControllerTest
 import io.tolgee.development.testDataBuilder.data.NamespacesTestData
-import io.tolgee.dtos.request.key.ComplexEditKeyDto
 import io.tolgee.dtos.request.key.CreateKeyDto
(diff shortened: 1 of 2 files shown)


backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/V2KeyInNamespaceControllerTest.kt
```diff
@@ -106,6 +107,35 @@ class V2KeyInNamespaceControllerTest : ProjectAuthControllerTest("/v2/projects/"
...
+      .andIsBadRequest
+      .andAssertError
+      .isCustomValidation.hasMessage("key_exists")
+
+    performProjectAuthPut("keys/${keyId}/complex-update", ComplexEditKeyDto(name = keyName, namespace = null))
+      .andIsBadRequest
+      .andAssertError
+      .isCustomValidation.hasMessage("key_exists")
+
+    performProjectAuthPut("keys/${keyId}/complex-update", ComplexEditKeyDto(name = keyName))
```
JanCizmar, line 133: Since ComplexEditKeyDto has null set as default value for namespace, this just tests the same thing as the previous call. It would make sense to create the dto manually using `mapOf("name" to keyName)`, that way, the namespace would be really missing in the request body, so that would test the desired behaviour.

JanCizmar (changes requested):

JanCizmar (approved):

PR #1555, opened 2023-02-15
fix: New translations are outdated

JanCizmar:


stepan662:
The tests are failing probably because before the outdated flags were there after initial import, which I guess changed now.

stepan662 (approved):

JanCizmar:
There is also wrong condition on the backend

JanCizmar:
Can you fix the E2e test?

PR #1600, opened 2023-03-03
feat: Granular permissions FE

stepan662:


e2e/cypress/common/permissions/main.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type MenuItem = Exclude<
+  DataCy.Value & `project-menu-item-${string}`,
+  'project-menu-item-projects'
+>;
+
+type Settings = Partial<Record<MenuItem, boolean>>;
+
+export function checkItemsInMenu(projectInfo: ProjectInfo, settings: Settings) {
```
JanCizmar, line 30: `checkItemsInMenu` doesn't sound right. It's testing permissions in the views not just menu items. It should be probably called `testPermissions()`.

e2e/cypress/common/permissions/dashboard.ts
```diff
@@ -0,0 +1,24 @@
+import { pageIsPermitted, ProjectInfo } from './shared';
+
+function tryClickIfClickable(item: DataCy.Value) {
+  cy.gcy(item).then(($el) => {
```
JanCizmar, line 4: This only tests, you used correct `clickable` classes. But it doesn't test the case there is no clickable class and the user still has permissions to do that by mistake. In that case user get's error and this case doesn't cover that. 

You also know what should be and shouldn't be clickable before the test, so you don't need the condition, you can provide the expected values without conditional testing. 

The naming is also obscure. It doesn't only click if clickable, but it checks whether there is no error thrown when user clicks on clickable item.

e2e/cypress/common/permissions/shared.ts
```diff
@@ -0,0 +1,29 @@
...
+  waitForGlobalLoading();
+  cy.get('.SnackbarItem-variantError', { timeout: 0 }).should('not.exist');
+};
+
+export function getProjectInfo(projectId: number) {
+  return new Cypress.Promise<ProjectInfo>((resolve) =>
+    Promise.all([
+      v2apiFetchPromise(`projects/${projectId}`).then((r) => r.body),
+      v2apiFetchPromise(`projects/${projectId}/languages`).then(
+        (r_1) => r_1.body
```
JanCizmar, line 23: Are we using snake case, now? 😀

webapp/src/component/PermissionsSettings/RoleLanguages.tsx
```diff
@@ -1,55 +1,47 @@
...
-      {scopes.includes('translations.state-edit') && (
-        <RoleLanguage
-          scope="translations.state-edit"
-          state={state}
-          onChange={onChange}
-          dependencies={dependencies}
+  return show ? (
+    <StyledContainer>
+      <Box display="grid" gridAutoFlow="row" minWidth="200px">
+        {/* <Typography variant="caption">{getScopeTranslation(scope)}</Typography> */}
```
JanCizmar, line 38: commented

e2e/cypress/common/permissions/main.ts
```diff
@@ -0,0 +1,80 @@
...
+  checkNumberOfMenuItems(Object.keys(settings).length + 1);
```
JanCizmar, line 31: And what if by mistake, there is additional element and missing element in the same time. It should check the items separately.

e2e/cypress/common/permissions/main.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    if (value !== SKIP) {
+      // go to page
+      cy.gcy(item).click();
+      // check if there an error
+      pageIsPermitted();
+
+      switch (item as keyof MenuItem) {
+        case 'project-menu-item-dashboard':
+          testDashboard(projectInfo);
```
JanCizmar, line 43: IMO, the settings should not contain only RUN/SKIP items, but also configuration for each test. For exmaple for the dashboard test, you can specify, which options should be accessible and then you can use it in the specific test case.

JanCizmar (changes requested):

JanCizmar (approved):

PR #1603, opened 2023-03-03
fix: Git line endings on Windows

Martin005:
Force LF on text files for windows with autocrlf=true. This avoids having issues with lint requiring LF endings.

Note that for the change to take place people on windows will need to save their changes and rebuild the repository:

**Running commands below remove all local changes**:
```bash
> git rm --cached -r . 
> git reset --hard
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto eol=lf


JanCizmar (commented):
This setting would disable us from testing specific file formats with crlf -> can lead to confusion in the future. I would let this param unset and rely on lint tools to handle this.

Martin005:
@JanCizmar Hmmm, that's true. In that case we can add another entry to this `.gitattributes` file specifying what EOL should be used for that specific file at that specific path.

But the current issue with line endings in the `webapp` folder with Prettier can be also overcome by setting `"endOfLine": "auto"` in the `.prettierrc.json` config file.

Issue #1271, opened 2022-07-27
Allow use of IAM authentication for AWS translate/S3


kenske:
Currently, the setup for AWS S3 and Translate requires you to add an access key and id for authentication. For self-hosted Tolgee instances running in AWS, authentication can be done through IAM roles/service accounts/instance profiles, so no access key would be needed.

JanCizmar:
Feel free to submit PR. 🤓

jdimeo:
I've begun working on upgrading to AWS SDK v2 so we get the better default credentials chain (just FYI). We need this if we're going to self-host as well.

jdimeo:
PR is under review, still relevant

kenske:
Not stale

PR #1614, opened 2023-03-08
#1271 Update to AWS SDK v2

jdimeo:
This PR attempts to update the AWS SDK to v2 and resolve the need to use the default credentials chain (and thus IAM role) to authenticate as described in #1271 . Not sure how to fully test this change, but just getting it out there for review sooner rather than later.


backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/machineTranslation/AwsMachineTranslationProperties.kt
```diff
@@ -6,6 +6,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 open class AwsMachineTranslationProperties(
   override var defaultEnabled: Boolean = true,
   override var defaultPrimary: Boolean = false,
+  var enabled: Boolean = false,
```
jdimeo, line 9: This is regrettable to change the config file, but now that `accessKey` and `secretKey` are no longer required to be specified, we needed a way to show that this was enabled and a client should be created. Are we able to use `defaultEnabled` instead for this purpose?

backend/data/src/main/kotlin/io/tolgee/MtServicesConfiguration.kt
```diff
@@ -32,18 +32,17 @@ class MtServicesConfiguration(
...
-          BasicAWSCredentials(
-            awsMachineTranslationProperties.accessKey,
-            awsMachineTranslationProperties.secretKey
-          )
-        )
-      ).withRegion(awsMachineTranslationProperties.region).build()
+  fun getAwsTranslationService(): TranslateClient? {
+    val chain = when (awsMachineTranslationProperties.accessKey.isNullOrEmpty()
+                   || awsMachineTranslationProperties.secretKey.isNullOrEmpty()) {
+      true -> DefaultCredentialsProvider.create()
```
JanCizmar, line 38: So how do we authenticate, when we don't provide the credentials? Am I missing something?

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/machineTranslation/AwsMachineTranslationProperties.kt
JanCizmar: `defaultEnabled` has another purpose. When `true` this translation provider is enabled by default, for all the projects.

backend/data/src/main/kotlin/io/tolgee/MtServicesConfiguration.kt
jdimeo: @JanCizmar if you read the Javadoc here, it's pretty comprehensive, trying all manners of authentication in order. This responds to config files, environment variables, IAM roles, EC2 stuff, etc. all without special effort on our part: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/credentials/DefaultCredentialsProvider.html

jdimeo:
I see some tests failed but am not sure how to debug them further.

backend/data/src/main/kotlin/io/tolgee/MtServicesConfiguration.kt
JanCizmar: Oh, I see. I am wondering what is the use case for this. Is it just the simplicity of the configuration or is there also some must have?

jdimeo: Oh it's a huge **must have**. We can't actually move forward with self hosting until we have this. It's a violation of many security policies (including ours) to use access key and secret key now and instead you use delegated roles.

jdimeo:
This is not stale and is very much needed! :-)

JanCizmar:
It's not stale! It's on my todo list!

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/machineTranslation/AwsMachineTranslationProperties.kt
JanCizmar: Let's make the property nullable. If it's null it's going to be enabled when access and secret keys are both set. This way it's not breaking change.

jdimeo:
So sorry @JanCizmar - the whole point of doing this was to use STS web identities to authenticate to AWS inside a container. But I neglected to add the dependency:
https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-sts/1.12.483

However, do we want to add it by default? or should users "opt in"? I'm guessing the hassle of getting it into the Docker image after the fact means that we should add it now?

JanCizmar:
I don't understand how this thing should work. What's the use case? What's the scenario?

jdimeo:
So now that we're using v2 of the SDK, AWS looks at many different ways of authenticating: one is using the secret keys that we had before, another is looking at environment variables, profiles, etc. as I linked before:
image

This gives users full flexibility to configure AWS according to best practices, best security, and how their organization is set up.

STS web identities is the common way to do this from inside AWS hosted K8S or Docker containers because then you don't have to create special credentials for those apps that need to be rotated. They can just "assume the role" that they need to access other AWS resources, like S3.

By not having the STS jar on the classpath, it's skipping step 3 in the screenshot above. Therefore for our containers we would have to pursue a more explicit approach to grant access, which means maintenance pain and going against our organization's AWS policies.

That being said, not everyone will use option 3 and will use many of the other options, so that's why it's not strictly "required" to have this .jar and why 3.17.0 isn't necessarily "broken" - getting AWS SDK v2 shipped still brought a lot of other benefit!

JanCizmar:
Oh. I see. OK, no problem just add the required dependency.

jdimeo:
Woud a) you like to do it, b) should I do it here or c) should I do it in a new PR?

JanCizmar:
c) Please :)

kenske:
Thanks for doing this @jdimeo , I'm looking forward to this change as I too would like to use STS.

jdimeo:
@JanCizmar created: https://github.com/tolgee/tolgee-platform/pull/1746 thank you!

PR #1625, opened 2023-03-10
fix: Add state to oauth2 redirect url and validate that

waigel:


waigel:
Can we upgrade typescript to > 4.6.2 to fix this issue?

stepan662:
Yeah, feel free to update it, but also the login e2e test is failing.

stepan662:
The "namespace" one is flaky, so you can ignore that one (it should be fixed in main).

JanCizmar:
It still fails on main time to time, so you can be brave and fix it! 💪 I spend few hour on it, making it less flaky.

PR #1686, opened 2023-04-28
fix(mt): support more languages from baidu

wfjsw:


backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/BaiduTranslationProvider.kt
```diff
@@ -160,5 +273,72 @@ class BaiduTranslationProvider(
...
+    "ky" to "kir",
+    "ca" to "cat",
+    "gl" to "glg",
+    "fr-ca" to "frn",
+    "co" to "cos",
+    "tlh" to "kli",
+    "ks" to "kas",
+    "kn" to "kan",
+    "kw" to "cor",
+    "cr" to "cre",
```
Martin005, line 325: I don't think this mapping is correct.
Looking at Baidu, they say that "cre" is code for language "克里克语". Looking up what "克里克语" means reveals the Muscogee language (Creek).
But you are mapping "cr" to that and "cr" is language code for the Cree language.
I guess it would be best to check with Baidu if their table is correct.

Martin005 (changes requested):

backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/BaiduTranslationProvider.kt
wfjsw: I was mainly referring to https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes because attempting to search the language by their Chinese names would quickly subjects me to the rate limit of any search engines :(

JanCizmar:
So is this ready?

wfjsw:
This is about half of the entire table. I really hope that this can be automated. It is giving me emotional damage just by looking at that :S

JanCizmar:
I hope, you'll be emotionally fine.

PR #1688, opened 2023-04-28
fix(ui): case insensitive autocomplete (#1670)

### LINKED ISSUE #1670: Tag searching is case-sensitive
While a case-sensitive tag is intended, the requirement to search the tag in accurate cases is giving bad experience.

wfjsw:
fix #1670


stepan662:
It's failing on prettier, but other than that it's mergeable

stepan662 (changes requested):
It's failing on prettier, but other than that it's mergeable

stepan662 (approved):

PR #1728, opened 2023-05-22
fix: All languages should be exported by default

JanCizmar:


panther7:
Zdravím, nepozoruji žádnou změnu, že by se stahovali všechny jazyky, včera jsme si o tom psali.

JanCizmar:
It's deployed now. We don't immediately deploy after release. So there was still 3.14.1 on the Cloud instance.

PR #1742, opened 2023-06-03
Documentation for the Unreal Engine integration 

pasotee:


pasotee:
@Martin005 

Hey, thanks for the input. You're right, the endpoint calls can be made without the ProjectId indeed.

However, I was using it for 2 other purposes:
1. Button to launch the tolgee web dashboard directly with the project opened (``https://app.tolgee.io/projects/1840`) - for this one I could also just launch `https://app.tolgee.io/projects` instead and let the user press on it.
2. WebView when hovering over certain texts like `https://app.tolgee.io/projects/1840/translations/single?key=K&ns=NS` which displays a single key entry.

Martin005:
@pasotee
Ah, I see!

Then, you can get the `projectId` from the API key by calling this endpoint [`Returns current API key info`](https://tolgee.io/api#tag/API-keys/operation/getCurrent_1) which returns all kinds of metadata about the API key including the projectId 🙂

pasotee:
@Martin005 I've adjust the Unreal integration to fetch the data based on the API Key here: https://github.com/tolgee/tolgee-unreal/pull/1

And removed the steps from this PR.

The project id is still visible in the settings but not editable and it is fetched when the API Key is inserted. Currently looks like this:

https://github.com/tolgee/tolgee-platform/assets/21221169/a61c2924-9820-40d4-a9f7-9cb09e2459a7

webapp/src/views/projects/integrate/guides/Unreal.mdx
```diff
@@ -0,0 +1,41 @@
...
+After installing the plugin, you need to enable it and set the following properties in the Project Settings → Localization → Tolgee section:
+
+API key:
+```
+{{{apiKey}}}
+```
+
+If you are using the self-hosted version, you can change the URL to your instance.
+
+Endpoint URL:
```
Martin005, line 36:
```suggestion
API URL:
```

PR #1749, opened 2023-06-09
feat: billing plans ui

stepan662:


backend/data/src/main/kotlin/io/tolgee/security/IsServerAdmin.kt
```diff
@@ -0,0 +1,3 @@
+package io.tolgee.security
+
+annotation class IsServerAdmin()
```
JanCizmar, line 3: I would call this rather "NeedsSeverAdmin" or "AccessWithServerAdminPermission".

backend/api/src/main/kotlin/io/tolgee/security/JwtTokenFilter.kt
```diff
@@ -48,6 +50,9 @@ class JwtTokenFilter @Autowired constructor(
             throw PermissionException(Message.EXPIRED_SUPER_JWT_TOKEN)
           }
         }
+        if (this.needsToBeServerAdmin(req)) {
```
JanCizmar, line 53: There should be special test for this. When the annotation stops working it's huge security issue. I don't also think that this is right place to do this. Because being server admin is not directly related to JWT token. You can be server admin while logging in with PAK or PAT and then this whole check might be bypassed.

I actually guess that currently you can call annotated endpoints method with **ANY** PAT token and will be executed

JanCizmar (changes requested):

backend/api/src/main/kotlin/io/tolgee/security/JwtTokenFilter.kt
```diff
@@ -26,7 +27,8 @@ class JwtTokenFilter @Autowired constructor(
   @param:Qualifier("handlerExceptionResolver")
   private val resolver: HandlerExceptionResolver,
   private val requestMappingHandlerMapping: RequestMappingHandlerMapping,
-  private val currentDateProvider: CurrentDateProvider
+  private val currentDateProvider: CurrentDateProvider,
+  private val securityService: SecurityService
```
cyyynthia, line 31: Unused dependency

cyyynthia (commented):
In my opinion the `@AccessWithServerAdminPermission` annotation should always imply the use of a super auth token; though right now it seems all endpoints resulting in data modification are protected with it. Seems reasonable enough for the time being.

stepan662:
@JanCizmar told me you are doing some security enhancements. Could you improve this in that pull request? This was not intended to refactor security stuff, it just kinda happened little bit.

cyyynthia:
The changes I'm planning on making are rather "invasive" since I'm planning on moving certain bits of code around and group all security-related code under a single place for easier auditing and what not, so making the changes in this pull request might go out of scope rather quickly 😅

I think it is reasonable to merge as-is, and I'll make the changes later on while I do some overall sanity checks. (This is also valid for the other PR)

stepan662:
Ok, cool 🙂

PR #1790, opened 2023-07-12
fix: Backend queue & major logging fixes

JanCizmar:


JanCizmar:
Tests can be fixed later. It doesn't look like backend issues.

stepan662:
I'd try to fix it right away, as now we at least know where the issue might be.

PR #1814, opened 2023-07-19
feat: Add telemetry collection

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/component/reporting/BusinessEventPublisher.kt
```diff
@@ -3,30 +3,87 @@ package io.tolgee.component.reporting
...
+      publish(event)
+      cachePublished(key)
+    }
+  }
+
+  private fun cachePublished(key: String) {
+    getEventThrottlingCache()?.put(key, ThrottledEventInCache(currentDateProvider.date.time))
+  }
+
+  private fun shouldPublishOnceInTIme(
```
cyyynthia, line 68: Casing; should be `Time`

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/TelemetryProperties.kt
```diff
@@ -0,0 +1,29 @@
...
+
+We don't collect any other data. Please leave telemetry enabled to help us improve Tolgee.""",
+  displayName = "Telemetry"
+)
+class TelemetryProperties {
+  @DocProperty(description = "Whether telemetry is enabled")
+  var enabled: Boolean = true
+
+  @DocProperty(hidden = true)
+  var server: String = "https://app.tolgee.io"
```
cyyynthia, line 28: This seems to be used only once and never modified. Should it be a constant instead, or is the property here in case E2E tests need them?

JanCizmar: it's modified in tests, search for `tolgee.telemetry.server`

cyyynthia: I must be blind, because I double checked and I don't see it being used besides in the TelemetryService... The tests do modify whether it is enabled and the reporting time in ms, but I couldn't find an instance of `server` being modified/altered 🤔

JanCizmar:
Its in the billing repoOn 20. 7. 2023, at 18:43, Cynthia ***@***.***> wrote:﻿
@cyyynthia commented on this pull request.



In backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/TelemetryProperties.kt:
> +  @DocProperty(hidden = true)
+  var server: String = "https://app.tolgee.io"

I must be blind, because I double checked and I don't see it being used besides in the TelemetryService... The tests do modify whether it is enabled and the reporting time in ms, but I couldn't find an instance of server being modified/altered 🤔

cyyynthia:
Oooooh alright, that makes sense. Got it 👍🏻

cyyynthia (approved):

PR #1849, opened 2023-08-08
fix: Missing crypto.randomUUID()

JanCizmar:


webapp/src/component/security/OAuthService.tsx
```diff
@@ -4,6 +4,7 @@ import GoogleIcon from '@mui/icons-material/Google';
 import LoginIcon from '@mui/icons-material/Login';
 import { LINKS, PARAMS } from 'tg.constants/links';
 import { T } from '@tolgee/react';
+import { v4 as uuidv4 } from 'uuid';
```
stepan662, line 7: We should specify this in package.json.

stepan662 (changes requested):

PR #1879, opened 2023-08-28
Update de.json

paulwer:
"Waffe" means weapon in german and ins therefore misused.

If the change has to be placed within a tolgee-instance, this can be closed and changed there.

Thank you

### DIFF
--- a/webapp/src/i18n/de.json
+++ b/webapp/src/i18n/de.json
@@ -446,7 +446,7 @@
   "import-successful-message" : "Import erfolgreich",
   "import_translations_title" : "Übersetzungen importieren",
   "integrate-api-key-hidden-description" : "<Nur neue Schlüssel können angezeigt werden>",
-  "integrate_choose_your_weapon" : "Wählen Sie Ihre Waffe",
+  "integrate_choose_your_weapon" : "Wählen Sie Ihre Plattform",
   "integrate_guides_go_to_docs" : "Zur Dokumentation gehen",
   "integrate-initial-api-key-description-value" : "Mein Integrations-API-Schlüssel",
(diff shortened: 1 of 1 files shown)


paulwer:
Alternatives could be "Framework", which can be used also in German in the same way.
This would then not include libraries, for whole languages like php.

Therefore I guess Plattform is the best fit. (my POV)

JanCizmar:
I think the "weapon" joke works the same in german...

paulwer:
From my POV (german native), I only know this from the english language. 
My opinion: It sounds weird and i was just confused instead of thinking about the english joke.
(funny afterwards, after figuring it out)

JanCizmar:
What do you think @eweren ?

eweren:
I understand your opinion @paulwer. Guess we should go with "Werkzeug" (tool) to transfer the joke a bit from english to german. Or "framework" since it is more explaining - but we will lose the joke with this.

eweren:
Same goes for if you could call rest or php a plattform 🤷‍♂️

eweren:
Eitherway... with Wekrzeug we shouldn't have a problem

paulwer:
seems the way to go :)

JanCizmar:
I've fixed in the Tolgee app, so it will be released with next release. Anyway, if you would like to contribute to german translation, I can invite you to the Tolgee platform. We've translated the strings using the Tolgee Translator, so it might need a review.

paulwer:
Good Idea.
Do you need anything from me?

JanCizmar:
I've sent you an link via the community slack. :)

JanCizmar:
I am closing this, since it's fixed and released already. Thanks to all! 🚀

PR #1899, opened 2023-09-07
fix: null errors in activity and translation view

stepan662:


JanCizmar (approved):
Thats weird... It should be available even when deleted

PR #1903, opened 2023-09-11
fix: language settings fixes

JanCizmar:


stepan662 (commented):
Please add providers sorting, so that Tolgee is first by default

PR #1915, opened 2023-09-25
feat: onboarding with demo project

stepan662:


e2e/cypress/e2e/translations/mtSettings.cy.ts
```diff
@@ -24,7 +24,7 @@ describe('Machine translation settings', () => {
     }
   });
 
-  it('will update default settings', () => {
+  it('will update default settings', { retries: 2 }, () => {
```
JanCizmar, line 27: Is this the retries 2 intended? If it's really fleaky and it's unrsolvable, I would allow 5 retrials, otherwise, let's remove it.

e2e/cypress/e2e/translations/mtSettings.cy.ts
```diff
@@ -55,7 +55,7 @@ describe('Machine translation settings', () => {
     );
   });
 
-  it('will update language specific settings', () => {
+  it('will update language specific settings', { retries: 2 }, () => {
```
JanCizmar, line 58: Is this the retries 2 intended? If it's really fleaky and it's unrsolvable, I would allow 5 retrials, otherwise, let's remove it.

webapp/src/component/layout/DashboardPage.tsx
```diff
@@ -54,8 +57,15 @@ export const DashboardPage: FunctionComponent<Props> = ({
 
   const rightPanelWidth = useGlobalContext((c) => c.rightPanelWidth);
 
-  const quickStartGuideOpen = useGlobalContext((c) =>
-    Boolean(c.quickStartGuide.open && c.userInfo)
+  const { quickStartSetOpen } = useGlobalActions();
```
JanCizmar, line 60: I guess the convention is to name setters starting "set"

e2e/cypress/e2e/translations/mtSettings.cy.ts
stepan662: It is the issue with sockets because I'm checking the auto translation is set correctly. I can set it to 5 if you like

webapp/src/component/layout/DashboardPage.tsx
stepan662: Ok, might be so

PR #1946, opened 2023-10-13
feat: switch to roboto font

stepan662:


JanCizmar:
So we are sure, we really want to do this?

stepan662:
I don't know

PR #1952, opened 2023-10-16
feat: add danish locale

stepan662:


webapp/package.json
```diff
@@ -74,7 +74,7 @@
     "tsc": "tsc",
     "unused-exports": "ts-unused-exports tsconfig.json --findCompletelyUnusedFiles --ignoreFiles=GoToDocsButton",
     "tsc:prod": "tsc --project tsconfig.prod.json",
-    "load-translations": "tolgee pull ./src/i18n -o",
+    "load-translations": "tolgee pull ./public/i18n -o",
```
JanCizmar, line 77: I don't think we need to store it in public, when we are importing the data using `import`. This way, it will be included two times for no reason - once as a generated chunk, once as a static asset in public dir.

JanCizmar (changes requested):

webapp/package.json
stepan662: I've moved it to public, so it can be loaded with `BackendFetch` plugin, so we don't have to list every single language in static data. It won't be loaded twice.

JanCizmar: Well in that case, it can never be cached. Are you sure you want to do that?

stepan662: Hmm, yeah you are right. It's probably better in static data. I just didn't like the duplication, in this case is worth it.

JanCizmar (approved):

PR #1956, opened 2023-10-16
fix(oauth2): generate valid auth url when auth url already contains query params

ammmze:
Currently, if you use an authorization url such as `https://auth.example.com/oauth2/auth?foo=bar`, then the generated url when authenticating becomes `https://auth.example.com/oauth2/auth?foo=bar?client_id=...` instead of `https://auth.example.com/oauth2/auth?foo=bar&client_id=...`. By using the browsers URL api, we can let it figure that out and we can set the parameters on top of it. In addition, the `toString()` call on `URL` will also ensure the url is encoded, so the `encodeURI` method call is no longer necessary.

Some things to note as a result of this change:
* The spaces `scope` parameter will now be encoded as `%20` rather than a `+`. But in general, the `%20` is the safer option and is covered by the RFC spec.
* Each of the query params will now be properly url encoded, so if somebody did hack around this issue previously by injecting parameters through something like the scope, or client id, etc (say perhaps by something like `tolgee.authentication.oauth2.scopes=openid email profile&foo=bar` or `tolgee.authentication.oauth2.client-id=my-client-id&foo=bar` producing something akin to `https://example.com?scope=openid+email+profile&foo=bar`) will no longer work (they will now produce something akin to `https://example.com?scope=openid%20email%20profile%26foo%3Dbar`.

### DIFF
--- a/webapp/src/component/security/OAuthService.tsx
+++ b/webapp/src/component/security/OAuthService.tsx
@@ -58,13 +58,15 @@ export const oauth2Service = (
   const redirectUri = LINKS.OAUTH_RESPONSE.buildWithOrigin({
     [PARAMS.SERVICE_TYPE]: 'oauth2',
   });
+  const authUrl = new URL(authorizationUrl);
+  authUrl.searchParams.set('client_id', clientId);
+  authUrl.searchParams.set('redirect_uri', redirectUri);
+  authUrl.searchParams.set('response_type', 'code');
+  authUrl.searchParams.set('scope', scopes.join(' '));
+  authUrl.searchParams.set('state', state);
   return {
     id: 'oauth2',
-    authenticationUrl: encodeURI(
-      `${authorizationUrl}?client_id=${clientId}&redirect_uri=${redirectUri}&response_type=code&scope=${scopes
-        .map((scope) => `${scope}`)
-        .join('+')}&state=${state}`
-    ),
+    authenticationUrl: authUrl.toString(),
     buttonIcon: <LoginIcon />,
     loginButtonTitle: <T keyName="login_oauth2_login_button" />,
     signUpButtonTitle: <T keyName="login_oauth2_signup_button" />,


webapp/src/component/security/OAuthService.tsx
```diff
@@ -58,13 +58,15 @@ export const oauth2Service = (
   const redirectUri = LINKS.OAUTH_RESPONSE.buildWithOrigin({
     [PARAMS.SERVICE_TYPE]: 'oauth2',
   });
+  const authUrl = new URL(authorizationUrl);
+  authUrl.searchParams.set('client_id', clientId);
+  authUrl.searchParams.set('redirect_uri', redirectUri);
+  authUrl.searchParams.set('response_type', 'code');
+  authUrl.searchParams.set('scope', scopes.map((scope) => `${scope}`).join('+'));
```
ammmze, line 65: Actually i need to look at this line a little bit ... because using the `+` here we are more/less manually doing an encode on this parameter, so when `URL.toString` runs those are then getting encoded as `%2B` (a literal plus rather than a url encoded space)

ammmze: I've updated it to `.join(' ')`, so it will get encoded properly as a space. Note: It will be encoded as a `%20` instead of a `+`. But technically `%20` is preferred from a compatibility stand point.

JanCizmar:
Thanks a lot for the PR! 🚀

cyyynthia (approved):
LGTM

webapp/src/component/security/OAuthService.tsx
```diff
@@ -58,13 +58,18 @@ export const oauth2Service = (
   const redirectUri = LINKS.OAUTH_RESPONSE.buildWithOrigin({
     [PARAMS.SERVICE_TYPE]: 'oauth2',
   });
+  const authUrl = new URL(authorizationUrl);
+  authUrl.searchParams.set('client_id', clientId);
+  authUrl.searchParams.set('redirect_uri', redirectUri);
+  authUrl.searchParams.set('response_type', 'code');
+  authUrl.searchParams.set(
+    'scope',
+    scopes.map((scope) => `${scope}`).join('·')
```
cyyynthia, line 67: This `.map` looks a bit weird, I don't think it's necessary. But the real problem is with the join, there seem to be an unexpected character `·` instead of a plain space.

ammmze: Oops...missed that one when copying the format from the prettier output in the earlier build (because I was too lazy to install dependencies and run prettier locally 😬).

As far as having the map at all, I only left it because we had it in there from before, but I'm happy to take it out if you want and just make it a `scopes.join(' ')`.

ammmze: Updates have been made, and I ran prettier locally this time ... 3rd times the charm 😄

ammmze:
I saw there were failing tests (triggering timeouts). Not sure they were related. I just rebased my changes and pushed it up to ensure i've got the latest changes.

cyyynthia:
Tests sometimes are a bit capricious 😓 But all green this time! @JanCizmar this is ready for merge 😄

PR #1966, opened 2023-10-21
feat: Disabling languages for key

JanCizmar:


cyyynthia (commented):
Couldn't the setting of disabled key be done with just two SQL request? While everything is wrapped in a transaction I'm still concerned doing the foreach at application level causes a lot of network roundtrip with the database which might not be very efficient compared to just letting the database figure out what to update 🤔 

```
UPDATE translations SET state = "UNTRANSLATED" WHERE id NOT IN ?;
UPDATE translations SET state = "DISABLED", outdated = false, (...) WHERE id IN ?;
```

Though I could understand some requirement to do it this way to receive events that this data was updated; although it'd be quite unfortunate to do N+1 queries for that in terms of efficiency 😕

backend/app/src/test/resources/application.yaml
```diff
@@ -4,7 +4,7 @@ spring:
       - org.redisson.spring.starter.RedissonAutoConfiguration
       - org.springframework.boot.autoconfigure.data.redis.RedisAutoConfiguration
   jpa:
-#    show-sql: true
+    show-sql: true
```
stepan662, line 7: Is this intentional?

backend/data/src/main/kotlin/io/tolgee/service/key/KeyService.kt
```diff
@@ -273,4 +273,41 @@ class KeyService(
...
+    val currentlyDisabledLanguages = keyRepository.getDisabledLanguages(projectId, key.id)
+    val languagesToEnable = currentlyDisabledLanguages.filter { !languageIdsToDisable.contains(it.id) }
+    languagesToEnable.forEach { language ->
+      val translation = translationService.find(key, language).orElse(null) ?: return@forEach
+      translation.clear()
+      translationService.save(translation)
+    }
+  }
+
+  private fun disableLanguagesToDisable(
```
stepan662, line 299: Funny name

backend/data/src/main/kotlin/io/tolgee/service/key/KeyService.kt
```diff
@@ -273,4 +273,41 @@ class KeyService(
...
+  }
+
+  @Transactional
+  fun setDisabledLanguages(projectId: Long, keyId: Long, languageIds: List<Long>): List<Language> {
+    val key = keyRepository.findByProjectIdAndId(projectId, keyId) ?: throw NotFoundException()
+    enableRestOdLanguages(projectId, languageIds, key)
+    return disableLanguagesToDisable(projectId, languageIds, key)
+  }
+
+  private fun enableRestOdLanguages(projectId: Long, languageIdsToDisable: List<Long>, key: Key) {
```
stepan662, line 289: Typo in method name

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
```diff
@@ -17,4 +17,5 @@ interface LanguageRepository : JpaRepository<Language, Long> {
   fun findAllByTagInAndProjectId(abbreviation: Collection<String?>?, projectId: Long?): List<Language>
   fun deleteAllByProjectId(projectId: Long?)
   fun countByIdInAndProjectId(languageIds: Set<Long>, projectId: Long): Int
+  fun findAllByProjectIdAndIdInOrderById(projectId: Long, languageIds: List<Long>): List<Language>
```
stepan662, line 20: 3x Id is too much IMHO

stepan662:
@cyyynthia there is a problem with the activity events as you write. We'll leave it as it is as there is no expectation of a high usage of this feature.

stepan662 (approved):

PR #1971, opened 2023-10-25
fix: Allow API access to specific project management endpoints

JanCizmar:


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2ProjectsController.kt
```diff
@@ -96,6 +96,7 @@ class V2ProjectsController(
   @Operation(summary = "Returns all projects where current user has any permission")
   @GetMapping("", produces = [MediaTypes.HAL_JSON_VALUE])
   @IsGlobalRoute
+  @AllowApiAccess
```
cyyynthia, line 99: This should be marked with `AuthTokenType.ONLY_PAT` otherwise any PAK would be allowed to access the endpoint despite not being project-specific.

cyyynthia (approved):

PR #1972, opened 2023-10-25
feat: disabled translations

stepan662:


backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MtService.kt
```diff
@@ -93,7 +94,18 @@ class MtService(
     publishBeforeEvent(project)
 
     checkTextLength(baseTranslationText)
-    val targetLanguageIds = targetLanguages.map { it.id }
+
+    // filter only translations that are not disabled
+    val targetLanguageIds = keyId?.let { kId ->
```
JanCizmar, line 99: I don't think this is a right place to do this. The purpose of this method is only to return the machine translations, not to handle disabled translations. Also, performance will suffer here, since you are retrieving key and all it's translations for every element in batch. 

The right place to do this would be `io.tolgee.service.translation.AutoTranslationService#softAutoTranslate` and `io.tolgee.service.translation.AutoTranslationService#autoTranslateSync`

This solution doesn't fix transaltion via translation memory.

backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationService.kt
```diff
@@ -372,9 +372,9 @@ class TranslationService(
   }
 
   @Transactional
-  fun setState(keyIds: List<Long>, languageIds: List<Long>, state: TranslationState) {
+  fun setStateBatch(keyIds: List<Long>, languageIds: List<Long>, state: TranslationState) {
     val translations = getTargetTranslations(keyIds, languageIds)
-    translations.forEach { it.state = state }
+    translations.filter { it.state !== TranslationState.DISABLED }.forEach { it.state = state }
```
JanCizmar, line 377: The filtering can be done by database. Also !== is identity check. In this case it will work, but normaly it would be source of bugs.

JanCizmar (changes requested):

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/BigMetaController.kt
```diff
@@ -17,21 +28,42 @@ import javax.validation.Valid
...
   private val projectHolder: ProjectHolder,
+  private val keyService: KeyService,
+  private val keyWithBaseTranslationModelAssembler: KeyWithBaseTranslationModelAssembler,
+  private val projectService: ProjectService,
+  private val translationService: TranslationService
 ) {
   @PostMapping("/big-meta")
   @Operation(summary = "Stores a bigMeta for a project")
-  @RequiresProjectPermissions([ Scope.TRANSLATIONS_EDIT ])
+  @RequiresProjectPermissions([Scope.TRANSLATIONS_EDIT])
```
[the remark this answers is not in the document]
stepan662: The formatting is correct IMHO, this happened when I applied the formatting. Or what do you mean?

JanCizmar (approved):

PR #1979, opened 2023-10-30
Make /v2/projects API accessible via PAT

### LINKED ISSUE #1973: Create project via API
We want to automatically create projects via API, as we have another master/leading system owning the projects. For ease of use we want to create the Tolgee projects automatically.

Currently, the `POST /v2/projects` endpoint can only be used with a bearer token. Can we enable this API to be accessible via personal access token (PAT)?

To my understanding this only requires an annotation on the request method in `backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2ProjectsController.kt`. Is this correct?

raman-nbg:
This PR resolves #1973


JanCizmar (approved):

JanCizmar:
It's merged! Thanks a lot 🚀

PR #1989, opened 2023-11-17
feat: Content delivery front-end

stepan662:


webapp/src/constants/links.tsx
```diff
@@ -322,4 +322,14 @@ export class LINKS {
   );
 
   static PROJECT_IMPORT = Link.ofParent(LINKS.PROJECT, 'import');
+
+  static PROJECT_DEVELOPER = Link.ofParent(LINKS.PROJECT, 'developer');
```
JanCizmar, line 326: I was thinking that we should probably add the `content delivery configs` path as well, although its a default path for developer, since it's semantic. For example if we want to link there from the documentation, feature changing of the default for developers path (possibly because of adding some cool developer dashboard or something) would break such link.

webapp/src/views/projects/developer/storage/StorageFormAzure.tsx
```diff
@@ -0,0 +1,135 @@
...
+import * as Yup from 'yup';
+
+import { TextField } from 'tg.component/common/form/fields/TextField';
+import { components } from 'tg.service/apiSchema.generated';
+import { StorageFormActions } from './StorageFormActions';
+
+type ContentStorageRequest = components['schemas']['ContentStorageRequest'];
+type ContentStorageModel = components['schemas']['ContentStorageModel'];
+
+const formSchemaNew = Yup.object({
```
JanCizmar, line 13: This should go to `GlobalValidationSchema.tsx`

webapp/src/views/projects/developer/storage/StorageFormAzure.tsx
```diff
@@ -0,0 +1,135 @@
...
+  name: Yup.string().required(),
```
JanCizmar, line 14: I totally forgot to limit the lengths.

JanCizmar (approved):

webapp/src/views/projects/developer/storage/StorageFormAzure.tsx
JanCizmar: I've added max length for name to 100 and everything else to 255.

webapp/src/constants/links.tsx
stepan662: 👍

PR #2014, opened 2023-12-08
fix: XLIFF export is now compliant with XLIFF1.2 Schema (XSD)

GeertZondervan:
Fix for #2015

The XLIFF generated by the exporter is not valid according to the official XLIFF 1.2 Schema (XSD).
The XLIFF schema can be found here:
https://docs.oasis-open.org/xliff/v1.2/os/xliff-core-1.2-transitional.xsd

XLIFF 1.2 Documentation: https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html

Currently the output from Tolgee looks like this:
```
<?xml version="1.0" encoding="UTF-8"?>

<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
  <file xmlns="" datatype="plaintext" source-language="en" target-language="en">
    <body>
      <trans-unit id="Z key" datatype="html">
        <source>A translation</source>
        <target>A translation</target>
      </trans-unit>
    </body>
  </file>
</xliff>
```
When validating this output against the XSD, it will show a couple of errors. You can use any online validator for this, but the JDK validator also confirms this
- The `file` element has an empty namespace (xmlns) attribute. This attribute should not be there.
- The `file` element is missing an `original` attribute.

The XLIFF 1.2 documentations shows an example of what it should look like:
```
<xliff version='1.2' xmlns='urn:oasis:names:tc:xliff:document:1.2'>
   <file original='hello.txt' source-language='en' target-language='fr' datatype='plaintext'>
    <body>
    </body>
   </file>
</xliff>
```

This can easily fixed by adding the following to the `XliffFileExporter` class in the `createBaseDocumentStructure` method.

```
...
.addElement("file", "urn:oasis:names:tc:xliff:document:1.2")
.addAttribute("original", "undefined")
...
```
which results in the following output:
```
<?xml version="1.0" encoding="UTF-8"?>

<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
  <file original="undefined" datatype="plaintext" source-language="en" target-language="en">
    <body>
      <trans-unit id="Z key" datatype="html">
        <source>A translation</source>
        <target>A translation</target>
      </trans-unit>
    </body>
  </file>
</xliff>
```


JanCizmar (approved):

JanCizmar:
You just have to run `./graldew ktLintFormat` to fix this.

PR #2036, opened 2023-12-20
fix: Onboarding & other UI fixes, security improvements

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/dtos/request/UserMfaRecoveryRequestDto.kt
```diff
@@ -7,6 +7,6 @@ import jakarta.validation.constraints.Size
 @JsonIgnoreProperties(ignoreUnknown = true)
 data class UserMfaRecoveryRequestDto(
   @field:NotBlank
-  @field:Size(max = 50)
+  @field:Size(max = 200)
```
cyyynthia, line 10: That's bad; the BCrypt algorithm has a limit before which the password gets truncated. This limit is specified to be 56 bytes, although a lot of implementations have different interpretations of the full wording used. The safe take is that any decent BCrypt implementation supports password up to 50 characters.

Any character past this is effectively ignored; assuming a string `a` of ~50 characters, `a + "b"`  would be a "valid password" in the eyes of BCrypt, the extra byte being ignored.

backend/data/src/main/kotlin/io/tolgee/service/security/UserAccountService.kt
```diff
@@ -92,42 +91,22 @@ class UserAccountService(
...
   @CacheEvict(cacheNames = [Caches.USER_ACCOUNTS], key = "#result.id")
+  @Transactional
   fun createUser(userAccount: UserAccount): UserAccount {
     userAccountRepository.saveAndFlush(userAccount)
     applicationEventPublisher.publishEvent(OnUserCreated(this, userAccount))
     applicationEventPublisher.publishEvent(OnUserCountChanged(this))
     return userAccount
   }
 
+  @CacheEvict(cacheNames = [Caches.USER_ACCOUNTS], key = "#result.id")
```
cyyynthia, line 102: Any valid turn of events would make this path a no-op. Unsure it is necessary to include.

I'm also unsure if the entity will automatically be mutated to have the correct ID; I recall the JPA spec specifying the `save` method and others may return a completely new object which should be used for further operations on the entity. *Even if it does mutate the original object, it is not something mandated by the spec and subject to one day explode out of nowhere*

webapp/src/constants/GlobalValidationSchema.tsx
```diff
@@ -41,18 +44,13 @@ Yup.setLocale({
...
-      .required(),
-  };
-
-  static readonly USER_PASSWORD_WITH_REPEAT = Yup.object().shape(
-    Validation.USER_PASSWORD_WITH_REPEAT_NAKED
-  );
+  static readonly USER_PASSWORD = (t: TFunType) =>
+    Yup.string()
+      .min(8)
+      .max(200)
```
cyyynthia, line 50: Same as on the backend, >50 does not make sense for the BCrypt algorithm

webapp/src/component/security/ResetPasswordSetView.tsx
```diff
@@ -16,14 +16,13 @@ import LoadingButton from 'tg.component/common/form/LoadingButton';
...
 import { DashboardPage } from '../layout/DashboardPage';
-import { SetPasswordFields } from './SetPasswordFields';
+import { NewPasswordLabel, SetPasswordField } from './SetPasswordField';
 import { useLogout } from 'tg.hooks/useLogout';
 
 const globalActions = container.resolve(GlobalActions);
 
 type ValueType = {
   password: string;
-  passwordRepeat: string;
```
cyyynthia, line 26: Why was this removed? It generally is common practice to have a double check measure to ensure there are no accidental typos in the password (if the user types their passwords manually).

I do know there's a whole debate about that within the UX community, that a single field is less work and therefore better for the user; but another dimension of UX is being helpful to the user and avoid letting them do mistakes and help correct them.

backend/data/src/main/kotlin/io/tolgee/service/InvitationService.kt
```diff
@@ -231,4 +234,15 @@ class InvitationService @Autowired constructor(
   fun userOrInvitationWithEmailExists(email: String, organization: Organization): Boolean {
     return invitationRepository.countByUserOrInvitationWithEmailAndOrganization(email, organization) > 0
   }
+
+  @Transactional
+  fun getInvitationOnRegistration(invitationCode: String?): Invitation? {
+    if (invitationCode == null) {
+      if (!tolgeeProperties.authentication.registrationsAllowed) {
+        throw AuthenticationException(Message.REGISTRATIONS_NOT_ALLOWED)
+      }
```
cyyynthia, line 243: I'm not a fan of this, the code flow is quite obscure here. It's not trivial when observing the sign up function to confirm this check is performed, and I wouldn't consider this function to be responsible for this type of check.

This function receiving a nullable argument to perform this check feels wrong to me as well, you shouldn't encounter cases where "get null" is a valid option and is indicative of a code smell IMHO.

Maybe the naming is just not right? Although I'd consider checking if signing up is allowed something of the concern of the sign up service...

webapp/src/constants/GlobalValidationSchema.tsx
```diff
@@ -41,18 +44,13 @@ Yup.setLocale({
...
+      .required()
+      .minSymbols(1, t('password_validation_min_symbols', { count: 1 }))
+      .minNumbers(1, t('password_validation_min_numbers', { count: 1 }));
```
cyyynthia, line 53: It might be desirable to introduce a check for this in the backend as well; not necessary but doesn't do harm.

cyyynthia (changes requested):

webapp/src/component/security/ResetPasswordSetView.tsx
JanCizmar: Users don't type passwords manually anymore. Repeating passwords is deprecated. You can always reset it if you make a mistake.

backend/data/src/main/kotlin/io/tolgee/dtos/request/UserMfaRecoveryRequestDto.kt
JanCizmar: Completely forgot about this.

backend/data/src/main/kotlin/io/tolgee/service/security/UserAccountService.kt
JanCizmar: Hmm, do we actually need to evict the cache there? We can remove it from the creation methods, right?

webapp/src/constants/GlobalValidationSchema.tsx
JanCizmar: Mybe later. 🤘

stepan662 (approved):

cyyynthia (approved):

PR #2044, opened 2023-12-25
fix: Import improvements

JanCizmar:


webapp/src/views/projects/import/ImportView.tsx
```diff
@@ -83,6 +88,7 @@ export const ImportView: FunctionComponent = () => {
         ],
       ]}
       maxWidth="wide"
+      overflow="scroll"
```
stepan662, line 91: I think this makes the scrollbar always visible on windows even if not necessary, is it intended? I usually use "auto".

stepan662 (approved):

Issue #2042, opened 2023-12-22
Allow to import properties file


benallard:
In the java world, it is pretty common to store the various translations as properties file (resource bundle). It would be great to be able to import them without having to convert them to json first.

PR #2050, opened 2023-12-28
feat: Add simple properties file importer

benallard:
See #2042


JanCizmar:
Hey! Thanks a lot for the PR! It would be great to have some tests and add the support also to front-end? Are you open to do it as well? 🚀

benallard:
By front-end support, you mean extending the text on the page ?

And about the tests, can you point me to some related ones that can inspire me?

Thanks.

JanCizmar:
> By front-end support, you mean extending the text on the page ?

Yes. I forgot, there is actually no front-end validation for selected files extension. So extending the text should do it. Which is hard coded in Tolgee (unfortunately), so I will fix it later. 

> And about the tests, can you point me to some related ones that can inspire me?

io.tolgee.unit.service.dataImport.processors.processors.po.PoParserTest

This one should do it.

benallard:
Both points have been addressed. Thanks!

JanCizmar (approved):

PR #2113, opened 2024-01-29
fix: Current permissions -> PAK with no-auth mode 

JanCizmar:


backend/security/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFilter.kt
```diff
@@ -91,6 +92,16 @@ class AuthenticationFilter(
       // Attempt PAK auth even if it doesn't have the prefix
       // Might be a legacy key
       pakAuth(apiKey)
+      return
+    }
+
```
cyyynthia, line 97: Maybe a comment describing the reason why this is at the bottom to allow PAK auth to "work" (be emulated?) when using no-auth is a good idea, so no one ends up wondering why this is here 😅

cyyynthia (approved):

PR #2142, opened 2024-02-26
feat: Single step import & export enhancements for CLI

JanCizmar:


JanCizmar:
Requirements for the CLI:

Flow with Tolgee SDK:

 - adding a new key with tag "draft" (or "draft-feature-name") (part of SDK)
 - when releasing, find keys in the code
   - mark all existing as "production" and remove "draft"
   - mark all that were "production" before and are not now as "deprecated" (or "deprecated-in-X.X.X")

Flow without Tolgee SDK:
 - push english translations - all non-existant keys marked as "draft"
 - ... rest is the same

```
Possible syntax

$ tolgee push ... --tag-new "draft"
$ tolgee tag --filter "extracted" --add "production" --remove "draft"
$ tolgee tag --filter "tag:production & not:extracted" --add "deprecated"
```

stepan662:
Issue with Apple "STRINGS" format on single-step-import, not accepting the file which was previously exported from the platform (Throwing 500).

Also it would be nice, if we'd merged "STRINGS" and "STRINGSDICT" into "APPLE_STRINGS_STRINGSDICT" as it is in export, so it can be defined on one place in the config.

stepan662:
Seems like `filterTagNotIn` works somehow weirdly - it exports only the namespace, where the tag is and doesn't filter by it.

stepan662 (approved):

stepan662:
For the tag command to work, it requires a new parameter `filterKeysNot`, which will select oposite of listed keys. That way, we can find deprecated keys like this:
```sh
tolgee tag --filter-not-extracted --filter-tag production --tag-filtered deprecated
```

There should also be possibility to mark newly created keys as draft, for which we need parameter in the `single-step-import` - something like `tagCreatedKeys`, with list of tags that should be added to newly created keys.

PR #2162, opened 2024-03-08
feat: Slack integration

huglx:


JanCizmar:
Good job 🚀 Just few comments: 

-  Naming stuff: don’t repeat the “slack” or “tolgee” everywhere 
    - please, think more about how you name stuff, other developers will have to read it in the future
- NEVER add any secrets to the code, not even for PFCs. Add it only to the files contained in .gitignore
- Don’t use jacksonObjectMapper directly, use objectMapper bean when possible. It’s expensive operation to create new mapper instance
- [ ] Rename `/v2/slack/events` to `/v2/slack`
- [ ] Move `/v2/slack/events/tolgee` to `/v2/slack`
- [ ] Rename `/v2/slack/events/event` to `/v2/slack/on-event`
- [ ] Validate then the `slashCommand`, `connectSlack`, `fetchEvent` endpoints are really executed by slack using the signature, so we are sure that slack have send them and not some attacker
- [ ] This is ugly:
     image
    We want to show the login link to the user directly, we dont want to show them ugly ❌. We want to talk with the user nice way, reduce the amount of errors, increase the amount of ways how to achieve their goals.
- [ ] Move all the messages rendered to the I18n_en.properties file I created in the PR
- [ ] This is ugly too. Add some positive message.
       image
- [ ] When user's provides invalid command, don't reply `Invalid command`, but show him nice list of commands they can use 
  - This is why it would make sense to use some library for this, so it's simply extendable
- When adding translatable texts on frontend, always use t method `useTranslate` or Tolgee T component to translate the strings
- [ ] Add `/tolgee logout` to disconnect Slack
- [ ] @ZuzanaOdstrcilova will prepare a UI proposal for the connection dialog in the UI, @stepan662, can you please implement it then?
- [ ] @ZuzanaOdstrcilova @stepan662, can you please add Slack Integration into UI, so user can disconnect slack account from the App & review the UI code?
- [ ] When I run the unsubscribe multiple times, it doesn't warn me that I wasn't subscribed
- [ ] It doesn't show me german inputs when I update the base (`en` in my case) string or create a new key. I am only subscribed to `de`, but I am still getting only English inputs. Only when I actually change the german `de` string I get the desired input.
- [ ] Would be great if we provided the first translation memory suggestion and the machine translation suggestions (these should be explicitely enabled by some property)
- [ ] I would like to get list of my active subscriptions using `/tolgee subscriptions`
- [ ] Globally subscribing in public channel does't post any messages at all
- [ ] @ZuzanaOdstrcilova prepares some light proposal on the UI of the slack messages

JanCizmar (approved):

JanCizmar:
What a nice request!

backend/data/src/main/kotlin/io/tolgee/component/automations/processors/slackIntegration/SlackSubscriptionProcessor.kt
```diff
@@ -0,0 +1,60 @@
...
+        onlyCountInListAbove = 10,
+      ).get() ?: return
+
+    val activityModel = activityModelAssembler.toModel(view)
+
+    val data = SlackRequest(activityData = activityModel)
+    val config = action.slackConfig ?: return
+    slackExecutor.setHelper(data = data, slackConfig = config)
+
+    when (activityModel.type) {
```
JanCizmar, line 40: These types can modify relevant data:
```
  UNKNOWN,
  SET_TRANSLATIONS,
  CREATE_KEY,
  COMPLEX_EDIT,
  IMPORT,
  BATCH_PRE_TRANSLATE_BY_TM,
  BATCH_MACHINE_TRANSLATE,
  AUTO_TRANSLATE,
  BATCH_CLEAR_TRANSLATIONS,
  BATCH_COPY_TRANSLATIONS,
  AUTOMATION, // possibly, since it's generic
```

backend/data/src/main/kotlin/io/tolgee/service/slackIntegration/SavedSlackMessageService.kt
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  fun update(
+    id: Long,
+    langTags: Set<String>,
+  ): SavedSlackMessage? {
+    val savedMessage = savedSlackMessageRepository.findById(id).orElse(null) ?: return null
+    savedMessage.langTags = langTags
+
+    // Сохраняем изменения
```
JanCizmar, line 33: Please, only add comments in english.

backend/data/src/main/kotlin/io/tolgee/service/slackIntegration/SavedSlackMessageService.kt
```diff
@@ -0,0 +1,73 @@
...
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.util.*
+
+@Service
+class SavedSlackMessageService(
+  private val savedSlackMessageRepository: SavedSlackMessageRepository,
+  private val slackConfigRepository: SlackConfigRepository,
+) {
+  fun create(savedSlackMessage: SavedSlackMessage): SavedSlackMessage {
```
JanCizmar, line 17: Usually, the service methods are @Transactional. Please add transactional to the service methods and remove it from controller methods if you don't have good reasons to keep it on controller level.

JanCizmar (changes requested):

PR #2192, opened 2024-03-17
fix: Android Import/Export escaping

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/formats/android/contsants.kt
```diff
@@ -0,0 +1,3 @@
+package io.tolgee.formats.android
+
+const val ANDROID_CDATA_CUSTOM_KEY = "_androidWrapWithCdata"
```
stepan662, line 3: I don't like the underscosre

stepan662 (approved):

backend/data/src/main/kotlin/io/tolgee/formats/android/contsants.kt
JanCizmar: What don't you like about it?

stepan662: It's in section described as "Custom attributes specific for some formats", so it's basically the section exactly for this. The "_" looks hacky.

Snímek obrazovky 2024-03-19 v 15 16 34

JanCizmar: https://tolgee.io/platform/translation_keys/custom_values#tolgee-internal-custom-values

PR #2210, opened 2024-03-22
feat: improve routing and redirects

stepan662:


backend/data/src/main/java/io/tolgee/security/payload/JwtAuthenticationResponse.java
```diff
@@ -20,4 +23,12 @@ public String getTokenType() {
     public void setAccessToken(String accessToken) {
         this.accessToken = accessToken;
     }
+
+    public Long getUserId() {
```
JanCizmar, line 27: This can be removed.

webapp/src/component/layout/RightSidePanel.tsx
```diff
@@ -39,9 +39,20 @@ export const RightSidePanel = ({
...
 }: Props) => {
   const { setRightPanelWidth } = useGlobalActions();
+
   const containerRef = useRef<HTMLDivElement>(null);
-  const topBannerHeight = useGlobalContext((c) => c.topBannerHeight);
-  const topBarHeight = useGlobalContext((c) => c.topBarHeight);
+  const topBannerHeight = useGlobalContext((c) => c.layout.topBannerHeight);
+  const topBarHeight = useGlobalContext((c) => c.layout.topBarHeight);
+
+  // useEffect(() => {
```
JanCizmar, line 47: This is probably some pohrobek.

webapp/src/component/security/ResetPasswordSetView.tsx
```diff
@@ -105,10 +104,23 @@ const PasswordResetSetView: FunctionComponent = () => {
               }
               //@ts-ignore
```
JanCizmar, line 105: Is the @ts-ignore still required?

JanCizmar (changes requested):

Issue #2022, opened 2023-12-14
Allow setting a default namespace for the project


stepan662:

gwBear1:
👍

PR #2214, opened 2024-03-26
feat: Support default namespace

gwBear1:
## Backgorund
- Issue ticket https://github.com/tolgee/tolgee-platform/issues/2022
- Slack discussion https://tolgeecommunity.slack.com/archives/C03A0RQR62H/p1708318014384279
## Change log
- Backend
  - Added a property `baseNamespace` to the `ProjectModel`
- Frontend
  - Made a `useProjectNamespaces` hook for handling namespace related service
  - Support to change `baseNamespace` at the project settings page

## How the baseNamespace works
- If there are no namespaces in the project, they are set by the `<none>` namespace as a default.
- If there are namespaces, you can change the baseNamespace.
- If the baseNamespace (example: which name is "temp") set but all keys that have "temp" namespace are removed, still project setting's baseNamespace  is "temp". Once you set to the other baseNamespace, then, it will be removed from backend.

## Demo
https://github.com/tolgee/tolgee-platform/assets/138835573/a5973000-f0c1-4f14-b44b-6140ebf445d2

## Review Guide
- This is an **INITIAL COMMIT** so feel free to tell me any review.
- **[resolved]** ~~What is `schema.xml`? I added it by hand but felt it should be auto-generated.~~
- **[resolved]** ~~As you can see, It's very similar with `baseLanguage` property which is included `project` query. However, It was not too easy for me to add `baseNamespace` to there because they have a different structure, and I'm not familiar with Kotlin language. Do you guys think, although, adding `baseNamespace` to the project query would be better?~~
- Need to be translated a `project_settings_base_namespace` key but I don't know where should I fix it.


stepan662:
Hey, thanks a lot for the pull request. Great job 🙂
I'd be more in favor of adding the baseLanguage directly to the `ProjectModel` as a relation. I think that it clearly signifies that only one namespace can be default. I think it will also be easier to eliminate the `useProjectNamespaces` hook and just look at the project entity itself - as that is always fetched for the project view.

To update schema.xml it should be possible to just update the model in code and then run `./gradlew diffChangeLog`.

gwBear1:
@stepan662 Thanks for review! I'll address your comment

gwBear1:
Hi @stepan662 I fully update the PR! Can you review this again?

Now, `baseNamespaceId` is included in the project model. One different thing is that I want to leave the `useProjectNamespace` hook because that query is called repeatedly at some codes. It would be good if they are replaced to this hook.

stepan662:
Hey, it would be better to add it as a database relation, like:
```
val baseNamespace: NamespaceModel?
```

What I ment by removing the hook, is that if we do it like this, it can be automatically fetched from database when we are getting the project data and so we will have it automatically available with `useProject` hook on the project object.

PS: could you also please rebase the project - github doesn't allow me to run CI, if there are conflicts in the files.

gwBear1:
@stepan662 Hello again~! I changed it to the `NamespaceId` -> `NamespaceModel` as you mentioned. Would you mind reviewing this again? Of course, I updated PR description as well.

backend/data/src/main/kotlin/io/tolgee/model/Project.kt
```diff
@@ -97,6 +97,10 @@ class Project(
   @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, mappedBy = "project")
   var namespaces: MutableList<Namespace> = mutableListOf()
 
+  @OneToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST])
```
JanCizmar, line 100: Please don't use cascading at all. We are trying to avoid this when possible.

JanCizmar (commented):
There is also failing ktlint. You can fiz this by running `./gradlew ktlintFormat` :)

https://github.com/tolgee/tolgee-platform/wiki/Development#static-analysis

JanCizmar:
Thanks a lot for the PR! :)

backend/data/src/main/kotlin/io/tolgee/model/Project.kt
```diff
@@ -97,6 +97,10 @@ class Project(
...
+  @ActivityLoggedProp
+  var baseNamespace: Namespace? = null
```
stepan662, line 102: Could we name it `defaultNamespace`?

We also need to solve deleting - currently there is a logic that when user stops using the namespace it is automatically deleted. Now when the user selects the namespace as default it won't get deleted (which is correct I guess), but then it should get deleted once the user selects a different default (which is not happening now).

stepan662 (commented):
It works and the code looks great, good job :)
Could you please also run `./gradlew ktlintFormat` on the code, that should fix the kotlin complains.

The backend tests are failing because this is PR from fork, so that should be fine. However we should write some e2e tests, if you feel like writing them on your own we have a wiki how to run them locally:  https://github.com/tolgee/tolgee-platform/wiki/Development

If you don't feel like it, you can leave tests to me.

stepan662:
I've also noticed there is also single key view `/projects/{projectId}/translations/single`, where the default namespace is not applied now.

backend/data/src/main/kotlin/io/tolgee/model/Project.kt
gwBear1: > We also need to solve deleting - currently there is a logic that when user stops using the namespace it is automatically deleted. Now when the user selects the namespace as default it won't get deleted (which is correct I guess), but then it should get deleted once the user selects a different default (which is not happening now).

opps yes you are right. I haven't covered the last case. I'll cover that.

gwBear1:
@stepan662 @JanCizmar Hello again! I addressed your comments.
- applied kotlin linter
- handled edge case
  - removing namespace properly
  - applied default namespace on single creation
- added e2e tests

I'm not a professional in Cypress syntax, so I'm glad if you want to modify my test codes. Thank you!

stepan662:
Ok, seems good. Let me finish the tests and I think it's ready for release.

stepan662:
@gwBear1 Thanks a lot for your contribution, it's merged (#2233). 🎉

gwBear1:
Thanks a lot!

PR #2233, opened 2024-04-05
feat: project default namespace

stepan662:


backend/data/src/main/kotlin/io/tolgee/model/Project.kt
```diff
@@ -97,6 +97,10 @@ class Project(
   @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, mappedBy = "project")
   var namespaces: MutableList<Namespace> = mutableListOf()
 
+  @OneToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST])
+  @ActivityLoggedProp
```
JanCizmar, line 101: When we add new activityLoggedProp, we have to also update the activity showing frontend, otherwise it shows empty action.

JanCizmar (changes requested):

backend/data/src/main/kotlin/io/tolgee/model/Project.kt
stepan662: Hmm, for some reason, I don't get any activity when I update the project.

JanCizmar: Oh, maybe because Project entity is missing the @ActivityLoggedEntity?

PR #2328, opened 2024-05-30
feat: ability to remove other keys when importing

stepan662:


backend/data/src/main/kotlin/io/tolgee/dtos/request/SingleStepImportRequest.kt
```diff
@@ -39,4 +39,9 @@ class SingleStepImportRequest : ImportAddFilesParams(), IImportSettings {
         "It add tags only to new keys. The keys that already exist will not be tagged.",
   )
   var tagNewKeys: List<String> = listOf()
+
+  @get:Schema(
+    description = "Delete keys that were not included in import.",
+  )
+  var removeOtherKeys: Boolean? = false
```
JanCizmar, line 46: This doesn't have to be nullable.

backend/data/src/main/kotlin/io/tolgee/service/dataImport/StoredDataImporter.kt
```diff
@@ -141,6 +146,15 @@ class StoredDataImporter(
     }
   }
 
+  private fun deleteOtherKeys() {
+    (importSettings as? SingleStepImportRequest)?.removeOtherKeys?.let {
```
JanCizmar, line 150: The `let` body will be executed even if the `removeOtherKeys` is `false`. Also currently I am trying to use the `let` more. It adds more structuring depth to the code and it makes it less readable. 

Instead, I would create a variable and do the check separately. 
```
val removeOtherKeys = (importSettings as? SingleStepImportRequest)?.removeOtherKeys == true
if(removeOtherKeys){
...
}
```

JanCizmar (commented):
Cool, but it would be nice to have it tested, this seems like pretty sensitive feature.

backend/data/src/main/kotlin/io/tolgee/dtos/request/SingleStepImportRequest.kt
JanCizmar: OK, I forgot about the springdoc openapi bug.

backend/data/src/main/kotlin/io/tolgee/service/dataImport/StoredDataImporter.kt
JanCizmar: Yes.

JanCizmar (approved):
Nice.

PR #2373, opened 2024-06-17
feat: unique slack connection

huglx:


backend/data/src/main/kotlin/io/tolgee/model/slackIntegration/SlackUserConnection.kt
```diff
@@ -2,14 +2,12 @@ package io.tolgee.model.slackIntegration
...
-import jakarta.persistence.Entity
-import jakarta.persistence.FetchType
-import jakarta.persistence.ManyToOne
-import jakarta.persistence.Table
-import jakarta.persistence.UniqueConstraint
+import jakarta.persistence.*
 
 @Entity
-@Table(uniqueConstraints = [UniqueConstraint(columnNames = ["slack_user_id", "user_account_id"])])
+@Table(
```
JanCizmar, line 8: Well, when I am thinking about this, user might be connected with single Tolgee user account to multiple slack account across multiple slack workspaces. So we should probably add workspaceId column here and add the column to the unique index. I think that way, we would be safe.

JanCizmar (changes requested):

JanCizmar (approved):

JanCizmar:
Good job 👍

PR #2374, opened 2024-06-18
feat: new pricing plans design

stepan662:


webapp/src/fixtures/wrapIf.ts
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+/**
+ * If condition is met, wraps conent with element
+ */
+export function wrapIf(
```
JanCizmar, line 6: Wouldn't this be easier to read if it was React component accepting the content as a body and condition and element as props?

JanCizmar (approved):

PR #2382, opened 2024-06-24
feat: Ability to resend registration email

huglx:
Added: 
- limit the number of the requests per hour to some constant per e-mail and ip
- Enable to the user to login, but without any permissions, telling them, they should verify
- backend endpoint to resend the verification link
Hi @JanCizmar, could you please provide feedback so I can see if I am doing this correctly? Thanks


backend/security/src/main/kotlin/io/tolgee/security/authorization/OrganizationAuthorizationInterceptor.kt
```diff
@@ -52,7 +56,16 @@ class OrganizationAuthorizationInterceptor(
     response: HttpServletResponse,
     handler: HandlerMethod,
   ): Boolean {
-    val userId = authenticationFacade.authenticatedUser.id
+    val user = authenticationFacade.authenticatedUserEntity
```
JanCizmar, line 59: Don't use entity here. We don't want to fetch user entity from the DB with every request. Thats why authenticatedUser is DTO, which can be returned from cache.

backend/security/src/main/kotlin/io/tolgee/security/authorization/OrganizationAuthorizationInterceptor.kt
```diff
@@ -52,7 +56,16 @@ class OrganizationAuthorizationInterceptor(
...
+
+    if (!emailVerificationService.isVerified(
```
JanCizmar, line 61: This code is duplicated, across the codebase. Create some method for this.

backend/data/src/main/kotlin/io/tolgee/service/EmailVerificationService.kt
```diff
@@ -63,6 +66,49 @@ class EmailVerificationService(
...
+    if (policy != null) {
+      rateLimitService.consumeBucketUnless(policy) {
+        createForUser(userAccount, callbackUrl, email)
+        isVerified(userAccount)
+      }
+    }
+  }
+
+  @Transactional
+  fun fakeSend(
```
JanCizmar, line 88: This method us unused.

backend/data/src/main/kotlin/io/tolgee/service/EmailVerificationService.kt
```diff
@@ -63,6 +66,49 @@ class EmailVerificationService(
     return null
   }
 
+  @Transactional
+  fun resendEmailVerification(
```
JanCizmar, line 70: I think, maybe strict `@RateLimited` annotation on the endpoints would be enough and simplier.

backend/data/src/main/kotlin/io/tolgee/security/ratelimit/RateLimitService.kt
```diff
@@ -139,6 +139,23 @@ class RateLimitService(
...
   }
 
+  fun getIEmailVerificationIpRateLimitPolicy(
+    request: HttpServletRequest,
+    email: String?,
+  ): RateLimitPolicy? {
+    if (!rateLimitProperties.emailVerificationRequestLimitEnabled || email.isNullOrEmpty()) return null
+
+    val ip = request.remoteAddr
+    val key = "global.ip.$ip.email.$email::auth"
```
JanCizmar, line 149: So If I resend it from other IP, I can bypass the rate limit? Why so?

backend/data/src/main/kotlin/io/tolgee/service/EmailVerificationService.kt
JanCizmar: Tell me if I am wrong.

JanCizmar (changes requested):

backend/data/src/main/kotlin/io/tolgee/security/ratelimit/RateLimitService.kt
huglx: I did this because I thought that if a user entered the wrong email and sent the verification N times (so he has a rate limit), he could change the email and send the verification more times, but you correctly noticed the problem with different IPs

backend/data/src/main/kotlin/io/tolgee/service/EmailVerificationService.kt
huglx: I think you're right, but I wanted to solve the problem with incorrect email(more in conversation below).

huglx:
I added the FE part. @stepan662, please review it. And I have some questions on the FE part. I added icons in 2 variants (dark mode and light mode), is this a normal way to store multiple icons? And I wrote a part in ApiErrorHandling that handles 429 error, check if it's ok.

I also have a question about the speed limiter. Is it OK that I only store the IP of the user? That way, if the user entered the wrong email and sent too many requests, he has to wait for a while after entering the correct email. On the other hand, someone could change the IP and spam.

webapp/src/svgs/icons/mailDark.svg
stepan662, line 1: Use fill="currentColor" in the icon and then you can set the color from the code rather having two separate icons

webapp/src/views/projects/ProjectListView.tsx
stepan662, line 1: I think this is overly complex, could we have two separate files like `ProjectListView` and `EmailNotVerifiedView`. There is a router rule, that it redirects you to the page `/projects` after login, so you can change that in case the email is not verified. Or if thats too complicated, you can just switch that directly in the router file.

stepan662 (changes requested):

webapp/src/component/layout/TopBanner/TopBanner.tsx
```diff
@@ -17,9 +17,15 @@ const StyledContainer = styled('div')`
   display: grid;
   grid-template-columns: 50px 1fr 50px;
   width: 100%;
-  background: ${({ theme }) => theme.palette.topBanner.background};
+  background: ${({ theme }) =>
+    useIsEmailVerified()
```
stepan662, line 21: This is not a good way how to use hooks. Use it in the component itself and only apply class to change styles.

https://react.dev/reference/rules/rules-of-hooks

webapp/src/i18n/en.json
```diff
@@ -1399,5 +1399,14 @@
...
+  "your_email_was_changed_verification_message" : "When you change your email, new email will be set after its verification.",
+  "announcement_verify_email" : "We have already sent you a verification email.",
+  "didnt_receive_email_hint" : "Didn't receive the email?",
+  "email_already_verified" : "The email is already verified. ",
+  "email_resend_message" : "Your verification link has been resent.",
+  "too_many_requests" : "Too many requests. Please try again later.",
+  "verify_email_description" : "The final step is to verify your email. After verification, you will be able to use the platform.\nPlease check your inbox at <b>{email}</b>.",
+  "verify_email_now_title" : "Verify your email now",
+  "verify_email_resend_button" : "resend verification email",
+  "verify_email_title" : "Verify your email address"
```
stepan662, line 1411: You don't have to edit i18n files manually, use in-context tools and create the translations in the tolgee platform, i18n files are then updated automatically in the release pipeline

webapp/src/service/http/ApiHttpService.tsx
```diff
@@ -59,8 +59,9 @@ export class ApiHttpService {
             if (r.status >= 400) {
               const responseData = await ApiHttpService.getResObject(r);
               const resultError = new ApiError('Api error', responseData);
+              const currentPath = location.pathname;
               resultError.setErrorHandler(() =>
-                handleApiError(r, responseData, init, options)
+                handleApiError(r, responseData, init, options, currentPath)
```
stepan662, line 64: It would be nice to get rid of this (as mentioned before)

webapp/src/colors.tsx
```diff
@@ -306,6 +317,11 @@ export const colors = {
       mainText: '#BEF4E9',
       linkText: '#dddddd',
     },
+    emailNotVerifiedBanner: {
+      background: '#822343',
+      mainText: '#FDECF2',
+      titleText: '#F6A7C2',
+    },
```
stepan662, line 324: This way of defining colors is deprecated, we now use tokens directly exported from figma (in `figmaTheme.ts`). If you check what colors are used in figma, the same colors should be available in `theme.palette.tokens...`. If there are new colors, which are not in the file, you need to use our figma plugin to update the palette https://github.com/tolgee/figma-color-extractor (I can help you with that if needed).

e2e/cypress/e2e/security/signUp.cy.ts
```diff
@@ -133,6 +134,16 @@ context('Sign up', () => {
...
+  it('Signs up and resend email verification', () => {
+    fillAndSubmitSignUpForm(TEST_USERNAME);
+    cy.contains('Thank you for signing up!').should('be.visible');
+
+    cy.contains('Verify your email now');
+
+    gcy('resend-email-button').click();
+    cy.contains('Your verification link has been resent.');
+  });
+
```
stepan662, line 146: We should also check that verification link works (if it's not tested somewhere else)

webapp/src/component/CustomIcons.tsx
```diff
@@ -40,6 +42,12 @@ export const ImportIcon: React.FC<IconProps> = (props) => (
 export const ProjectsIcon: React.FC<IconProps> = (props) => (
   <CustomIcon icon={ProjectsSvg} {...props} />
 );
+export const MailIcon: React.FC<IconProps> = (props) => (
+  <CustomIcon icon={MailSvg} {...props} />
+);
+export const MailIconDark: React.FC<IconProps> = (props) => (
```
stepan662, line 48: This is not icon, it shoudn't be here.

webapp/src/component/EmailNotVerifiedView.tsx
```diff
@@ -0,0 +1,117 @@
...
+            </StyledHeader>
+            <StyledDescription variant="body1" mb={2}>
+              <T
+                keyName="verify_email_description"
+                params={{ email: email, b: <StyledEnabled /> }}
+              />
+            </StyledDescription>
+            <StyledImg src={imageSrc} alt="Verify email" />
+            <StyledHint variant="body2">
+              <T keyName="didnt_receive_email_hint" />
```
stepan662, line 91: It's good to have some common prefix, like here I'd prefix everything with `verify_email_`, so it's better organized

stepan662 (changes requested):

webapp/src/service/http/handleApiError.tsx
```diff
@@ -29,6 +30,16 @@ export const handleApiError = (
       return;
     }
     if (r.status == 403) {
+      const emailVerificationTemplate =
+        /^\/login\/verify_email\/\d+\/[A-Za-z0-9]+$/;
+
+      if (resObject?.code === 'email_not_verified') {
+        if (emailVerificationTemplate.test(currentPath)) return;
+
+        globalContext.actions?.redirectToUnverifiedView();
```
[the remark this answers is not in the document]
huglx: There's an issue that the global context always sends requests (even if it's login/verify/ link). So without this check, when you click on the verify link, it redirects to an unverified view.

webapp/src/colors.tsx
huglx: aha, ok

webapp/src/service/http/handleApiError.tsx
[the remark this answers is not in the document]
huglx: You're right. But when the user clicks on the verification link, it redirect him to the EmailVerificationHandler, but the global context calls the protected endpoind and throws a 403. this check looks if path is /login/verify_email/{id}/{code}, so the error handler doesn't redirect the user to the unverifiedView.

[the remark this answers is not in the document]
stepan662: So, can we prevent the global context from calling the unprotected endpoint?

webapp/src/component/CustomIcons.tsx
huglx: Why not? Then how can I use them?

huglx: Should I use this like image?

webapp/src/component/RootRouter.tsx
```diff
@@ -122,6 +122,9 @@ export const RootRouter = () => (
     </PrivateRoute>
     <RequirePreferredOrganization>
       <Switch>
+        <PrivateRoute exact path={LINKS.PROJECTS.template}>
+          <Redirect to={LINKS.ROOT.template} />
+        </PrivateRoute>
```
stepan662, line 127: Cool 👍

webapp/src/component/RootRouter.tsx
```diff
@@ -122,6 +122,9 @@ export const RootRouter = () => (
...
         <PrivateRoute path={LINKS.PROJECTS.template}>
           <ProjectsRouter />
         </PrivateRoute>
```
stepan662: This can be removed, no?

stepan662 (approved):

stepan662:
Ok, seems like everything is working, but there are some e2e tests failing.

stepan662 (approved):

PR #2388, opened 2024-06-25
Change Java from 17 to 21

M3rcena:
It currently builds and run correctly using Java 21 so its a good update.


JanCizmar:
Hey! Why did you update the npm packages?

JanCizmar (commented):
We will also need to update the docker images to java 21. This is on me.

M3rcena:
> Hey! Why did you update the npm packages?

So they had severe vulnerabilities especially with the packages you use. It's a security flaw while they were outdated for a long time..

Also there is another security flaw which needs to be resolved. I will tell you in Slack while it's not good to mention it here

M3rcena:
> We will also need to update the docker images to java 21. This is on me.

Of course! I could do it but if you want it's all yours

M3rcena:
> We will also need to update the docker images to java 21. This is on me.

Also consider updating and changing the builder because it's deprecated with the latest Gradle versions.

JanCizmar:
What do you mean by the builder?

M3rcena:
> What do you mean by the builder?

If you go and build the docker image you will see you will get a Deprecated message for Gradle 9.0 because you have set the docker builder to 0. You need to either update the builder so it's not deprecated or use the docker builder itself, the builder kit

JanCizmar:
I cannot see update of Gradle in your PR. How are you actually running it?

M3rcena:
I also removed jCenter()

image

PR #2431, opened 2024-08-12
fix: slack > auto translate event

huglx:
- Fixed the bug where users don't receive messages during auto translation.
- Added tests for Slack integration with auto translation.
- Moved `MachineTranslationTest` to a common package for use in Slack integration tests.


huglx:
@JanCizmar,I had to disable email verification in the e2e login tests because it sometimes breaks the test.

e2e/cypress/e2e/security/login.cy.ts
```diff
@@ -55,6 +56,7 @@ context('Login', () => {
   });
 
   it('login with github', () => {
+    disableEmailVerification();
```
JanCizmar, line 59: This looks kinda unrelated to the issue.

JanCizmar (approved):

PR #2432, opened 2024-08-12
feat: more common json formatting

huglx:
remove a space before the colon at json format.


JanCizmar (changes requested):
Please no! ❌❌❌

What if some value contained this string: `Name : {name}`?

https://stackoverflow.com/questions/75937451/how-to-remove-spaces-between-key-and-a-during-serialization-using-jackson

JanCizmar (approved):

PR #2433, opened 2024-08-18
fix: activity label zero keys

huglx:
- do not show activities with an empty list of modified entities

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/repository/activity/ActivityRevisionRepository.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/repository/activity/ActivityRevisionRepository.kt
@@ -14,7 +14,6 @@ interface ActivityRevisionRepository : JpaRepository<ActivityRevision, Long> {
     """
     from ActivityRevision ar
     where ar.projectId = :projectId and ar.type is not null and ar.batchJobChunkExecution is null and ar.type in :types
-    and ar.modifiedEntities is not empty
   """,
   )
   fun getForProject(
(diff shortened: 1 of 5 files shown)


backend/data/src/main/kotlin/io/tolgee/repository/activity/ActivityRevisionRepository.kt
```diff
@@ -14,6 +14,7 @@ interface ActivityRevisionRepository : JpaRepository<ActivityRevision, Long> {
     """
     from ActivityRevision ar
     where ar.projectId = :projectId and ar.type is not null and ar.batchJobChunkExecution is null and ar.type in :types
+    and ar.modifiedEntities is not empty
```
JanCizmar, line 17: There is `saveWithoutModification` property in `ActivityType` `io/tolgee/activity/data/ActivityType.kt:21`. It's not used, but it's prepared there, so potentially, there can be activities in the list without any modified entity. I believe, this activity should have never been stored and the question is why it's being stored in here. `io.tolgee.activity.ActivityService#storeActivityData`

JanCizmar: Maybe you can fix this by not storing the activity data when (modifiedEntities.isEmpty() && !actvityType.saveWithoutModification).

JanCizmar: In the same time I would fix the UI, rather then fixing the backend for this particular case.

JanCizmar (changes requested):

huglx:
@JanCizmar great! I've already added a test and your code passed it, so we can merge I guess.

JanCizmar (approved):

JanCizmar (approved):

Issue #2243, opened 2024-04-14
Allow import XLS and CSV


webard:
Hello,

option to import XLS and CSV files with multiple languages at time will be very useful. In many cases translators uses Excel to translate repeatable translations, e.g. product names like:
- Modern chair Ikea: black, yellow
- Modern chair Ikea: white, red

It is much easier to complete hundreds such translations in Excel than in the platform, and now to upload such a file, transformation to JSON is required.

JanCizmar:
The problem of XLS and CSVs are that the structure is unclear. Can you share some example of standard XLS we should support?

webard:
Structure of CSV can be forced to use e.g. ";" as delimiter, or can be set before import. First column contains key, next columns contains translations. First row contains information about languages.

Here are samples:
[excel.xlsx](https://github.com/tolgee/tolgee-platform/files/14994581/excel.xlsx)
[csv.csv](https://github.com/tolgee/tolgee-platform/files/14994582/csv.csv)

JanCizmar:
OK! thanks, this looks doable. :)

webard:
It would be perfect if the upload supoort adding multiple languages at once.

Thank you! :)

trupu:
+1

Importing multiple languages at once via xls/csv would be great feature. You could f.e. send google doc link to independent translators -> download xls/csv -> import filled doc. By doing so you minimize translators needed background knowledge.

Anty0:
CSV support is complete and will be released shortly.
XLS issue will be tracked here #2555

PR #2444, opened 2024-08-26
Allow import CSV

xReav3r:


Shebuka (approved):
Everything seem correct and very much a needed addition.
Will partialy solve #2243

PR #2445, opened 2024-08-29
fix: ICU plural-case handling during Machine-Translation

balk-sp:
# ICU plural element mishandling explicit case values during translation
When submitting ICU to the machine-translation, it will correctly handle plural-cases 'one' and 'other', but not explicit value cases like '=1' and '=0'

```
Input A: {productCount,plural,one{You have one product.}other{You have # products.}}
Input B: {productCount,plural,=1{You have one product.}other{You have # products.}}

Output A: {productCount,plural,one{You have one product.}other{You have # products.}}
Output B: {productCount,plural,one{You have 1 products.}other{You have # products.}}
```

- `Output B` outputs case 'one' for input case '=1'
- `Output B` uses the value of case 'other' instead of the value of case '=1', and therefore yields **'1 products'**, which is grammatically incorrect. 

# Machine translation
As the ICU text is correctly stored in the database, and correctly shown in the web-interface, but is corrupted during the machine-translation, there must be a text-transformation just prior to making the call to the translation-provider.

During machine-translation, the following queries are sent to the provider.
```
Input A: {productCount,plural,one{You have one product.}other{You have # products.}}
  Translate case 'one':   You have one product.
  Translate case 'other': You have <x id="tolgee-number">10</x> products.

Input B: {productCount,plural,=1{You have one product.}other{You have # products.}}
  Translate case '=1':    You have <x id="tolgee-number">1</x> products.
  Translate case 'other': You have <x id="tolgee-number">10</x> products.

```

# Bug
The reason this is happening is that Tolgee rewrites plural-case '=1' to 'one' (similarly, it maps '=0' to 'zero'), and then tries to find 'one' in the original plural-element, which does not exist, as the plural-element contains the '=1' case. This causes it to fallback to the 'other' case.

# Solution
Besides looking for the 'one' case, we should also look for the '=1' plural-case when doing the lookup.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/PluralTranslationUtil.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/PluralTranslationUtil.kt
@@ -21,16 +21,44 @@ class PluralTranslationUtil(
     return result
   }
 
-  private val preparedFormSourceStrings: Sequence<Pair<String, String>> by lazy {
-    return@lazy targetExamples.asSequence().map {
+  private fun preparedFormSourceStrings(): Sequence<Pair<String, String>> {
+    val formBasedCases = preparedFormSourceStringsBasedOnForm().toList()
+    val localeBasedCases = preparedFormSourceStringsBasedOnLocale().toList()
+    val localeBasedCaseKeys = localeBasedCases.map { e -> e.first }.toSet()
+
+    // N.B.: locale-based cases take precedence over form-based cases,
+    // yet form-cases appear first so that we naturally end with: 'few', 'many', 'other'
+    val mergedCases = ArrayList<Pair<String, String>>()
+    formBasedCases.stream().filter { e -> !localeBasedCaseKeys.contains(e.first) }.forEach { e -> mergedCases.add(e) }
+    localeBasedCases.stream().forEach { e -> mergedCases.add(e) }
+    return mergedCases.asSequence()
+  }
+
+  private fun preparedFormSourceStringsBasedOnLocale(): Sequence<Pair<String, String>> {
+    return targetExamples.asSequence().map {
       val form = sourceRules?.select(it.value.toDouble())
-      val formValue = forms.forms[form] ?: forms.forms[PluralRules.KEYWORD_OTHER] ?: ""
+      val formValue = forms.forms[form] ?: forms.forms["=" + it.value] ?: forms.forms[PluralRules.KEYWORD_OTHER] ?: ""
       it.key to formValue.replaceReplaceNumberPlaceholderWithExample(it.value)
     }
   }
 
+  private fun preparedFormSourceStringsBasedOnForm(): Sequence<Pair<String, String>> {
+    return forms.forms.asSequence().map {
+      if (it.key.startsWith("=") && it.key.substring(1).toDoubleOrNull() != null) {
+        it.key to it.value.replaceReplaceNumberPlaceholderWithExample(it.key.substring(1).toDouble())
+      } else {
+        val numValue = targetExamples[it.key]?.toDouble() ?: 10.0
+        val formValue =
+          forms.forms[it.key] ?: forms.forms[sourceRules?.select(numValue)] ?: forms.forms["=" + it.value]
+            ?: forms.forms[PluralRules.KEYWORD_OTHER] ?: ""
+
+        it.key to formValue.replaceReplaceNumberPlaceholderWithExample(numValue)
+      }
+    }
+  }
+
   private val translated by lazy {
-    preparedFormSourceStrings.map {
+    preparedFormSourceStrings().map {
       it.first to translateFn(it.second)
     }
   }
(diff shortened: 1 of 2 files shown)


JanCizmar:
Thanks a lot for the PR! 🎉 Looks like it breaks the tests. Also this feature will require some unit tests as well.

JanCizmar:
Hey! Sorry It actually doesn't break the tests. It's only the report task which always fail for PRs. But ktlint fails. You have to run `./gradlew ktlintFormat` to fix this.

Anyway, adding unit tests testing this new functionality need to be done before merging this. Are you willing to do this or should we do this ourselves?

balk-sp:
It seems to me that my PR merely patches a side effect of the bug.
To me it feels odd that during translation the plural-cases are rewritten.

**Issue A**
If there is a need to transform the cases (like '=0', '=1', '=2', '=3' to resp. 'zero', 'one', 'two', 'few') then it should be transform in all layers. It doesn't make sense to transform case-value '=1' to 'one', and then look for 'one' in the original plural-statement, and then fall back to '=1' (as is the workaround in my patch). If you go the route of transformation, you also have to transform all the plural-cases (in the AST representation), such that the lookup for 'one' actually works, as opposed to now, where it could only possibly work if somebody were to write both '=1' and 'one' in their plural-cases.

**Issue B**
As a theoretical example, let's say I start off with the following ICU-text:
```
{var,
 plural,
 =0{No products},
 =1{A product},
 =2{A pair of products},
 =3{A typical yet odd amount of products},
 =4{A typical yet even amount of products},
 other,{# products}
}
```

The case-values would be transformed to:
```
 =0 -> zero
 =1 -> one
 =2 -> two
 =3 -> few
 =4 -> few
 other -> other
}
```
Note the 2 'few' case-values. Now I realize this is a contrived example, but I hope it illustrates that the transformation is not as innocent as it seems, it has side-effects in the sense that different cases in the original icu-statement will be mapped to one.

balk-sp:
The root-cause seems to be that the following code:
```
# PluralTranslationUtil.kt

  private val targetExamples by lazy {
    val targetLanguageTag = context.getLanguage(item.targetLanguageId).tag
    val targetULocale = getULocaleFromTag(targetLanguageTag)
    val targetRules = PluralRules.forLocale(targetULocale)
    getPluralFormExamples(targetRules)
  }
```

For target-locale "en-US" returns:
```
{one=1, other=10}
```
which leads to it producing translations for these, where their *intersect* with the plural-cases.

balk-sp:
Would it be a good idea to simply leave explicit-values as they are?

Instead of iterating over `targetExamples`, I here iterate over `forms`
```
# PluralTranslationUtil.kt

  private val preparedFormSourceStrings: Sequence<Pair<String, String>> by lazy {
    return@lazy forms.forms.asSequence().map {
      if (it.key.startsWith("=") && it.key.substring(1).toDoubleOrNull() != null) {
        it.key to it.value.replaceReplaceNumberPlaceholderWithExample(it.key.substring(1).toDouble())
      } else {
        val numValue = (targetExamples[it.key] ?: it.value).toString().toDoubleOrNull() ?: 10.0
        val formValue = forms.forms[sourceRules?.select(numValue)] ?: forms.forms["=" + it.value]
        ?: forms.forms[PluralRules.KEYWORD_OTHER] ?: ""

        it.key to formValue.replaceReplaceNumberPlaceholderWithExample(numValue)
      }
    }
  }
```

This outputs a structure matching the original one:
```
{var,
 plural,
 =0{No products}
 =1{A product}
 =2{A pair of products}
 =3{A typical yet odd amount of products}
 =4{A typical yet even amount of products}
 other{# products}
}
```

JanCizmar:
Yeah, this looks fine!

balk-sp:
I am curious as to why the current implementation was put in place. It seems quite a complex transformation that, I presume, must have been created with a specific purpose.

My observation is that the implementation does more harm than good, given that it transforms basically any plural-statement to merely [one/other] (for locale en-US at least), as it even breaks on trivial cases like '=0'  and '=1', which seems like reasonable cases for many projects using Tolgee to run into...

My implementation basically discards all this complex logic, by making an attempt to keep the output-icu structure the same as the input-icu when the strings are submitted to the translation-provider. It seems that the ideal transformation is no transformation at all, which means we could remove quite a bit of code from the repo.

JanCizmar:
Hey! As I am thinking about it now, maybe no transformation would indeed be better solution. The reason why I added the complex transformation was that I wanted to give the machine translators number example. But I forgot that the form is actually also pretty good example value for the machine translators. So now I believe, that maybe we can just use the variant name (`zero`, `one` ...) As the example. However, we still need to find example for other and handle the situation when there is collision with some exact form.

JanCizmar:
Oh! Sorry. As I am checking the code, I can see that the transformation actually is required. I will try fix the issue and add some tests.

JanCizmar:
For many cases the source forms doesn't match the target forms. For example in English there is only one and other required, but the same string requires 4 forms to be provided in Czech language. That's why this transformation is necessary. I probably did the same thing as you by adding the exact forms to the data separately, so we got them provided as well. This should handle all the situations. https://github.com/tolgee/tolgee-platform/pull/2454

I just need to test, how the UI will handle this situation.

JanCizmar:
OKI doke, so I am closing this. Thanks for cooperation. :)

PR #2446, opened 2024-08-29
fix: single-step-import lack of auto-translation

balk-sp:
# Issue
The auto-translation step is not executed after single-step-import, when the project is configured to have auto-translation executed on import.

# Solution
The PR is functional, but may not be in line with the event triggering/handling of Tolgee.
It may be possible that this PR can be boiled down to 1 line: firing the correct Event.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/dataImport/ImportService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/dataImport/ImportService.kt
@@ -2,6 +2,8 @@ package io.tolgee.service.dataImport
 
 import io.sentry.Sentry
 import io.tolgee.api.IImportSettings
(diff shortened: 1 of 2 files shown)


JanCizmar:
Hey! Thanks for the PR. The AutoTranslation is handled by this listener: `io.tolgee.component.autoTranslation.AutoTranslationListener#onApplicationEvent` so maybe there's error somewhere around it?

JanCizmar:
Oh! Maybe it's because single step import doesn't have  `@RequestActivity(ActivityType.IMPORT)` annotation in the controller method: `io.tolgee.api.v2.controllers.dataImport.SingleStepImportController#doImport`

JanCizmar:
I've fixed it here: https://github.com/tolgee/tolgee-platform/pull/2447

Issue #2443, opened 2024-08-26
Time to test is too high for backend integration tests 


JanCizmar:
**Heads up:** This task requires significant knowledge of Gradle and Spring Boot. If you have never worked with these technologies, this is probably not an issue for you, and attempting it might be a waste of your time.

When I run integration test e.g. (io.tolgee.api.v2.controllers.v2ProjectsController.ProjectsControllerTest) it takes too much time on my high end MacBook Pro M3. 
- 10.7 seconds for subsequent build without any change to codebase
- 11.7 seconds to start the spring boot context
   `Started ProjectsControllerTest in 10.551 seconds (process running for 11.711)`

Lower time test is crucial for our developer experience. We need to get time to test for integration tests under **10** seconds (build + context load).

- [x] Tune up gradle build, so subsequent builds are immediate (or fast as possible)
- [ ] Tune up Spring Integration Tests, so it doesn't take so long to start the context. We need to get under **5** seconds.

To setup the project, follow [this guide](https://github.com/tolgee/tolgee-platform/wiki/Development).

Vayras:
@JanCizmar can I get assigned to this I have almost one year of testing experience now, and have worked testing for wiremock, stakwork and nabers

JanCizmar:
Hey! Thanks for attempting. I don't know if wiremock experience will help here much. This task will involve a lot of Spring Boot and Gradle. I want to keep the task unassigned. I want to keep space for other contributors. Please update me if you proceed in any way with this, and then I will assign you. 🚀

exi:
I spent roughly a day on this and here are my findings, although none of them will achieve the goal you set since this is a "death by 1000 paper cuts" situation.

Container startup is very slow, for a couple of reasons, which leads to very slow test times even for completely empty test classes.

# Suggestions first

My proper long-term suggestion if you _really_ want faster dev times, is to split your monolith up into maybe 2-3 separate service with a lot less stuff going on in each since you have reached the practical limits of JPA and Hibernate who are generally not very fast to startup.

A lighter variant would be to refactor your code to make spring test slices like `@WebMvcTest` work again. They currently do not work because of some dependency issues. This would however not be a complete integration test.

Another much more toilsome approach would be to drop hibernate/jpa and migrate to plain Spring Data JDBC, which is much simpler and faster.

[...]

PR #2449, opened 2024-09-02
chore: Optimize gradle branch

c0d33ngr:
This PR is to close issue #2443 
/claim #2443

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -162,3 +162,14 @@ subprojects {
         }
     }
 }
+
+tasks.withType(Test).configureEach {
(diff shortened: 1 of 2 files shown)


c0d33ngr:
I'll have to remove code that disable Gradle to generate test reports as the CI/CD tests for presence of them

JanCizmar:
Hello! Unfortunately according to the code it doesn't look like you're on the right path. Running tests in parallel will make them unstable and wouldn't improve time to test of single test, which is the main target.

PR #2454, opened 2024-09-03
fix: Exact plural forms for basic MT translators

JanCizmar:


balk-sp:
Looking at your current implementation of `getPreparedSourceStrings()`

Let's say the target locale is 'en-US' (with target-examples: 'one', 'other')
Let's say my ICU string contains 'zero' (as opposed to '=0').
If I interpret the logic correctly, it would remove the '=0' case.

In my closed PR, I merge the cases quite carefully, to retain as many cases as possible.

Please also note that your unit-tests typically have target-locale 'cz', which happens to have an abundance of target-examples (one, few many, other), which might yield the desired output in your unit-tests, but would not produce the correct result for English as a target.

JanCizmar:
Hey! Look into the test. This case is tested there with `=1` and `one`

stepan662 (approved):

balk-sp:
I tested your branch, and just as I predicted, it removes the zero case.

JanCizmar:
Oh, I see. I believe this is correct. English doesn't require the `zero` case, because this can be usually covered by `other`. We don't want to keep unnecessary cases. So if you really want to force `one` case to be there, you can add your exact case.

balk-sp:
I'm not sure that when defining "zero/one/other" in your source plural-statement, it would be intuitive that most translations would drop the "zero" case, because it happens not to be strictly required in the target-locale. This would be so tricky, that Tolgee would have to advise anybody to use "=0/=1/other" unless they are *really* sure what they are doing and have investigated all target-locale PluralRules (which is hidden in an ibm/icu-package)

JanCizmar:
> "Your shopping cart is empty."

This is great example of the right case for exact`=0` case.

`zero` might be incorrect. See this [table](https://www.unicode.org/cldr/charts/45/supplemental/language_plural_rules.html)
and Prussian example. By using zero form, you might unexpectedly get `10` rendered as "Your shopping cart is empty."

image

balk-sp:
Well I learned something... there even is a whole world of trickery behind even 'one'.So it would be "20 dogs" and "21 dog" in Prussian. Who would have thought.

JanCizmar:
Yeah...

PR #2455, opened 2024-09-06
Import support for i18next

Anty0:
Supported features in current implementation:

- [x] Interpolation
  - [x] Replaceable keys
  - [ ] Keep unescaped flag
  - [ ] Other
- [x] Format
  - [x] Number (only without options)
  - [ ] Currency
  - [ ] Datetime
  - [ ] Relativetime
  - [ ] List
  - [ ] Custom format functions
  - [ ] Legacy format function
- [x] Plurals
  - [x] Ordinal plurals
  - [ ] Interval plurals
- [ ] Nesting
- [ ] Context


Anty0:
Replaced by  #2463

PR #2470, opened 2024-09-18
fix: posthog record only when user is logged in

stepan662:


JanCizmar (approved):
can we please add filter for users, whose e-mail doesn't end by @tolgee.io?

JanCizmar (approved):

Issue #2391, opened 2024-06-25
export description with strings as comment for Android


Nailik:
**Is your feature request related to a problem? Please describe.**
When exporting strings.xml for android the description in tolgee is missing.

**Describe the solution you'd like**
before every string that contains a not empty description there should be a comment 
`<!-- this is a comment in xml -->`

Nailik:
no it's not completed! @JanCizmar the bot should really stop just closing issues.

stepan662:
This shouldn't be complicated, right @JanCizmar?

PR #2472, opened 2024-09-19
feat: Export description as comment for Android

vaclavsauer:
Implementation of https://github.com/tolgee/tolgee-platform/issues/2391

I'm not sure if tests are OK. A lot of them failed because there is a difference in line endings  on windows.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/android/out/AndroidStringsXmlFileWriter.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/android/out/AndroidStringsXmlFileWriter.kt
@@ -1,13 +1,9 @@
 package io.tolgee.formats.android.out
 
-import io.tolgee.formats.android.AndroidStringValue
-import io.tolgee.formats.android.AndroidStringsXmlModel
-import io.tolgee.formats.android.AndroidXmlNode
-import io.tolgee.formats.android.PluralUnit
-import io.tolgee.formats.android.StringArrayUnit
-import io.tolgee.formats.android.StringUnit
+import io.tolgee.formats.android.*
 import io.tolgee.util.attr
 import io.tolgee.util.buildDom
+import io.tolgee.util.comment
 import io.tolgee.util.element
 import org.w3c.dom.Element
 import java.io.InputStream
@@ -25,6 +21,10 @@ class AndroidStringsXmlFileWriter(private val model: AndroidStringsXmlModel) {
   private fun Element.addToElement(unit: Map.Entry<String, AndroidXmlNode>) {
     when (val node = unit.value) {
       is StringUnit -> {
+        val comment = node.comment
+        if (!comment.isNullOrEmpty()) {
+          comment(comment)
+        }
         element("string") {
           attr("name", unit.key)
           appendXmlIfEnabledOrText((unit.value as StringUnit).value)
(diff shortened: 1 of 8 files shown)


backend/data/src/test/kotlin/io/tolgee/unit/formats/android/out/AndroidXmlFileExporterTest.kt
```diff
@@ -160,13 +161,13 @@ class AndroidXmlFileExporterTest {
         )
         add(
           languageTag = "cs",
-          keyName = "percent and paceholders",
-          text =
-            "I am not just a percent {name} % sign!",
+          keyName = "percent and placeholders",
+          text = "I am not just a percent {name} % sign!",
+          description = "This is a description",
```
vaclavsauer, line 166: had to do this this way insted of 
```
{
  key.description = "Description"
}
```
Is it OK?

backend/data/src/main/kotlin/io/tolgee/service/export/dataProvider/ExportTranslationView.kt
```diff
@@ -8,4 +8,5 @@ class ExportTranslationView(
   val state: TranslationState = TranslationState.TRANSLATED,
   val key: ExportKeyView,
   val languageTag: String = "en",
+  val description: String? = null,
```
vaclavsauer, line 11: Also added description here for the test

backend/data/src/test/kotlin/io/tolgee/unit/formats/android/out/AndroidXmlFileExporterTest.kt
```diff
@@ -23,8 +23,9 @@ class AndroidXmlFileExporterTest {
     |<resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
     |  <string name="key1">Ahoj! I%d, %s, %e, %f</string>
     |  <string name="percent_no_placeholders">I am just a percent \% sign!</string>
-    |  <string name="percent_and_paceholders">I am not just a percent %s %% sign!</string>
-    |  <string name="percent_and_paceholders_and_tags"><![CDATA[I am not just a percent <b>%s</b> %% sign!]]></string>
+    |  <!--This is a description-->
```
JanCizmar, line 26: You're testing only the plain string comment, but not plural and string array item.

backend/data/src/main/kotlin/io/tolgee/formats/android/out/AndroidStringsXmlFileWriter.kt
```diff
@@ -25,6 +21,10 @@ class AndroidStringsXmlFileWriter(private val model: AndroidStringsXmlModel) {
   private fun Element.addToElement(unit: Map.Entry<String, AndroidXmlNode>) {
     when (val node = unit.value) {
       is StringUnit -> {
+        val comment = node.comment
+        if (!comment.isNullOrEmpty()) {
```
JanCizmar, line 25: It wold be great to Create a private extension function Element.comment, which will add the comment only if it's not null to remove the deduplicate the duplicated null checking logic.

JanCizmar (changes requested):

Anty0:
Superseded by  #2560

PR #2475, opened 2024-09-19
feat: allow import and export CSV

Anty0:
Supersedes  #2444


Anty0:
It should be ready for review.
To be done:
- I'm not sure if one of the tests behaves as expected (marked with FIXME)
- I need to set the proper logo for the CSV format

PR #2481, opened 2024-09-23
fix: Content delivery configured check & docs

JanCizmar:
- Checks storage is configured to show CD options in the UI
- Noted that CD storage needs to be configured in the docs


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/ConfigurationController.kt
```diff
@@ -25,6 +26,7 @@ class ConfigurationController
     private val applicationContext: ApplicationContext,
     private val publicBillingConfProvider: PublicBillingConfProvider,
     private val versionProvider: VersionProvider,
+    private val CDFileStorageProvider: ContentDeliveryFileStorageProvider,
```
Anty0, line 29: Why start with an uppercase letter?

Anty0 (changes requested):

PR #2523, opened 2024-10-06
feat: sso implementation

huglx:
What I have done:

- adding SSO providers
- logging via SSO
- New users now become a member of the organization that issued the SSO provider
- customized login page for self-hosted instances
- started writing tests

What I plan to add:

- collect the user's role that the organization set for them and grant these roles to the user
- more test coverage
- simplify adding a provider (there is a URL that calls .../well-known/openid-configuration, and through this URL, I can fetch all needed URLs)

tbc


ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/SsoProviderController.kt
```diff
@@ -0,0 +1,25 @@
...
+  @PostMapping("")
+  fun addProvider(
+    @RequestBody request: CreateProviderRequest,
+    @PathVariable organizationId: Long,
+  ): Tenant = tenantService.saveOrUpdate(request, organizationId)
+
+  @GetMapping("")
+  fun getProvider(
+    @PathVariable organizationId: Long,
+  ): TenantDto? = tenantService.findTenant(organizationId)?.toDto()
```
JanCizmar, line 24: Use conventional Model & ModelAssembler approach like e.g. 
```
io.tolgee.hateoas.apiKey.ApiKeyModel
io.tolgee.hateoas.apiKey.ApiKeyModelAssembler
```

webapp/src/views/organizations/sso/OrganizationSsoView.tsx
```diff
@@ -0,0 +1,82 @@
...
+        redirectUri: providersLoadable.data.redirectUri || '',
+        tokenUri: providersLoadable.data.tokenUri || '',
+        jwkSetUri: providersLoadable.data.jwkSetUri || '',
+      };
+
+      setShowForm(providersLoadable.data.isEnabled);
+    }
+  }, [providersLoadable.data]);
+  const handleSwitchChange = (event) => {
+    setShowForm(event.target.checked);
```
JanCizmar, line 51: Why is this called showForm, when in controlls the disabled property.

webapp/src/views/organizations/sso/OrganizationSsoView.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const providersLoadable = useApiQuery({
+    url: `/v2/{organizationId}/sso/providers`,
+    method: 'get',
+    path: {
+      organizationId: organization.id,
+    },
+  });
+
+  const credentialsRef = useRef({
```
JanCizmar, line 26: Is there any special reason to use useRef? IMO, you can just pass the loadable and handle it all in the Form class by setting initial values. This logic can be pretty much copied from other pars of the system.

webapp/src/views/organizations/sso/OrganizationSsoView.tsx
```diff
@@ -0,0 +1,82 @@
...
+    clientSecret: '',
+    redirectUri: '',
+    tokenUri: '',
+    jwkSetUri: '',
+  });
+  const [showForm, setShowForm] = useState(false);
+
+  useEffect(() => {
+    if (providersLoadable.data) {
+      credentialsRef.current = {
```
JanCizmar, line 38: You cululd just assign the providersLoadable.data. It's all not-nullable.

webapp/vite.config.ts.timestamp-1728215969537-7889e0fea9aee.mjs
```diff
@@ -0,0 +1,120 @@
+// vite.config.ts
```
JanCizmar, line 1: It looks like this file can be deleted.

backend/data/src/main/kotlin/io/tolgee/constants/Message.kt
```diff
@@ -235,6 +235,11 @@ enum class Message {
   SLACK_WORKSPACE_ALREADY_CONNECTED,
   SLACK_CONNECTION_ERROR,
   EMAIL_VERIFICATION_CODE_NOT_VALID,
+  THIRD_PARTY_AUTH_FAILED,
+  TOKEN_EXCHANGE_FAILED,
+  USER_INFO_RETRIEVAL_FAILED,
+  ID_TOKEN_EXPIRED,
+  DOMAIN_NOT_ENABLED,
```
JanCizmar, line 242: This message is not very good key for translation. I would at least prefix it with SSO_

backend/data/src/main/kotlin/io/tolgee/constants/Message.kt
```diff
@@ -235,6 +235,11 @@ enum class Message {
   SLACK_WORKSPACE_ALREADY_CONNECTED,
   SLACK_CONNECTION_ERROR,
   EMAIL_VERIFICATION_CODE_NOT_VALID,
+  THIRD_PARTY_AUTH_FAILED,
+  TOKEN_EXCHANGE_FAILED,
+  USER_INFO_RETRIEVAL_FAILED,
+  ID_TOKEN_EXPIRED,
```
JanCizmar, line 241: This message is not very good key for translation. I would at least prefix it with SSO_

JanCizmar (changes requested):

ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/SsoProviderController.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import io.tolgee.ee.data.CreateProviderRequest
+import io.tolgee.ee.data.TenantDto
+import io.tolgee.ee.data.toDto
+import io.tolgee.ee.model.Tenant
+import io.tolgee.ee.service.TenantService
+import org.springframework.web.bind.annotation.*
+
+@RestController
+@RequestMapping(value = ["/v2/{organizationId:[0-9]+}/sso/providers"])
```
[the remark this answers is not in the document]
huglx: Yes, I think your naming is more appropriate.

huglx:
I have gone through all the change requests. I just have a question about the validation schema. I created `SSO_PROVIDER` in `GlobalValidationSchema` and added it to the form. Is this the correct way to do it? Or is there another way to do it?

On the weekend, I plan to write tests and documentation, so I hope it will be ready by Monday.

JanCizmar:
> I have gone through all the change requests. I just have a question about the validation schema. I created `SSO_PROVIDER` in `GlobalValidationSchema` and added it to the form. Is this the correct way to do it? Or is there another way to do it?
> 
> On the weekend, I plan to write tests and documentation, so I hope it will be ready by Monday.

yes, that's correct approach.

e2e/cypress/common/apiCalls/common.ts
```diff
@@ -195,6 +195,26 @@ export const setTranslations = (
     method: 'POST',
   });
 
+export const setSsoProvider = () => {
+  const sql = `insert into ee.tenant (id, organization_id, domain, client_id, client_secret, authorization_uri,
```
JanCizmar, line 199: Why don't you create test data for this?

e2e/cypress/e2e/security/login.cy.ts
```diff
@@ -31,6 +35,22 @@ context('Login', () => {
     cy.gcy('global-language-menu').should('be.visible');
   });
 
+  context('Test Suite for SSO Login', () => {
```
JanCizmar, line 38:
```suggestion
  context('SSO Login', () => {
```

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/OAuthService.kt
```diff
@@ -163,7 +152,7 @@ class OAuthService(
         domain = tenant.domain,
         organizationId = tenant.organizationId,
       )
-    val user = oAuthUserHandler.findOrCreateUser(userData, invitationCode, tenant.domain)
+    val user = oAuthUserHandler.findOrCreateUser(userData, invitationCode, "sso")
```
JanCizmar, line 155: Search only for users with the same SSO domain.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/OAuthTest.kt
```diff
@@ -61,11 +68,50 @@ class OAuthTest : AbstractControllerTest() {
...
-  }
 
   @Test
-  fun authorize() {
+  fun `creates new user account and return access token on sso log in`() {
+    addTenant()
     val response = oAuthMultiTenantsMocks.authorize("registrationId")
     assertThat(response.response.status).isEqualTo(200)
+    val result = jacksonObjectMapper().readValue(response.response.contentAsString, HashMap::class.java)
+    Assertions.assertThat(result["accessToken"]).isNotNull
```
JanCizmar, line 78: use result["accessToken"].assert.isNotNull syntax

ee/backend/tests/src/test/kotlin/io/tolgee/ee/utils/OAuthMultiTenantsMocks.kt
```diff
@@ -69,17 +70,23 @@ class OAuthMultiTenantsMocks(
...
         eq(tenant.tokenUri),
         eq(HttpMethod.POST),
         any(),
         eq(OAuth2TokenResponse::class.java),
       ),
-    ).thenReturn(defaultTokenResponse)
+    ).thenReturn(tokenResponse)
+
+    // mock parsing of jwt
     mockJwk()
```
JanCizmar, line 90: Should be called mockJwt

JanCizmar (changes requested):

Anty0 (changes requested):
There are also a lot of unrelated formatting changes as part of the PR. They seem irrelevant to the PR, and I'd rather see them reverted.

Anty0 (changes requested):
I'll merge the changes to the `sso` branch to start writing code on top of them, but all open review notes still apply—please check and fix them. Notably, there are still many formatting changes throughout the patch set.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/OAuthService.kt
```diff
@@ -0,0 +1,231 @@
...
+        domain = tenant.domain,
+        organizationId = tenant.organizationId,
+        refreshToken = refreshToken,
+      )
+    val user = oAuthUserHandler.findOrCreateUser(userData, invitationCode, ThirdPartyAuthType.SSO)
+    val jwt = jwtService.emitToken(user.id)
+    return JwtAuthenticationResponse(jwt)
+  }
+
+  override fun verifyUserIsStillEmployed(
```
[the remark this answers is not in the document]
JanCizmar: Yes, we don't know the relationship between user and the company. I was speking in examples.

backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
```diff
@@ -53,7 +45,14 @@ data class UserAccount(
   var emailVerification: EmailVerification? = null
 
   @Column(name = "third_party_auth_type")
-  var thirdPartyAuthType: String? = null
+  @Enumerated(EnumType.STRING)
+  var thirdPartyAuthType: ThirdPartyAuthType? = null
```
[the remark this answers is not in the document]
huglx: I can't understand why this is incompatible with the db schema. `@Enumerated` was already used in the project and worked fine. 🙃

backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
```diff
@@ -53,7 +45,14 @@ data class UserAccount(
...
+
+  @ManyToOne()
+  var ssoConfig: SsoConfig? = null
```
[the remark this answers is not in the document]
huglx: Sorry, you're right, it is misunderstanding

backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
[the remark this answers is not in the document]
Anty0: Unfortunately, since we are using an already existing column, we have to make sure it is compatible with the previous values. Since we used previously lowercase values like `github` and `google`, we have to either keep using them or write a database migration to migrate them to the new values.

Due to the way the `@Enumerated` works, the real values saved in DB will be in uppercase, making them incompatible with previous values. To fix this, you can write a custom converter that will use your `code()` method instead of the enum's name.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/OAuthService.kt
```diff
@@ -0,0 +1,231 @@
...
+    ssoDomain: String?,
+    userId: Long,
+    refreshToken: String?,
+    thirdPartyAuth: String?,
+  ): Boolean {
+    val thirdPartyAuthType =
+      thirdPartyAuth?.let {
+        runCatching { ThirdPartyAuthType.valueOf(it.uppercase()) }.getOrNull()
+      }
```
Anty0, line 177: We should rather fail with a bad request than fall back to `null` for invalid values. I would also do the conversion outside this function and pass the `ThirdPartyAuthType` instead.

PR #2550, opened 2024-10-08
feat: Add Plausible to the app

JanCizmar:


webapp/src/hooks/plausible.ts
stepan662, line 1: Should be named `usePlausible.ts`

Issue #2464, opened 2024-09-16
Parse comments from android resource xml files


JanCizmar:
Currently android parser doesn't parse comments.
```
 <!-- Error subtitle while processing abuse and trying to find replacement space, Is followed by the error description. -->
 <string name="error_abuseRequest_subtitle">There was an error processing the request.</string>
 ```

- [ ] Extract the comments from xml resources files
- [ ] Test it
- [ ] Handle comments too long

PR #2560, opened 2024-10-10
feat: export and import comments as description for Android XML

### LINKED ISSUE #2391: export description with strings as comment for Android

Anty0:
Supersedes  #2472
Fixes  #2391 and  #2464


backend/data/src/main/kotlin/io/tolgee/formats/android/out/AndroidStringsXmlFileWriter.kt
```diff
@@ -67,11 +66,13 @@ class AndroidStringsXmlFileWriter(private val model: AndroidStringsXmlModel) {
...
-        this.appendChild(imported)
-      }
+    contentToAppend.children?.forEach {
+      val imported = this.ownerDocument.importNode(it, true)
+      this.appendChild(imported)
     }
   }
+
+  private fun Element.optionalComment(comment: String?) {
+    comment?.let { comment(" $it ") }
```
Anty0, line 76: I prefer how the comments look when there are spaces around them. Is it ok to do this?

JanCizmar: Yes

JanCizmar, line 76: Don't add it if it's empty.

JanCizmar (changes requested):

JanCizmar (approved):

backend/data/src/main/kotlin/io/tolgee/formats/android/out/AndroidStringsXmlFileWriter.kt
Anty0: good point

PR #2572, opened 2024-10-11
chore: Don't truncate database tables between tests

JanCizmar:


laz-001:
@JanCizmar

A) I started to process this locally, hitting on some merge-conflicts affecting sql-queries. I believe this would need the expertise of the core-devs that know the code in-and-out. The code-changes seem quite complex, possibly unmanagable - but I'm not sure about this assessment.

So the question: how confident are you about the changes in this PR and the overall tactic and the neccessity.

B) Would you be willing to rebase this, so there's a cleaner foundation to go on?

C) Over the last days I started locally different attempts that may have less effort:
- provide opt-in/opt-out `CleanDbBefore` tags (very simple tagging, nothing fancy: this can work, but possibly limited results)
- change the tests to be more transactional (this led to problems with inNewTransaction sections in test-code)
- add opt-out uniqueness-generators with minimum(!) effort (again, not trivial, due to interweaved problems)

@coderabbitai can you assess this PR as for feasibility of beeing completed? Start with rebase problems (e.g. merge-conflicts in critical/compelx code), and go on.

JanCizmar:
Hey! I can rebase this, but there still will be lot of work to do, since the tests don't pass. However, truncating the tables has huge impact on the total test time. Truncate takes c.a. 500ms on my machine, which for 800 test is 400 seconds ~5min. 🤷

laz-001:
> Hey! I can rebase this,

that would be nice, as its quite a (mental/real) barrier.

> but there still will be lot of work to do,

To clarify: you have applied a working pattern you introduced to SOME classes (which pass now), but many classes need still to be updated with the new pattern, thus they pass.

So, essentially, it's grinding work, possibly with some adjustments here and there?

laz-001:
> Truncate takes c.a. 500ms on my machine, which for 800 test is 400 seconds ~5min.

Sounds very impressive. and if a test-class has per-method-cleanup, it should speed-up single test, too.

### @Transactional

I did some experiments with `@Transactional`. Is this ok, or do your tests mostly rely on committing? (the "with new transaction" was a heavy barrier)

JanCizmar:
Yes, we must not use `@Transactional` in tests, beauase it behaves different and can lead to actual issues. @Transactional on test actually creates a transaction level above the controller method. Then if you for example forget to create transaction on service layer, it can paas in the tests, but fail in the production.

That's why I consider `@Transactional` in tests anti-pattern.

> To clarify: you have applied a working pattern you introduced to SOME classes (which pass now), but many classes need still to be updated with the new pattern, thus they pass.

> So, essentially, it's grinding work, possibly with some adjustments here and there?

Exactly.

laz-001:
> That's why I consider `@Transactional` in tests anti-pattern.

Very well. Personally I'd prefere that the db behaves real, too. And tests can run in parallel like crazy, with not negative effect. So unique data is the solution, i guess.

As I saw it, the rebase should be trivial if you know the code.

laz-001:
i'm going over this locally, using the original branch and slightly rewriting it, without rebasing for now.

PR #2590, opened 2024-10-13
Dream recorder

RashmitTopG:
Created Dream Recorder using React+VIte and added Language Support Using Tolgee.


RashmitTopG:
Is there any way to dynamically translate the dreams that will be added without them being already translated and saved as keys.

JanCizmar (changes requested):
Hello! It's a great demo! 🚀

Please add the localization jsons to the code or use CDN, so the localization works even withtout providing the API key.

JanCizmar:
@RashmitTopG We don't have to translate the dreams since it's user defined content, which is out-of-scope for Tolgee. Usually you can use services like DeepL or Google Translate, but we don't have to cover this in the demo.

RashmitTopG:
Hey !@JanCizmar can you please review the changes and also check my other PR for Demo Workout Dice

JanCizmar:
Hello! Where can I find the link for demo and dev.to article?

RashmitTopG:
Thank you for your patience. I will make sure to add them by the end of the
day today.

RashmitTopG:
I will add the live link and blog by EOD can you open the issue ??

JanCizmar:
It seems like there already is a dreamRecorder. So maybe you need to rename the folder?

JanCizmar:
Any updates to this?

RashmitTopG:
@JanCizmar made a new PR for the dreamRecorder . Sorry for the delay .

PR #2599, opened 2024-10-15
feat: import resolvable force override

huglx:
I added new option `force-override` to import resolvable endpoint. It should forcefully update/create/keeps the translation to the new value. Also I wrote unit test.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/key/ResolvingKeyImporter.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/key/ResolvingKeyImporter.kt
@@ -12,7 +12,11 @@ import io.tolgee.exceptions.NotFoundException
 import io.tolgee.exceptions.PermissionException
 import io.tolgee.formats.convertToIcuPlurals
 import io.tolgee.formats.convertToPluralIfAnyIsPlural
-import io.tolgee.model.*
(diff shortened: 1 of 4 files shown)


backend/data/src/main/kotlin/io/tolgee/service/key/ResolvingKeyImporter.kt
```diff
@@ -116,6 +125,28 @@ class ResolvingKeyImporter(
...
+  ) {
+    if (new) {
+      val translation =
+        Translation(translationText).apply {
+          this.key = key
+          this.language = entityManager.getReference(Language::class.java, language.id)
+        }
+      addToList(TranslationToModify(translation, translationText, true))
+    } else {
+      if (existingTranslation != null) {
```
stepan662, line 144: This condition feels weird, as you are checking if it's new in the condition before, so the else statement suggests that existing translation should exist but you are still checking it.

huglx: I did this because `TranslationToModify` needs a non-nullable existingTranslation, so if I don't check this on null, the compiler will yell at me

stepan662 (approved):

PR #2600, opened 2024-10-15
fix: add missing indexes for FKs (first part)

Anty0:


backend/data/src/main/kotlin/io/tolgee/repository/KeyRepository.kt
```diff
@@ -204,7 +204,7 @@ interface KeyRepository : JpaRepository<Key, Long> {
     """
     from Language l
     join l.translations t
-    where t.id in (select t.id from Key k join k.translations t where k.id = :keyId)
+    where t.key.id = :keyId
```
Anty0, line 207: I feel like the previous version was doing the same as this. Am I correct? Please check

Anty0:
I've noticed the autogenerated names for indexes could be better. Should I change them all to something more readable or leave them as is?

JanCizmar (approved):

PR #2617, opened 2024-10-23
chore: Update liquibase

JanCizmar:


settings.gradle
```diff
@@ -61,8 +61,8 @@ dependencyResolutionManagement {
             library('amazonSTS', "software.amazon.awssdk:sts:$amazonAwsSdkVersion")
             library('amazonTranslate', "software.amazon.awssdk:translate:$amazonAwsSdkVersion")
             library('googleCloud', "com.google.cloud:libraries-bom:24.0.0")
-            library('liquibaseCore', "org.liquibase:liquibase-core:4.26.0")
-            library('liquibaseHibernate', "org.liquibase.ext:liquibase-hibernate6:4.26.0")
+            library('liquibaseCore', "org.liquibase:liquibase-core:4.29.2")
+            library('liquibaseHibernate', "org.liquibase.ext:liquibase-hibernate6:4.29.2")
```
Anty0, line 65: According to https://github.com/liquibase/liquibase-gradle-plugin/blob/master/doc/releases.md, we need to update the gradle plugin to 3.X.X version before updating to the latest release of liquibase.

JanCizmar: That plugin version doesn't work :D

Anty0: I couldn't find which versions of liquibase are supported by the 2.x.x gradle plugin, plus there seems to be no info on the incompatibilities between the latest versions and the Gradle plugin. So... if it works like this, there is not much else we can do, I guess. :D

Anty0 (approved):

PR #2618, opened 2024-10-23
fix: use less grid (safari performance)

stepan662:


.kotlin/sessions/kotlin-compiler-10167436309493278633.salive
Anty0, line 1: Temporary file - remove

Anty0 (changes requested):

Anty0 (approved):

PR #2621, opened 2024-10-23
feat: order professional translation

stepan662:


webapp/src/ee/task/components/taskCreate/TaskCreateDialog.tsx
```diff
@@ -111,6 +113,8 @@ export const TaskCreateDialog = ({
   const { t } = useTranslate();
 
   const translateTaskType = useTaskTypeTranslation();
+  const { features } = useEnabledFeatures();
+  const taskFeature = features.includes('TASKS');
```
Anty0, line 117: Why not `isEnabled`?

Anty0 (approved):

webapp/src/ee/task/components/taskCreate/TaskCreateDialog.tsx
stepan662: 👍

PR #2629, opened 2024-10-25
fix: Respect project.icuPlaceholders in singleStepImport

### LINKED ISSUE #2574: Tolgee doesn't parse ruby yaml correctly for some cases
**Describe the bug**
In our Rails project, some of the values result in extra single quote (`'`) when I check on Tolgee admin.

Here's the cases we found:
1. upload `key: value 'value'` to tolgee result in `value 'value''` (there's an extra single quote in the end)
2. upload `key: value'%{variable}'` to tolgee result in `value''<variable>''` (there are two single quotes around variable, which should be one)
3. upload `key: value{{variable}}value` to tolgee result in `value'{{'variable'}}'value` (two single quotes around {{ and }} shouldn't exist)
4. upload `key: "{variable} value"` to tolgee result in `'{'file'}' value` (two single quotes around { and } shouldn't exist)

**To Reproduce**
1. Create tolgee config that use Ruby format:
```
{
  "$schema": "https://tolgee.io/cli-schema.json",
  "format": "YAML_RUBY",
  "push": {
    "files": [
       // these files should contain the patterns mentioned above
    ],
    "forceMode": "OVERRIDE",
    "convertPlaceholdersToIcu": false
  }
}
```
2. Push to the target url
```
npx @tolgee/cli@2.1.4 push \
      --config ./test.tolgeerc.json \
      --api-url ${TOLGEE_INSTANCE_URL} \
      --api-key ${TOLGEE_PERSONAL_ACCESS_KEY} \
      --project-id ${TOLGEE_PROJECT_ID_RAILS} \
      --verbose
```
3. Go to translations page in tolgee host
4. It shows unexpected result
e.g.
image

**Expected behavior**
I expect tolgee don't transform value that is not interpolated string (`%{xxx}`) as they are valid yaml values. e.g. single quote (`'`) and curly braces (`{`, `}`, `{{`, `}}`) should not be escaped.

**Screenshots**
see above

**Versions and environment**
- Tolgee Platform version: 3.68.2
- Environment: local docker (https://hub.docker.com/layers/tolgee/tolgee/v3.68.2/images/sha256-cd8a247cba511ed846cbc51a36b221c24b69a18b32cae63ed3db36f3a3e87a06?context=explore) 
- Browser: not relevant

**Additional context**

Anty0:
Fixes #2574

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/dataImport/ImportService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/dataImport/ImportService.kt
@@ -137,6 +137,7 @@ class ImportService(
         // author, project, file issues, etc
         import = import,
         params = params,
+        projectIcuPlaceholdersEnabled = project.icuPlaceholders,
         importSettings = params,
         saveData = false,
       )


Anty0:
I wanted to create a test for the issue but didn't find a good place for it. We test the underlying import mechanics, but the actual API is tested by E2E tests. Since the FE doesn't use the single-step import, I don't see an easy way to test that API endpoint. Do you have any ideas, @JanCizmar?

PR #2660, opened 2024-10-31
feat: Provide filtering by key tags when getting all translations

### LINKED ISSUE #2453: Add tag filter to the translation returning endpoint for JS SKDs
According to this [discussion](https://github.com/tolgee/tolgee-platform/discussions/2436), when client uses single project for multiple platforms, they need a way how to fetch only keys for specific platform tag. e.g. `web`, `mobile`. To achieve the JS SDK enhancement, we need to enable tag filtering on the endpoint.

- [ ] Enable tag filtering on this endpoint https://tolgee.io/api/get-all-translations

StaNov:
Closes #2453.

### DIFF
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
@@ -150,10 +150,10 @@ When null, resulting file will be a flat key-value object.
     structureDelimiter: Char?,
     @Parameter(
       description =
-      "Enables filtering of returned keys by their tags.\n" +
-        "Only keys with at least one provided tag will be returned.\n" +
-        "Optional, filtering is not applied if not specified.",
-      example = "?filterTag=productionReady&filterTag=nextRelease",
+        "Enables filtering of returned keys by their tags.\n" +
+          "Only keys with at least one provided tag will be returned.\n" +
+          "Optional, filtering is not applied if not specified.",
+      example = """["productionReady", "nextRelease"]""",
     )
     @RequestParam(value = "filterTag", required = false)
     filterTag: List<String>? = null,
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
@@ -132,7 +132,7 @@ class TranslationsController(
       description =
         "Comma-separated language tags to return translations in. " +
           "Languages you are not permitted to see will be silently dropped and not returned.",
-      example = "en,de,fr",
+      example = """["en","de","fr"]""",
     )
     @PathVariable("languages")
     languages: Set<String>,
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
@@ -148,6 +148,15 @@ When null, resulting file will be a flat key-value object.
     )
     @RequestParam(value = "structureDelimiter", defaultValue = ".", required = false)
     structureDelimiter: Char?,
+    @Parameter(
+      description =
+      "Enables filtering of returned keys by their tags.\n" +
+        "Only keys with at least one provided tag will be returned.\n" +
+        "Optional, filtering is not applied if not specified.",
+      example = "?filterTag=productionReady&filterTag=nextRelease",
+    )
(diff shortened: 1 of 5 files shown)


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
```diff
@@ -148,6 +148,15 @@ When null, resulting file will be a flat key-value object.
     )
     @RequestParam(value = "structureDelimiter", defaultValue = ".", required = false)
     structureDelimiter: Char?,
+    @Parameter(
+      description =
+      "Enables filtering of returned keys by their tags.\n" +
+        "Only keys with at least one provided tag will be returned.\n" +
+        "Optional, filtering is not applied if not specified.",
+      example = "?filterTag=productionReady&filterTag=nextRelease",
```
JanCizmar, line 156: The example should not look like this, it will used as default value in the swagger UI.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/TranslationsControllerViewTest.kt
```diff
@@ -245,6 +246,24 @@ class TranslationsControllerViewTest : ProjectAuthControllerTest("/v2/projects/"
       }
   }
 
+  @ProjectJWTAuthTestMethod
+  @Test
+  fun `returns keys filtered by provided tags`() {
+    testData.generateLotOfData()
+    testDataService.saveTestData(testData.root)
+    userAccount = testData.user
+    val expectedKeys = IntRange(1, 9).map { "key 0$it" }
```
JanCizmar, line 255: In kotlin, you can define range this way `(1..9)`.

However, it's not very clean to rely on the range. What if someone later comes there and edit the test data like you did.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/TranslationsControllerViewTest.kt
```diff
@@ -245,6 +246,24 @@ class TranslationsControllerViewTest : ProjectAuthControllerTest("/v2/projects/"
...
+    performProjectAuthGet("/translations/en,de?filterTag=${FIRST_9_KEYS_TAGS}&filterTag=unknownTag")
```
JanCizmar, line 256: I would at least put the constant into the TranslationsTestData class and not to the top level.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TranslationsTestData.kt
```diff
@@ -221,10 +223,14 @@ class TranslationsTestData {
           addTranslation {
             language = germanLanguage
             text = "I am key $padNum's german translation."
+            if (it < 10)
```
JanCizmar, line 226: I don't like you're editing the `generateLotOfData` which has it's own purpose. You can create your custom method for this test, where you can directly return the tagged keys. Then you wouldn't have to rely on the range 1..9.

JanCizmar (changes requested):

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/translation/TranslationsController.kt
StaNov: Thanks for the catch, fixed.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/TranslationsControllerViewTest.kt
StaNov: Now deleted.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/TranslationsControllerViewTest.kt
StaNov: Now deleted.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TranslationsTestData.kt
StaNov: I used `addFewKeysWithTags` instead which was missing translations, so I added them.

I see it as an equal option to creating brand new test data for this usecase. If you find it better, I can do it like new data.

JanCizmar (approved):

PR #2667, opened 2024-11-01
fix: changes from first PR (sso)

huglx:
I'll try to summarize what I've done here:

1. Functions and fields have been renamed.
2. Moved the `SsoTenant` class from the EE repository.
3. Removed `SsoConfig` (entity to store `UserAccount` and domain name); now uses the relationship between `SsoTenant` and `UserAccount`.
4. Added a foreign key for the organization in `SsoTenant`.
5. Use of a cacheable user entity to validate SSO users instead of using a cache with an expiration manager.
6. Kept the global tenant in the database since the link between `SsoTenant` and `UserAccount` does not work without it.
7. Threw an exception when disabling native authentication and attempting to log in.
8. When saving the `thirdAuthType` enumeration, `@Convert` is used instead of `@Enumerated`.


Anty0 (changes requested):

Anty0:
I'll take care of fixing the rest of these. :)

PR #2669, opened 2024-11-04
chore: Do not depend on Spring test library as implementation in backend

StaNov:
## Issue

* We are now depending on `spring-boot-starter-test` and its transitive dependency on `org.json` which is from some strange "Android Vaadin" lib.

## Proposed solution

* Depend on the test lib only in tests.
* Use the `org.json` package from Mailjet which uses it as an interface.


JanCizmar:
Thanks a lot! 🚀

Issue #2675, opened 2024-11-09
Health check broken as image is missing curl


vexdev:
**Describe the bug**
The docker container reports as unhealthy 

**To Reproduce**
1. Setup docker compose with this image version: v3.81.0
2. docker compose up -d
3. docker ps
4. notice how the health check fails

**Expected behavior**
Health check succeeds

**Screenshots**
Unnecessary

**Versions and environment**
- Tolgee Platform version: 3.81.0
- Environment: Docker version 24.0.6, build ed223bc, Host Debian 6.1.52-1 (2023-09-07) x86_64 GNU/Linux
- Browser: Not UI

**Additional context**
Happens because https://github.com/tolgee/tolgee-platform/pull/2647/files introduced curl as health check while curl is not available on the image

vexdev:
Shall anyone encounter this issue, while waiting for a proper fix on the image you could just override the healthcheck in docker compose, like this:

```
    healthcheck:
      test: ["CMD", "wget", "-q", "-O", "-", "http://127.0.0.1:8080/actuator/health"]
```

(Not a workaround to keep)

PR #2676, opened 2024-11-09
Replace curl with wget (Addresses #2675)

vexdev:
Replace curl with wget as curl is not available from the base image.
Simple proposal to address #2675

### DIFF
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -28,4 +28,4 @@ CMD ["java", "-cp", "app:app/lib/*", "io.tolgee.Application"]
 
 # Health check to ensure the app is up and running
 HEALTHCHECK --interval=10s --timeout=3s --retries=20 \
-    CMD curl -f http://127.0.0.1:$HEALTHCHECK_PORT/actuator/health || exit 1
+    CMD wget -q -O - http://127.0.0.1:$HEALTHCHECK_PORT/actuator/health || exit 1


docker/app/Dockerfile
```diff
@@ -28,4 +28,4 @@ CMD ["java", "-cp", "app:app/lib/*", "io.tolgee.Application"]
 
 # Health check to ensure the app is up and running
 HEALTHCHECK --interval=10s --timeout=3s --retries=20 \
-    CMD curl -f http://127.0.0.1:$HEALTHCHECK_PORT/actuator/health || exit 1
+    CMD wget -q -O - http://127.0.0.1:$HEALTHCHECK_PORT/actuator/health || exit 1
```
StaNov, line 31: Thank you for your contribution, it looks working :)

Minor comment: `wget` can be called with the `--spider` parameter not to download the content, just check the HTTP code.

Also the `|| exit 1` is redundant, since it is the default `wget` functionality.

```suggestion
    CMD wget --spider -q http://127.0.0.1:$HEALTHCHECK_PORT/actuator/health
```

cyyynthia: It is not redundant; wget doesn't always exit with exit code 1 which is the expected code for unhealthy containers, not just non-zero. While it works, the Dockerfile reference explicitly expects an exit code 1, reserves the exit code 2, and makes no mention of other exit codes making them effectively undefined behavior. 

- https://docs.docker.com/reference/dockerfile/#healthcheck
- https://man.archlinux.org/man/wget.1.en#EXIT_STATUS

JanCizmar: @StaNov Can you please finalize this so we can merge it? 🙏

StaNov: I have to create a new PR because I cannot work on this branch.

@vexdev FYI, if you were able to finish it in a matter of hours, I'd leave it to you.

@cyyynthia, thanks for the explanation, I didn't know this healthcheck contract.

StaNov: FYI, I created a new PR #2694. I credited @vexdev in the description.

JanCizmar:
It's fixed here: https://github.com/tolgee/tolgee-platform/pull/2694

Thanks to all for contributions!

docker/app/Dockerfile
vexdev: No worries I'm not after credits only helpful contributions when I can :)

PR #2678, opened 2024-11-11
fix: update signup form

stepan662:


Anty0 (approved):
I didn't try running it, but otherwise it lgtm.

PR #2690, opened 2024-11-12
chore: Refactor public configuration so it's safe

JanCizmar:


backend/api/src/main/kotlin/io/tolgee/api/publicConfiguration/PublicConfigurationAssembler.kt
```diff
@@ -0,0 +1,106 @@
...
+      showVersion = properties.internal.showVersion,
+      internalControllerEnabled = properties.internal.controllerEnabled,
+      maxTranslationTextLength = properties.maxTranslationTextLength,
+      recaptchaSiteKey = properties.recaptcha.siteKey,
+      chatwootToken = properties.chatwootToken,
+      capterraTracker = properties.capterraTracker,
+      ga4Tag = properties.ga4Tag,
+      postHogApiKey = properties.postHog.apiKey,
+      postHogHost = properties.postHog.host,
+      contentDeliveryConfigured = properties.contentDelivery.publicUrlPrefix != null,
```
Anty0, line 47: Current implementation uses `cdFileStorageProvider.isServerContentDeliveryConfigured()` which has different logic to check if CD is configured. Is it correct like this?

Anty0 (changes requested):

Anty0 (approved):

PR #2691, opened 2024-11-12
sso existing tests fixes

Anty0:


Anty0:
Not everything is fixed, but should be in a usable state. Merged to have all changes in sso for initial review.

PR #2720, opened 2024-11-25
fix: Prevent ConcurrentModificationException in InterceptedEventsManager

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/activity/iterceptor/InterceptedEventsManager.kt
```diff
@@ -250,8 +251,8 @@ class InterceptedEventsManager(
     }
   }
 
-  private val annotatedMembersCache: MutableMap<Class<*>, Map<String, ActivityLoggedProp>> = mutableMapOf()
-  private val ignoredMembersCache: MutableMap<Class<*>, Set<String>> = mutableMapOf()
+  private val annotatedMembersCache: ConcurrentHashMap<Class<*>, Map<String, ActivityLoggedProp>> = ConcurrentHashMap()
+  private val ignoredMembersCache: ConcurrentHashMap<Class<*>, Set<String>> = ConcurrentHashMap()
```
StaNov, line 255: Minor: Can we keep the interface-like type and specify only the implementation?

```suggestion
  private val annotatedMembersCache: MutableMap<Class<*>, Map<String, ActivityLoggedProp>> = ConcurrentHashMap()
  private val ignoredMembersCache: MutableMap<Class<*>, Set<String>> = ConcurrentHashMap()
```

PR #2724, opened 2024-11-26
fix: Rate limit reset password requests

JanCizmar:


StaNov (approved):
Looks good.

If tests are feasible to create, I would vote for them to be created.

JanCizmar:
You can do it. I just need to fix it fast as possible.

PR #2726, opened 2024-11-26
chore: Create test base project with random UUID by default

StaNov:
The functionality of creating a base project with a particular name is not used anywhere. It can be replaced by creating a base project with UUID as a name for readability.


StaNov:
@JanCizmar Thanks for the info.

This PR should be useful to make the potential migration of the existing code easier.

PR #2729, opened 2024-11-27
chore: Make EE opt-out

JanCizmar:


webapp/src/views/projects/translations/BatchOperations/types.ts
```diff
@@ -1,20 +1,20 @@
...
+//   | 'mark_as_translated'
+//   | 'task_create'
+//   | 'task_add_keys'
+//   | 'task_remove_keys'
+//   | 'add_tags'
+//   | 'remove_tags'
+//   | 'change_namespace'
+//   | 'copy_translations'
+//   | 'clear_translations'
+//   | 'export_translations';
```
stepan662, line 17: Why?

webapp/src/views/projects/translations/BatchOperations/BatchSelect.tsx
```diff
@@ -21,114 +17,17 @@ const StyledSeparator = styled('div')`
 `;
 
 type Props = {
-  value: BatchActions | undefined;
-  onChange: (value: BatchActions | undefined) => void;
+  value: string | undefined;
+  onChange: (value: string | undefined) => void;
```
stepan662, line 21: This should stay typed

webapp/src/plugin/getEe.ts
stepan662, line 1: Is this file necessary? If `eePlugin.local` doesn't exist, doesn't it fail completely anyway?

webapp/src/fixtures/pluginAdder.ts
stepan662, line 1: Not sure I follow this. Why is the file named pluginAdder and the name of the function is createAdder?

webapp/scripts/prepareEe.js
```diff
@@ -0,0 +1,25 @@
+// if src/ee exists copy the webapp/src/eePlugin.ee.tsx to webapp/src/eePlugin.local.tsx
+// else copy the webapp/src/eePlugin.oss.tsx to webapp/src/eePlugin.local.tsx
```
stepan662, line 2: It's actually a symlink, does it need to be a symlink?

webapp/src/component/App.tsx
```diff
@@ -96,7 +58,12 @@ export class App extends React.Component {
     globalContext.actions?.setGlobalError(error as GlobalError);
     throw error;
   }
+
   render() {
+    const {
+      billing: { GlobalLimitPopover },
+    } = getEe();
```
stepan662, line 65: Can't we just do `import {GlobalLimitPopover} from tg.ee` or something like that?

Anty0 (changes requested):

stepan662 (approved):

Anty0 (approved):

PR #2731, opened 2024-11-28
fix: avoid escaping hash characters outside of plurals

Anty0:


backend/data/src/test/kotlin/io/tolgee/unit/formats/po/in/PoToICUConverterTest.kt
```diff
@@ -39,6 +39,50 @@ class PoToICUConverterTest {
     )
   }
 
+  @Test
+  fun testPhpPluralsWithHashtags() {
```
JanCizmar, line 43: Can you please name the tests conventionally?
```suggestion
  fun `php plurals with hashtags works`() {
```

JanCizmar (approved):

backend/data/src/test/kotlin/io/tolgee/unit/formats/po/in/PoToICUConverterTest.kt
Anty0: Other tests in this file are named like this. Should I change all of them?

PR #2744, opened 2024-12-03
chore: Add everything passed action for easier auto-merge conditions

StaNov:
Nowadays, all the actions need to be defined in GitHub UI to enable auto-merge.

This change adds an arbitrary `Everything passed` action dependent on all other, therefore it is the only one which needs to be defined to pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -454,3 +454,18 @@ jobs:
 
       - name: Run ktlint
         run: ./gradlew ktlintCheck
+
+  everything-passed:
+    name: Everything passed 🎉
+    needs:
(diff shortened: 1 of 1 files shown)


JanCizmar:
Can you elaborate more on the reasoning behind this?

StaNov:
@JanCizmar There is no way to easily say to github "require all the checks to pass before the PR can be merged".

It needs to be set manually in github UI to all the checks by hand. Which is tedious.

With this change, only the one `everything passes` check must be added to enable auto-merges.

It is not perfect, but better than now.

JanCizmar (approved):

PR #2748, opened 2024-12-03
feat: non-commercial licences indication

stepan662:


webapp/src/service/billingApiSchema.generated.ts
```diff
@@ -1394,6 +1407,49 @@ export interface operations {
       };
     };
   };
+  addUsageItemsToInvoiceAndFinalizeIt: {
```
StaNov, line 1410: As discussed in person, I would put the forgotten changes into a separate PR, but it is just a matter of taste.

PR #2759, opened 2024-12-09
feat: content delivery last publish files

stepan662:


backend/api/src/main/kotlin/io/tolgee/hateoas/contentDelivery/ContentDeliveryConfigModel.kt
```diff
@@ -20,6 +20,7 @@ class ContentDeliveryConfigModel(
   val publicUrl: String?,
   val autoPublish: Boolean,
   val lastPublished: Long?,
+  val files: Collection<String>,
```
JanCizmar, line 23: This is confusing name. Would future reader understand that these are the files which being published in past and not files to publish?

JanCizmar, line 23: This is confusing name. Would future reader understand that these are the files which being published in past and not files to publish? 

```suggestion
  val lastPublishedFiles: Collection<String>,
```

webapp/src/views/projects/developer/contentDelivery/CdItem.tsx
```diff
@@ -51,79 +65,122 @@ export const CdItem = ({ data }: Props) => {
...
+            }
+            color="primary"
+            loading={publishLoadable.isLoading}
+            data-cy="content-delivery-item-publish"
+          >
+            {t('content_delivery_item_publish')}
+          </LoadingButton>
+        </Box>
+      </StyledWrapper>
+      {data.lastPublished && (
```
JanCizmar, line 140: The last published info can be extracted to separate component.

JanCizmar (changes requested):
Good job for the backend part. 👍

We have to consult the UI with @ZuzanaOdstrcilova, it works and looks differently then it was what specified in the task.

PR #2774, opened 2024-12-12
fix: task creation when feature is disabled

stepan662:


Anty0 (approved):
There might be something broken about TaskCreateDialog - the tests failed there, but otherwise, LGTM.

stepan662:
https://github.com/tolgee/billing/pull/115

PR #2786, opened 2024-12-17
Make namespaces opt-in

### LINKED ISSUE #2594: Make namespaces opt-in
The namespaces feature is unclear for new users who tend to user use it and getting into trouble. We need to make the namespaces opt-in, so it is not in the UI by default.

StaNov:
Closes #2594.


webapp/src/views/projects/import/ImportView.tsx
```diff
@@ -123,6 +123,20 @@ export const ImportView: FunctionComponent = () => {
             addFilesMutation={dataHelper.addFilesMutation}
           />
         ))}
+        {dataHelper.addFilesMutation.data?.warnings.map((item) => (
+          <Box key={item.code} mt={4} data-cy="import-file-warnings">
+            <Alert severity="warning">
+              <AlertTitle>
+                <TranslatedError
+                  code={'import_file_warning_header_' + item.code}
```
stepan662, line 131: Can you just pass `code`? Why the concatenation?

webapp/src/translationTools/useErrorTranslation.ts
```diff
@@ -151,6 +151,18 @@ export function useErrorTranslation() {
...
+      case 'namespaces_cannot_be_disabled_when_namespace_exists':
+        return t('namespaces_cannot_be_disabled_when_namespace_exists');
+      case 'import_file_warning_header_namespace_cannot_be_used_when_feature_is_disabled':
+        return t(
+          'import_file_warning_header_namespace_cannot_be_used_when_feature_is_disabled'
+        );
+      case 'import_file_warning_message_namespace_cannot_be_used_when_feature_is_disabled':
+        return t(
+          'import_file_warning_message_namespace_cannot_be_used_when_feature_is_disabled'
+        );
```
stepan662, line 165: If we need specific messages for the import page (do we?). I'd move the translations to a separate hook.

e2e/cypress/common/namespace.ts
```diff
@@ -1,6 +1,33 @@
...
+    } else {
+      checkbox.uncheck();
+    }
+
+    cy.contains('button', 'Save').click();
+
+    if (originalPage !== 'about:blank') {
+      cy.visit(originalPage);
+    }
+  });
```
stepan662, line 28: This should be an API call or ideally baked into project test data. This way it will slow down a lot of tests.

stepan662 (changes requested):

webapp/src/views/projects/project/ProjectSettingsGeneral.tsx
```diff
@@ -158,6 +169,35 @@ export const ProjectSettingsGeneral = () => {
...
+                label={<T keyName="project_settings_use_namespaces" />}
+                data-cy="project-settings-use-namespaces-checkbox"
+              />
+              <Typography variant="caption">
+                {
+                  <T
+                    keyName="project_settings_use_namespaces_hint"
+                    params={{
+                      a: (
+                        <a
```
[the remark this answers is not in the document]
StaNov: Thanks :)

webapp/src/views/projects/import/ImportView.tsx
StaNov: There is the header and the message. Or maybe I could use just the message and omit the header.

webapp/src/views/projects/translations/BatchOperations/BatchSelect.tsx
```diff
@@ -75,3 +77,15 @@ export const BatchSelect = ({ value, onChange }: Props) => {
     />
   );
 };
+
+function hideChangeNamespaceOperationIfNamespacesAreDisabled(
```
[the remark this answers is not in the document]
StaNov: Thanks for the catch.

On the mentioned page, they state that there is an eslint plugin `eslint-plugin-react-hooks` catching breakings of those rules. And we seem to use the plugin in our codebase. Do you have an idea why it didn't work in the pipeline?

e2e/cypress/common/apiCalls/testData/testData.ts
```diff
@@ -126,13 +126,14 @@ export type PermissionsOptions = {
 };
 
 export const generatePermissionsData = {
-  generate: (options: Partial<PermissionsOptions>) => {
+  generate: (options: Partial<PermissionsOptions>, useNamespaces = false) => {
```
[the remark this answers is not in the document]
StaNov: This was tricky, because everything in `PermisionsOptions` is iterable and it gets iterated over below.

I tried various approaches and wasn't successful, so I decided to use a new parameter.

JanCizmar (changes requested):
Good job!

backend/data/src/main/kotlin/io/tolgee/model/Project.kt
```diff
@@ -125,6 +125,9 @@
   @ColumnDefault("true")
   override var icuPlaceholders: Boolean = true
 
+  @ColumnDefault("false")
+  var useNamespaces: Boolean = false
```
JanCizmar, line 129: You have to add @ActivityLoggedProp and also check the activity log in the UI whether all is rendered correctly.

JanCizmar (changes requested):
One more thing!

backend/data/src/main/kotlin/io/tolgee/model/Project.kt
StaNov: Yes, thank you, I debugged to it the hard way :smile:

e2e/cypress/common/apiCalls/testData/testData.ts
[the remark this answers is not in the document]
stepan662: I think you should be able to add it as an iterable as well, it's iterable because generally the searchParams can have multiple values, but I would reuse the options object that is already there.

webapp/src/views/projects/import/ImportView.tsx
StaNov: Created a new component for translated warning box.

StaNov:
Merging despite of changes requested since we discussed them personally with @JanCizmar.

PR #2794, opened 2024-12-19
fix: Telemetry blocks startup when offline

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/service/TelemetryService.kt
```diff
@@ -27,12 +30,17 @@ class TelemetryService(
...
-      Unit::class.java,
-    )
+    try {
+      httpClient.requestForJson(
+        "${telemetryProperties.server}/v2/public/telemetry/report",
+        data,
+        HttpMethod.POST,
+        Unit::class.java,
+      )
+    } catch (e: Throwable) {
```
Anty0, line 40:
``` suggestion
    } catch (e: Exception) {
```

Catch exception only. Catching errors as well is probably a bad idea.

Anty0 (changes requested):

backend/data/src/main/kotlin/io/tolgee/service/TelemetryService.kt
JanCizmar: Noob question... but... Why is it bad idea? 🤔

Anty0: It is usually discouraged because it is generally impossible to recover from such errors programmatically anyway, and ignoring them (by catching them) only makes the issue worse. See: https://stackoverflow.com/a/2274130.

PR #2807, opened 2025-01-06
feat: Revert "REVIEWED" to "TRANSLATED" when text changes

JanCizmar:


backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/TranslationsControllerModificationTest.kt
```diff
@@ -342,6 +342,7 @@ class TranslationsControllerModificationTest : ProjectAuthControllerTest("/v2/pr
       ),
     ).andAssertThatJson {
       node("translations.en.outdated").isEqualTo(false)
+      node("translations.en.state").isEqualTo("TRANSLATED")
```
stepan662, line 345: Could we add some more tests for imports/batch translations? There were issues with this in the past.

stepan662 (changes requested):

stepan662 (approved):

Issue #2577, opened 2024-10-11
Notifications core


JanCizmar:
## Pre-specification
- [ ] We've setup responsible person
- [ ] We've checked how other tools solve the similar issues - @ZuzanaOdstrcilova 
- [ ] We've got users who require this feature and will pilot it and consult it (or we are the users) - Tolgee company itself

## Specification
- [ ] The user is able to see the bell in the top bar
- [ ] They can change notifications for each projects individually
- [ ] They can choose whether they want to be notified in app or via e-mail
- [ ] Notify users about
  - [ ] New task created
  - [ ] Task completed
  - [ ] Security stuff (password updated, MFA enabled)
  - [ ] Mentioned in translation comment  

## What's not in the scope
- [ ] Grouping - should be done on Activity Groups level
- [ ] Notifications about localization data edits (key / translation manipulation)
  -  Will be done with Activity Grouping / Activity improvements 

## UI Design Proposal
- [ ] @ZuzanaOdstrcilova can you please link here the design proposal

### The user stories are implemented
- [ ] ... Copy the stories from specification here ...

### Implementation checklist
- [ ] The code is well tested
  - [ ] User stories are covered by tests
    - [ ] Backend
    - [ ] e2e
  - [ ] ... Specify the test cases here, so you don't forget about important stuff ...
- [ ] I've checked that all endpoints are well secured
- [ ] The swagger documentation is OK including descriptions
- [ ] The documentation is updated
- [ ] Telemetry data is sent to PostHog

## Before release checklist
- [ ] Preview is created & user stories are tested on the preview 
- [ ] Blog post about this new feature
- [ ] Announcement in Tolgee App
- [ ] New keys are translated

## After release checklist
- [ ] Roadmap is updated
- [ ] We posted about this in social media
  - [ ] Twitter 
  - [ ] Linkedin
  - [ ] Facebook
  - [ ] Reddit
  - [ ] Slack Community
  - [ ] Github Discussions
- [ ] In App pricing is updated
- [ ] Web pricing is updated
- [ ] FAQ is updated
- [ ] Website is updated

ZuzanaOdstrcilova:
Style of emails: 
https://github.com/tolgee/tolgee-platform/issues/2710
we can discuss it together

ZuzanaOdstrcilova:
link to design of activity:
[...]

PR #2813, opened 2025-01-07
feat: Notifications MVP

StaNov:
Partially implements #2577.


backend/data/src/main/kotlin/io/tolgee/model/Notification.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@Entity
+@Table(
+  indexes = [
+    Index(columnList = "user_id"),
+  ],
+)
+class Notification(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
```
JanCizmar, line 15: Extend "StandardAuditModel" instead.

backend/data/src/main/kotlin/io/tolgee/service/NotificationService.kt
```diff
@@ -0,0 +1,21 @@
...
+    return notificationRepository.fetchNotificationsByUserId(userId)
+      .map { NotificationModel(it.id, it.linkedTask?.project?.id, it.linkedTask?.number, it.linkedTask?.name) }
+  }
+}
+
+data class NotificationModel(
+  val id: Long,
+  val linkedProjectId: Long?,
+  val linkedTaskNumber: Long?,
+  val linkedTaskName: String?,
```
JanCizmar: Provide the task using the TaskModelAssembler and TaskModel. Otherwise we will provide many different shapes in the API, which we don't want. Also that way you can re-use the existing model assemblers.

JanCizmar: Also the Models ahould be always created using their model assemblers in separate according to the conventions.

ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/NotificationsController.kt
```diff
@@ -0,0 +1,39 @@
...
+@Tag(name = "Notifications", description = "Manipulates notifications")
+class NotificationsController(
+  private val notificationService: NotificationService,
+  private val authenticationFacade: AuthenticationFacade,
+) {
+
+  @GetMapping
+  @Operation(summary = "Get notifications")
+  @AllowApiAccess
+  fun getNotifications(): NotificationsResponse {
```
JanCizmar: Make it pageable and use cursor, so users can scroll trough notifications forever.

ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/NotificationsController.kt
```diff
@@ -0,0 +1,39 @@
...
+    return NotificationsResponse(
+      notificationService.getNotifications(authenticationFacade.authenticatedUser.id)
+    )
+  }
+}
+
+data class NotificationsResponse(
```
JanCizmar: Don't use custom wrappers, instead follow HATEOAS conventions.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/TaskAssigneeUpdatedNotificationListener.kt
```diff
@@ -0,0 +1,53 @@
...
+          linkedTask = entity,
+          originatingUser = entity.author,
+          project = entity.project,
+        )
+      )
+    }
+  }
+
+  private fun getNewAssignees(event: OnEntityPreUpdate, entity: Task) : List<UserAccount> {
+    val assigneesIndex = event.propertyNames?.indexOf("assignees")
```
JanCizmar, line 41: It might maybe make more sense to trigger the event manually in the method which takes care of the creating / updating task to get properly typed event data. 

Also there is the OnProjectActivityEvent, which I would like stick to using. It's executed before the commit, so it should be OK for this case. However, I would still probably prefer to trigger it manually for this case if you don't really have super good reasons for this.

settings.gradle
```diff
@@ -61,8 +61,8 @@ dependencyResolutionManagement {
             library('amazonSTS', "software.amazon.awssdk:sts:$amazonAwsSdkVersion")
             library('amazonTranslate', "software.amazon.awssdk:translate:$amazonAwsSdkVersion")
             library('googleCloud', "com.google.cloud:libraries-bom:24.0.0")
-            library('liquibaseCore', "org.liquibase:liquibase-core:4.29.2")
-            library('liquibaseHibernate', "org.liquibase.ext:liquibase-hibernate6:4.29.2")
+            library('liquibaseCore', "org.liquibase:liquibase-core:4.26.0")
```
JanCizmar, line 64: Can you please revert these changes of the liquibase versions "4.29.2", but submit the revert of the "org.liquibase.gradle:org.liquibase.gradle.gradle.plugin" as the separate PR?

backend/data/src/main/kotlin/io/tolgee/service/NotificationService.kt
```diff
@@ -0,0 +1,21 @@
...
+) {
+  fun getNotifications(userId: Long): List<NotificationModel> {
+    return notificationRepository.fetchNotificationsByUserId(userId)
+      .map { NotificationModel(it.id, it.linkedTask?.project?.id, it.linkedTask?.number, it.linkedTask?.name) }
+  }
+}
+
+data class NotificationModel(
+  val id: Long,
+  val linkedProjectId: Long?,
```
JanCizmar: I would name it just "projectId". So we keep the "linked" prefix only for the "notification event data"

backend/data/src/main/kotlin/io/tolgee/model/Notification.kt
```diff
@@ -0,0 +1,31 @@
...
+  var id: Long = 0L,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  var user: UserAccount,
+
+  var createdAt: Date = Date(),
```
JanCizmar, line 19: The notification type is missing.

backend/data/src/main/kotlin/io/tolgee/repository/NotificationRepository.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.stereotype.Repository
+
+@Repository
+@Lazy
+interface NotificationRepository : JpaRepository<Notification, Long> {
+  @Query(
+    """
+    SELECT n
+     FROM Notification n
+     LEFT JOIN FETCH n.project
```
JanCizmar, line 16: IMO, you don't need to fetch the project

JanCizmar (changes requested):
Can we make a `notifications` branch. Where we would commit instead of to main, so I don't have to check the same changes when we are going to release it?

THX!

StaNov:
> Can we make a `notifications` branch. Where we would commit instead of to main, so I don't have to check the same changes when we are going to release it?
> 
> THX!

I would released it in this scope after a bit of additional polish and then added the new features in separate PRs.

According to continuous delivery of value.

The problem with the need to review everything again would therefore be bypassed.

I am open for discussion.

Tolgee

backend/data/src/main/kotlin/io/tolgee/repository/NotificationRepository.kt
StaNov: @JanCizmar Would you suggest to save only `projectId` instead of the project reference?

Or keep the project reference and just keep it lazy? I wouldn't do this because we need the project ID everytime so it would fetch the whole project. Or maybe I lack some hibernate skills.

StaNov: `SimpleProjectModel` fetching implemented.

ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/NotificationsController.kt
StaNov: Paging implemented on BE with fetching everything on FE for now.

I would implement the cursor stuff later, I put it on my roadmap.

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationService.kt
```diff
@@ -0,0 +1,23 @@
...
+  private val notificationRepository: NotificationRepository,
+) {
+  fun getNotifications(
+    userId: Long,
+    pageable: Pageable,
+  ): Page<Notification> {
+    return notificationRepository.fetchNotificationsByUserId(userId, pageable)
+  }
+
+  fun createNotification(notification: Notification) {
```
JanCizmar, line 20: This should be named `save`, create usually transforms Dto to the entity and saves it or does some advanced logic.

backend/data/src/main/kotlin/io/tolgee/repository/NotificationRepository.kt
```diff
@@ -0,0 +1,30 @@
...
```
JanCizmar, line 18: Because we have the enhancers, you probably don't want to fetch it, right?

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/TaskNotificationEnhancer.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+class TaskNotificationEnhancer(
+  private val taskService: TaskService,
+  private val taskModelAssembler: TaskModelAssembler,
+) : NotificationEnhancer {
+  override fun enhanceNotifications(notifications: Map<Notification, NotificationModel>) {
+    val taskIds = notifications.mapNotNull { (source, _) -> source.linkedTask?.id }
+    val convertedTasks = taskService.getTasksWithScope(taskIds)
```
JanCizmar, line 17: it wold be better to use .associateBy{it.id} on it and then to just get it from the map instead of searching in it using fing.

JanCizmar: Btw. Task have Id, so you should use it. Don't use the project id and number, because it will require you fetch it or it will be n+1.

JanCizmar (changes requested):

backend/data/src/main/kotlin/io/tolgee/repository/NotificationRepository.kt
StaNov: True, removed.

JanCizmar (approved):

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/TaskNotificationEnhancer.kt
StaNov: Hibernate log looks reasonable also, no `n+1`'s.

PR #2814, opened 2025-01-07
fix: move notifications to top-right corner

stepan662:


JanCizmar (approved):

JanCizmar:
Maybe for the future we can add close button and prevent duplicate?

https://notistack.com/features/basic#prevent-duplicate

PR #2821, opened 2025-01-08
chore: Retry `npm install` to make CI builds more stable

StaNov:
[Example of an intermittently failing build](https://github.com/tolgee/tolgee-platform/actions/runs/12672324621/job/35316100659)


JanCizmar:
I will close this, since it just adds complexity and reduces readability to the code. If we retry something we have to be sure why we do that.

PR #2840, opened 2025-01-14
feat: Notifications badge

StaNov:


webapp/src/ThemeProvider.tsx
```diff
@@ -290,6 +290,13 @@ const getTheme = (mode: PaletteMode) => {
           },
         },
       },
+      MuiBadge: {
+        styleOverrides: {
+          colorSecondary: {
+            color: c.background,
+          },
+        },
+      },
```
stepan662, line 299: This will change the color globally, which is probably not what you want.

webapp/src/component/layout/TopBar/Notifications.tsx
```diff
@@ -63,7 +64,12 @@ export const Notifications: FunctionComponent<{ className?: string }> = () => {
         onClick={handleOpen}
         size="large"
       >
-        <Bell01 />
+        <Badge
+          badgeContent={notifications?.page?.totalElements}
+          color="secondary"
```
stepan662, line 69: If you need some specific color of the badge, you can use the `sx` param or wrap the Badge component with `styled` function and you can override the styles as you want.

PR #2842, opened 2025-01-15
chore: Run `npm install` before localhost FE start in Idea

StaNov:


JanCizmar:
I believe that this can be effectively handled by gradle, which might include caching and not runing npm install when up to date.

StaNov:
> I believe that this can be effectively handled by gradle, which might include caching and not runing npm install when up to date.

@JanCizmar This is about running it from the IDEA run menu where we currently run it as a direct `npm run` task, not as a gradle task.

image

JanCizmar:
Well yeah, but this does not work for anyone but you. When you have the setup for billing it just doesn't work. So I would't add new tasks to idea configuration until this issue is resolved.

We use Gradle for managing task dependencies. Not IDEA. Now we are mixing multiple different approaches, which will get messy. Gradle approach is universal, so you can use it even when you don't use IDEA.

StaNov:
@JanCizmar Ok, let's do it by gradle, closing.

PR #2843, opened 2025-01-15
feat: Notifications real-time update with websockets

StaNov:


JanCizmar (approved):

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationService.kt
```diff
@@ -19,5 +23,19 @@ class NotificationService(
 
   fun save(notification: Notification) {
     notificationRepository.save(notification)
+    websocketEventPublisher(
+      "/users/${notification.user.id}/${WebsocketEventType.NOTIFICATIONS_CHANGED.typeName}",
+      WebsocketEvent(
+        actor = ActorInfo(
```
StaNov, line 29: @JanCizmar I am not sure about this field, but it was not-nullable.

Should it be the user who created the task, in this case?

JanCizmar: You can make it nullable.

backend/app/src/test/kotlin/io/tolgee/websocket/AbstractWebsocketTest.kt
```diff
@@ -184,39 +200,80 @@ abstract class AbstractWebsocketTest : ProjectAuthControllerTest("/v2/projects/"
     }
   }
 
+  @Test
+  @ProjectJWTAuthTestMethod
+  fun `notifies user on change of his notification`() {
```
JanCizmar, line 205:
```suggestion
  fun `notifies user on change of his notifications`() {
```

JanCizmar (approved):

PR #2846, opened 2025-01-15
feat: new screenshots view

stepan662:


webapp/src/views/projects/translations/Screenshots/Screenshots.tsx
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  return (
+    <StyledContainer
+      {...{ sx }}
+      className={clsx({ scrollLeft, scrollRight })}
+      onClick={stopAndPrevent()}
+    >
+      <StyledScrollWrapper ref={containerRef}>
+        {screenshots.map((sc, index) => {
```
JanCizmar, line 224: Can we move this to separate component so the file is not so big and we don't have to do so much logic in the JSX?

JanCizmar (approved):

JanCizmar:
Before we merge it. Wouldn't it makes sense to limit the height of the screenshot so the key doesn't have so large height? In most cases there is just one screenshot and most of the people have enabled just 2 languages, so by this we very probably make it much less keys visible on the single page scroll area. @ZuzanaOdstrcilova 

image

stepan662:
Yeah, sounds good. There is a limit of 350px. But I'd make it lower, like 200px

ZuzanaOdstrcilova:
We can make it smaller. If we want more keys on one page, I would try even 150.

JanCizmar:
I didn't see it, but I would probably also vote for less then more. I know it's a tradeoff, but the users currently appreciate the clarity of the view, so I don't want to break it by adding a lot of empty areas.

JanCizmar (approved):

PR #2851, opened 2025-01-16
feat: Mark notifications as seen + show count of unseen only

StaNov:


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/NotificationController.kt
```diff
@@ -37,8 +36,33 @@ class NotificationController(
...
+  @Operation(summary = "Marks notifications of the currently logged in user with given IDs as seen.")
+  @AllowApiAccess
+  fun markNotificationsAsSeen(
+    @RequestBody notificationIds: List<Long>,
+  ) {
+    notificationService.markNotificationsAsSeen(notificationIds, authenticationFacade.authenticatedUser.id)
+  }
+}
+
+class NotificationPagedModel(
```
JanCizmar, line 57: Move the model to proper directory.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/NotificationController.kt
```diff
@@ -37,8 +36,33 @@ class NotificationController(
...
+  ): NotificationPagedModel {
     val notifications = notificationService.getNotifications(authenticationFacade.authenticatedUser.id, pageable)
-    return pagedResourcesAssembler.toModel(notifications, NotificationModelAssembler(enhancers, notifications))
+    val unseenCount =
+      notificationService.getCountOfUnseenNotifications(authenticationFacade.authenticatedUser.id)
+    val pagedNotifications = pagedResourcesAssembler.toModel(notifications, NotificationModelAssembler(enhancers, notifications))
+    return NotificationPagedModel.of(pagedNotifications, unseenCount)
+  }
+
+  @PutMapping("/mark-seen")
```
JanCizmar, line 47: This is not correctly restful. The `/v2/notifications/{id}` has to be reserved for manipulating notification with specific id. You will have to change it so the url is `/v2/mark-notifications-seen`.

JanCizmar (changes requested):

StaNov:
@JanCizmar Comments addressed, please re-review.

backend/data/src/main/kotlin/io/tolgee/model/Notification.kt
```diff
@@ -21,4 +22,7 @@ class Notification : StandardAuditModel() {
 
   @ManyToOne(fetch = FetchType.LAZY)
   var linkedTask: Task? = null
+
+  @ColumnDefault("false")
+  var beenSeen: Boolean = false
```
JanCizmar, line 27: I would call it just `seen`

backend/api/src/main/kotlin/io/tolgee/hateoas/notification/NotificationPagedModel.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import org.springframework.hateoas.Link
+import org.springframework.hateoas.Links
+import org.springframework.hateoas.PagedModel
+
+class NotificationPagedModel(
+  content: Collection<NotificationModel> = emptyList(),
+  metadata: PageMetadata? = null,
+  links: Iterable<Link> = Links.NONE,
+  val unseenCount: Int,
```
JanCizmar, line 11: I don't think creating special model is the right way to do that. In sake of keeping the API consistent, I would much rather add filter for seen notifications. That way, you will get unseen count in the "total count" in the page info object.

webapp/src/component/layout/TopBar/Notifications.tsx
```diff
@@ -40,9 +40,31 @@ const ListItemHeader = styled(ListItem)`
 `;
 
 export const Notifications: FunctionComponent<{ className?: string }> = () => {
+  const notificationsLoadable = useApiQuery({
```
JanCizmar, line 43: It would be good idea to pass the initial notifications count in the initial data and then create some mechanism to poll it periodically like every 20 minutes or when the user reconnects, so we always get the values accurate.

We're trying to keep the amount of data loaded on the page loaded minimal, so we add the initial reuqirements to the initial data endpoint.

JanCizmar (changes requested):

webapp/src/component/layout/TopBar/Notifications.tsx
StaNov: Discussed in person - the next PR will add extended support for websockets which will handle the "polling" for updates.

backend/data/src/main/kotlin/io/tolgee/dtos/request/notification/NotificationFilters.kt
```diff
@@ -0,0 +1,5 @@
+package io.tolgee.dtos.request.notification
+
+class NotificationFilters(
+  val unseenOnly: Boolean = false,
```
JanCizmar, line 4: The convention is to name it `filterSeen`. When you make it nullable you can have this options:
- null: make this filter ignored
- true: filter seen
- false: filter unseen

see `io.tolgee.repository.UserAccountRepositoryKt#USER_FILTERS` for reference

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/NotificationsTestData.kt
```diff
@@ -1,26 +1,38 @@
 package io.tolgee.development.testDataBuilder.data
 
+import io.tolgee.development.testDataBuilder.builders.NotificationBuilder
+
 class NotificationsTestData : BaseTestData() {
+  private var nextTaskId = 100000L
```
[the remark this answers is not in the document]
StaNov: Yes, it should be `nextTaskNumber`, thank you.

There are some more tasks created with the parent test data builder, therefore there would be conflicts with those. Would you suggest another approach consistent with the current state?

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/builders/TestDataBuilder.kt
```diff
@@ -28,6 +28,7 @@ class TestDataBuilder(fn: (TestDataBuilder.() -> Unit) = {}) {
   fun addUserAccountWithoutOrganization(ft: UserAccount.() -> Unit): UserAccountBuilder {
     val builder = UserAccountBuilder(this)
     data.userAccounts.add(builder)
+    builder.self.name = "Test User without organization " + System.currentTimeMillis()
```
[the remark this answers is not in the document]
StaNov: The name of user is mandatory. It may be beneficiary to have a default name for other tests which need to create a user and don't care how it's going to be called.

The appended milliseconds are for debugging reasons - to differentiate between users.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/NotificationsTestData.kt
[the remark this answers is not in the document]
StaNov: Removed.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/builders/TestDataBuilder.kt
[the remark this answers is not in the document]
StaNov: Removed.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/notification/NotificationOperationsController.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.web.bind.annotation.*
+
+@RestController
+@CrossOrigin(origins = ["*"])
+@RequestMapping(
+  value = [
+    "/v2/",
+  ],
+)
+@Tag(name = "Notification Operations", description = "Operations with notifications")
```
JanCizmar, line 17: You can keep the same tag, so it's not in different group in the UI.

```suggestion
@Tag(name = "Notifications", description = "Manipulates notifications")
```

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/notification/NotificationController.kt
```diff
@@ -37,8 +38,17 @@ class NotificationController(
   @AllowApiAccess
   fun getNotifications(
     @ParameterObject pageable: Pageable,
```
JanCizmar, line 40: Don't forget to add @SortDefault, so we don't need to manually send it in all the test and the endpoint doesn't send different data when no sort is provided.

e2e/cypress/e2e/tasks/tasksNotifications.cy.ts
```diff
@@ -56,6 +56,10 @@ describe('tasks notifications', () => {
       .should('contain', 'New review task');
     dismissMenu();
     cy.gcy('notifications-button').click();
+
+    cy.gcy('notifications-count')
+      .find('.MuiBadge-badge')
```
JanCizmar, line 61: Don't rely on the MUI classes. Rather use your own data-cy.

JanCizmar (changes requested):

JanCizmar (approved):

PR #2856, opened 2025-01-20
chore: Keep running fakesmtp even after system restart

StaNov:
For easier local development - the only needed step is to run the cypress window. It doesn't eat significant amount of resources anyway.

### DIFF
--- a/e2e/docker-compose.yml
+++ b/e2e/docker-compose.yml
@@ -12,6 +12,7 @@ services:
       - tolgee.frontend-url=http://localhost:8201
   fakesmtp:
     image: reachfive/fake-smtp-server:0.8.1
+    restart: unless-stopped
     ports:
       - "21025:1025"
       - "21080:1080"


JanCizmar:
I believe that the right solution is to setup compound run in idea, not keeping a service running when system is restarted. 

```
<component name="ProjectRunConfigurationManager">
  <configuration default="false" name="Billing E2E" type="CompoundRunConfigurationType">
    <toRun name="Open E2E for Billing" type="GradleRunConfiguration" />
    <toRun name="Run Dokcer E2E dependencies" type="GradleRunConfiguration" />
    <toRun name="Run the Frontend (WebApp) for E2E " type="GradleRunConfiguration" />
    <toRun name="Backend for E2E Billing" type="SpringBootApplicationConfigurationType" />
    <method v="2" />
  </configuration>
</component>
```

StaNov:
@JanCizmar Thank you for your comment. That is also an option.

I'd close it for now and continue with it when there is some time.

PR #2860, opened 2025-01-22
feat: Websockets optimizations for notifications

StaNov:
* Sending new notification in websocket instead of just ping.
* Not loading notifications before first opening of the notification list.


e2e/cypress/common/notifications.ts
JanCizmar, line 1: Would me nice to wrap it with a class. Then you would be able to remove the word "notifications" from all the  methods.

JanCizmar (approved):

PR #2862, opened 2025-01-23
chore: Making intermittent test stable

StaNov:
It was failing in the builds a lot when the loading took more than the static 50 ms. This change waits for the element to be visible therefore making it more stable.


JanCizmar (approved):
Are you sure it fixes the issue?

PR #2864, opened 2025-01-23
chore: Distinct names for E2E artifacts

StaNov:
`actions/upload-artifact@v4` does not allow multiple artifacts with the same name which causes upload failures in case of multiple E2E suites failing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -255,7 +255,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: e2e
+          name: e2e_${{ matrix.job_index }}
           path: |
             ./e2e/cypress/videos/**/*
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -255,7 +255,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: e2e
+          name: e2e_${{ matrix.job_index }}
```
JanCizmar, line 258: It merges, doesn't it? So we don't have to add the index.

StaNov: It doesn't merge. It fails because an artifact with the same name already exists. At least in the v4 version, I don't know how it was behaving before.

JanCizmar: OK, it merged before.

JanCizmar (approved):

PR #2865, opened 2025-01-23
feat: Send notification on task completed, MFA enabled+disabled, password chaged

StaNov:


webapp/src/component/layout/TopBar/Notifications.tsx
```diff
@@ -96,6 +96,23 @@ const NotificationItemLinkedDetailNumber = styled(
       : theme.palette.emphasis[600]};
 `;
 
+function getLocalizedMessage(
+  notification: components['schemas']['NotificationModel']
+) {
+  switch (notification.type) {
```
JanCizmar, line 102: I will approve it now, but I would much rather see this implemented "composition way", so instead of trying to create one Notification component handling al the possible notification, I would create separate components for each notification type. 

We did this mistake with activity and now it's hard to maintain, bacause there is lot of switches all around. 

I believe we should reduce the amount of switches like this and enable the future developers to have as much freedom as possible.

JanCizmar (approved):

webapp/src/component/layout/TopBar/Notifications.tsx
StaNov: @JanCizmar I agree.

PR #2870, opened 2025-01-24
feat: Notifications basic paging

StaNov:


webapp/src/component/layout/TopBar/Notifications.tsx
```diff
@@ -120,33 +125,49 @@ export const Notifications: FunctionComponent<{ className?: string }> = () => {
...
+  const notificationsLoadable = useApiInfiniteQuery({
     url: '/v2/notifications',
     method: 'get',
-    query: { size: 10000 },
-    options: { enabled: false },
+    query: query,
+    options: {
+      enabled: false,
+      getNextPageParam: (lastPage) => {
+        if (
```
JanCizmar, line 144: To me, this is a bit hard to read. I would avoid the else by first checking if we are on the last page already by returning null and then I would continue with the standard case withot else branch.

JanCizmar (approved):

webapp/src/component/layout/TopBar/Notifications.tsx
StaNov: @JanCizmar This is the standard way we use at multiple other places.

I'd thought about it more when implementing the cursor if that's ok with you.

PR #2872, opened 2025-01-27
fix: Reset password callback from frontend URL

JanCizmar:


Anty0 (changes requested):

backend/api/src/main/kotlin/io/tolgee/controllers/resetPassword/ResetPasswordRequestHandler.kt
```diff
@@ -87,25 +91,45 @@ class ResetPasswordRequestHandler(
...
-    val callbackString = code + "," + request.email
-    return getCallbackUrlBase() + "/" + Base64.getEncoder().encodeToString(callbackString.toByteArray())
+    return getCallbackUrlBase()?.removeSlashSuffix() + "/" + getEncodedCodeString(code)
   }
 
   private fun getCallbackUrlBase(): String? {
-    return request.callbackUrl
+    if (frontEndUrlFromProperties == null) {
+      logger.warn(
+        "Frontend URL is not set in properties. Using frontend URL from the request can lead to single click " +
```
JanCizmar, line 100:
```suggestion
        "Frontend URL is not set in properties. Using frontend URL from the request can lead to one-click account takeover" +
```

Anty0 (approved):

PR #2883, opened 2025-01-29
feat: Allow switching to SSO provider

### LINKED ISSUE #2619: SSO > Changing third-party provider & improvements
- [ ] When a user tries to log in with a third-party provider, but his account is not set up for that provider, the user receives an error saying that the account already exists. Instead, we should allow the user to switch his account to that third-party login provider. We can ask the user to reenter his password and replace `thirdPartyAuthType` with the new provider.
- [x] Prevent Native login when using SSO
   - Once user switches to SSO we don't want them to use their native login, since in that case the user would be able to login also when they are not part of the company anymore
- [ ] Prevent native login for the domains used for SSO
   - When user registers to Tolgee with SSO domain, we have to prevent them from signing up with such e-mail
- [ ] Correctly prevent password change in the UI with disabling of the password with correct `<Alert>` banner.
- [ ] Enable user to transfer from Native to SSO

Anty0:
Fixes #2619


backend/api/src/main/kotlin/io/tolgee/security/thirdParty/GithubOAuthDelegate.kt
```diff
@@ -112,6 +103,39 @@ class GithubOAuthDelegate(
     throw AuthenticationException(Message.THIRD_PARTY_AUTH_UNKNOWN_ERROR)
   }
 
+  fun findAccount(
```
JanCizmar, line 106: This also creates account so it should be named findOrCreateAccount

backend/api/src/main/kotlin/io/tolgee/security/thirdParty/GoogleOAuthDelegate.kt
```diff
@@ -82,25 +87,10 @@ class GoogleOAuthDelegate(
           }
         }
 
-        val googleEmail = userResponse.email ?: throw AuthenticationException(Message.THIRD_PARTY_AUTH_NO_EMAIL)
+        val userAccount = findAccount(userResponse, invitationCode)
```
JanCizmar, line 90: Same as above

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/SsoGlobalProperties.kt
```diff
@@ -26,6 +26,9 @@ class SsoGlobalProperties : ISsoTenant {
   @DocProperty(description = "Enables SSO authentication on global level - as a login method for the whole server")
   var enabled: Boolean = false
 
+  @DocProperty(description = "Force sso authentication for all users using matching e-mail domain")
```
JanCizmar, line 29: Is this used now?

backend/data/src/main/kotlin/io/tolgee/dtos/request/auth/AuthProviderChangeData.kt
```diff
@@ -0,0 +1,31 @@
...
+  var userAccount: UserAccount,
+  var accountType: UserAccount.AccountType,
+  var authType: ThirdPartyAuthType,
+  var authId: String? = null,
+  var ssoDomain: String? = null,
+  var ssoRefreshToken: String? = null,
+  var ssoExpiration: Date? = null,
+) {
+  fun asAuthProviderChangeRequest(expirationDate: Date): AuthProviderChangeRequest {
+    return AuthProviderChangeRequest().also {
```
JanCizmar, line 20: Mapping to entity should not be done here. Rather, do it in the service.

webapp/src/component/security/AcceptAuthProviderChangeView.tsx
```diff
@@ -0,0 +1,200 @@
...
+                <LoadingButton
+                  loading={acceptChange.isLoading}
+                  variant="contained"
+                  color="primary"
+                  onClick={handleAccept}
+                  data-cy="accept-auth-provider-change-accept"
+                >
+                  {t('accept_auth_provider_change_accept')}
+                </LoadingButton>
+                {/*<Button
```
JanCizmar, line 179: Remove the commented out code

JanCizmar (changes requested):

PR #2884, opened 2025-01-29
feat: Notifications, visual fixes

StaNov:


webapp/src/component/layout/Notifications/NotificationItem.tsx
```diff
@@ -44,14 +53,19 @@ const StyledProject = styled(StyledTime)`
...
-  isLast: boolean;
   destinationUrl?: string;
 } & ListItemButtonProps;
 
+function postProcessTimeText(input: string) {
+  return input
+    .replaceAll(/\b(\d+?)\b ([Mm])inut.*?\b/g, '$1 $2in')
+    .replaceAll(/\b(\d+?)\b hodin.*?\b/g, '$1 hod')
+    .replaceAll(/\b(\d+?)\b (sekund|second|segund).*?\b/g, '$1 s');
+}
```
stepan662, line 64: Hmm, this doesn't look very clean. Is it necessary? If we keep it there should be some comment about it.

webapp/src/hooks/useWindowInnerHeight.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
+
+export function useWindowInnerHeight() {
```
stepan662, line 3: You can use 
```
import { useWindowSize } from 'usehooks-ts';
```
instead of this

webapp/src/component/layout/Notifications/NotificationItem.tsx
JanCizmar: I hope this is some kind of joke. Can you please explain what it does?

JanCizmar: Anyway, this has to very probably have to go as separate translations with proper pluralization.

JanCizmar (changes requested):

webapp/src/component/layout/Notifications/NotificationItem.tsx
StaNov: Thanks for your comments, guys.

The reasoning is this:

* The requirement from design was to show the age of the notification.
* We are already using the `date-fns` component at other places, so I decided to use it also here.
* After another design iteration, it was found that `15 minutes ago` is too long and it would be better to abbreviate `minutes` to just `min`, `seconds` to `s` and `hours` to `h`.
* The `date-fns` component doesn't offer a simple configuration for this (as far as I researched) besides of creating new `date-fns-locale`s for this case. It is supplied with two dates and a locale and it returns a localized string.
* So my low-level effort to at least prototype it to see it on the screen was this - take the output string and shorten the long words to the abbreviated ones.
* After another design iteration, it might be better to drop using the localized text and go for just `15m` or `7d`.

obrazek

JanCizmar: I am for making it simplier...

It doesn't read well. Classic meme issue with reading row first. 

I would remove all the not-necessary information.

- we don't need the language there (the user probably knows which languages they know)
- the user don't need to know task number
- the user probably doesn't need the project name, because there is usually single project
  - Actually, instead of showing the user avatar, I would show the project avatar, if it's project event. I am not that interested about the information who assigned me. (usually it will be my single manager)

Anyway, they will find out when they open it.

The notifications should not replicate the function of activity log. Notifications only notify. It's role is not to inform about everything.

The time can be super short format. Like facebook and reddit have.

Reddit example:
image

Facebook example:

image

webapp/src/hooks/useWindowInnerHeight.ts
StaNov: Thanks, updated.

webapp/src/component/layout/Notifications/NotificationItem.tsx
StaNov: I put away the confusing regex shortenings.

Could we please merge it as it is for now since the fixes are according to the current design?

JanCizmar (approved):

PR #2887, opened 2025-02-02
feat: OpenAI translation support

ftp27:
Hi,

Thank you for the great project! GPT translations seem to be essential nowadays, and I really wanted to use them with your app.
I'm not familiar with Kotlin—I just took a glimpse at the implementation of other providers and made it look similar.


JanCizmar:
Hello! Thanks for the PR. 

I cannot accept it, since the Tolgee AI translator is based on OpenAI, it has many more features then simple translation as implemented here.

The current plan is to open-source the Tolgee AI translator service and make it part of the platform, similarly what you did here, but with all the Tolgee AI Translator features.

@stepan662 might use this PR as an inspiration.

@ftp27 Thanks a lot for this! 🙏

PR #2888, opened 2025-02-03
feat: Notification emails

StaNov:


backend/data/src/main/kotlin/io/tolgee/service/notification/EmailNotificationsService.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    try {
+      tolgeeEmailSender.sendEmail(params)
+    } catch (e: Exception) {
+      logger.error(e.message)
+      Sentry.captureException(e)
+    }
+  }
+
+  private fun composeEmailSubject(notification: Notification) =
```
JanCizmar, line 46: I would move the next 3 methods to separate class returning the subject and body of the e-mail.

backend/data/src/main/kotlin/io/tolgee/service/notification/EmailNotificationsService.kt
```diff
@@ -0,0 +1,117 @@
...
+    when (notification.type) {
+      NotificationType.TASK_ASSIGNED -> "New task assignment"
```
JanCizmar, line 48: I would move this texts to `resources/I18n_en.properties` so it's prepared for localization

e2e/cypress/e2e/tasks/tasksNotifications.cy.ts
```diff
@@ -87,10 +96,16 @@ describe('tasks notifications', () => {
 
     assertMessage('Task updated sucessfully');
 
-    getAssignedEmailNotification().then(({ toAddress, myTasksLink }) => {
-      assert(toAddress === 'organization.member@test.com', 'correct recipient');
-      cy.visit(myTasksLink);
-    });
+    getAssignedEmailNotification().then(
```
JanCizmar, line 99: I don't think e-mails should be tested primarily using e2e tests. It's a backend thing. 

You can keep this test, but in the future I would much rather see the stuff tested on the backend.

JanCizmar (changes requested):

e2e/cypress/e2e/tasks/tasksNotifications.cy.ts
StaNov: The test was already there, I just added testing also the correct link to the task.

I agree that email testing should be done on BE, but at least one E2E is beneficial imho because it tests also the SMTP layer.

backend/data/src/main/resources/I18n_en.properties
```diff
@@ -80,4 +80,26 @@ slack.common.message.global-subscription-error=You can't set global subscription
 slack.common.message.modification-info=*%s* has changed the %s at %s
 slack.common.message.feature-not-available=This feature is not available for your plan. Please upgrade your plan to use this feature.
 slack.common.message.bot-not-in-channel=Bot is not in the channel. Please invite the bot to the channel using `/invite @Tolgee` and try again.
-slack.help.message.advanced-subscribe-events-example=`/tolgee subscribe 1 fr --on new_key` - subscribe to new key events in project 1 for French translations
\ No newline at end of file
+slack.help.message.advanced-subscribe-events-example=`/tolgee subscribe 1 fr --on new_key` - subscribe to new key events in project 1 for French translations
+
+notifications.email.subject.TASK_ASSIGNED=New task assignment
```
JanCizmar, line 85: This is out of convention. It should be 
```suggestion
notifications.email.subject.TASK_ASSIGNED=New task has been assigned to you
```

backend/data/src/main/resources/I18n_en.properties
```diff
@@ -80,4 +80,26 @@ slack.common.message.global-subscription-error=You can't set global subscription
...
+notifications.email.subject.TASK_COMPLETED=Task has been completed
+notifications.email.subject.TASK_CLOSED=Task has been closed
+notifications.email.subject.MFA_ENABLED=Multi-factor authentication has been enabled for your account
+notifications.email.subject.MFA_DISABLED=Multi-factor authentication has been enabled for your account
+notifications.email.subject.PASSWORD_CHANGED=Password has been changed for your account
+
+notifications.email.task-type.TRANSLATE=translate
+notifications.email.task-type.REVIEW=review
+
+notifications.email.task-header.TASK_ASSIGNED=Password has been changed for your account
```
JanCizmar, line 95: This is wrong

backend/data/src/main/resources/I18n_en.properties
```diff
@@ -80,4 +80,26 @@ slack.common.message.global-subscription-error=You can't set global subscription
...
+notifications.email.task-header.TASK_COMPLETED=Task you've created has been completed
+notifications.email.task-header.TASK_CLOSED=Task you've created has been closed
+
+notifications.email.my-tasks-link=Check all your tasks <a href="${myTasksUrl}">here</a>.
```
JanCizmar, line 99:
```suggestion
notifications.email.my-tasks-link=Check your tasks <a href="${myTasksUrl}">here</a>.
```

backend/data/src/main/resources/I18n_en.properties
```diff
@@ -80,4 +80,26 @@ slack.common.message.global-subscription-error=You can't set global subscription
...
+notifications.email.security-settings-link=Check your security settings <a href="${securitySettingsUrl}"}">here</a>.
+
+notifications.email.mfa.changed=Multi-factor authentication has been ${enabledDisabled} for your account.
```
JanCizmar, line 102: Introduce separate strings for both states, don't combine the strings. (It can go wrong in languages we support or might support in the future)

JanCizmar (changes requested):

backend/data/src/main/resources/I18n_en.properties
StaNov: Thanks.

Just for context - I was just moving the logic, keeping the text as they were before.

backend/data/src/main/kotlin/io/tolgee/service/notification/EmailNotificationComposer.kt
```diff
@@ -0,0 +1,81 @@
...
+
+  private fun taskLink(notification: Notification): String {
+    val task =
+      notification.linkedTask
+        ?: throw IllegalStateException(
+          "Notification of type ${notification.type} must contain linkedTask.",
+        )
+    return """
+        |<a href="${frontendUrlProvider.getTaskUrl(task.project.id, task.number)}">
+        |  ${task.name} #${task.number} (${task.language.name}) - ${taskType(task.type)}
```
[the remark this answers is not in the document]
StaNov: @JanCizmar Do you want the View in Tolgee line to be duplicated in each of the task type (assigned, completed, closed)?

JanCizmar (approved):

PR #2893, opened 2025-02-04
feat: Apple String Catalog Support

### LINKED ISSUE #2300: Feature Request: Apple String Catalog Support
Currently, it is exported in .strings & .stringsdict format and manually changed to string catalog format.
It's uncomfortable every time.
When I create a .string & .stringsdict file in Xcode, I'm notating the legacy notation.
I ask you to improve this.

neo773:
This PR adds support for Apple String Catalog Support

demo

https://github.com/user-attachments/assets/0c7ef97a-ebcf-4f42-b9e9-af65d4469ef6

/claim #2300
/closes #2300


neo773:
@JanCizmar Hi, this is ready to be reviewed

JanCizmar:
Impressive job! I am reviewing it now.

ftp27:
I believe it should use the key as the original translation. Additionally, it doesn't recognize some format specifiers and eventually breaks them during export. Here is the export result. English was the original language for the file, by the way.
image

neo773:
Will look into it, thanks!

backend/data/src/test/kotlin/io/tolgee/unit/formats/apple/in/XcstringsFormatProcessorTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    @BeforeEach
+    fun setup() {
+        mockUtil = FileProcessorContextMockUtil()
+        mockUtil.mockIt("example.xcstrings", "src/test/resources/import/apple/example.xcstrings")
+    }
+
+    @Test
+    fun `returns correct parsed result`() {
+        processFile()
+        mockUtil.fileProcessorContext.assertLanguagesCount(2)  // en, fr
```
JanCizmar, line 23: I would like to see that by default it converts `%@` and `%lld` parameters into the Tolgee Universal ICU format by default. Please add such params to the example files and test them.

backend/data/src/test/kotlin/io/tolgee/unit/formats/apple/in/XcstringsFormatProcessorTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    @Test
+    fun `import with placeholder conversion (disabled ICU)`() {
+        mockPlaceholderConversionTestFile(convertPlaceholders = false, projectIcuPlaceholdersEnabled = false)
+        processFile()
+        mockUtil.fileProcessorContext.assertLanguagesCount(1)
+        mockUtil.fileProcessorContext.assertTranslations("en", "messages-count")
+            .assertSinglePlural {
+                hasText(
+                    """
+                    {value, plural,
```
JanCizmar, line 60: Here we also need to test what happens when there is some ICU syntax by any chance in the string. Such syntax has to be escaped. 

This is tested here:
io.tolgee.unit.formats.apple.in.StringsdictFormatProcessorTest#import with placeholder conversion (disabled ICU)

backend/data/src/main/kotlin/io/tolgee/service/export/ExportFilePathProvider.kt
```diff
@@ -18,6 +18,14 @@ class ExportFilePathProvider(
     languageTag: String? = null,
     replaceExtension: Boolean = true,
   ): String {
+    // If no template is provided, use a default template for xcstrings
+    if (params.fileStructureTemplate == null) {
```
JanCizmar, line 22: This should not happen on this level. This is generic file used for all the formats. Default templates are defined here:

io/tolgee/formats/ExportFormat.kt:8

backend/data/src/main/kotlin/io/tolgee/service/dataImport/processors/FileProcessorContext.kt
```diff
@@ -86,6 +87,13 @@ data class FileProcessorContext(
         _translations[keyName]!!.removeIf { it.language == language && !it.isPlural }
       }
       _translations[keyName]!!.add(entity)
+
+      val key = getOrCreateKey(keyName)
```
JanCizmar, line 91: The generic data storing implementation should not be touch by adding new format. There already is a mechanism for adding description. Please use that:

io.tolgee.service.dataImport.processors.FileProcessorContext#addKeyDescription

JanCizmar (changes requested):

JanCizmar:
It also looks like the exported json is not formatted.

JanCizmar:
@ftp27 I believe you are using it wrong. You have to enable the conversion to Tolgee ICU placeholders when importing or you need to disable ICU support on the project configuration level. 

It seems like you have the ICU support enabled, but you turned off the conversion when importing. This leads to escaped `%` sign when exporting so `%@` goes to `%%@`.

JanCizmar:
I have prepared PR to docs and CLI:
https://github.com/tolgee/documentation/pull/841
https://github.com/tolgee/tolgee-cli/pull/153

backend/data/src/main/kotlin/io/tolgee/service/dataImport/processors/FileProcessorContext.kt
```diff
@@ -90,9 +90,7 @@ data class FileProcessorContext(
 
       val key = getOrCreateKey(keyName)
```
JanCizmar, line 91: Can you please remove this whole part and remove the metadata field completely from here? Then call the description adding from directly form your processor.

backend/data/src/main/kotlin/io/tolgee/service/export/ExportFilePathProvider.kt
```diff
@@ -60,19 +53,21 @@ class ExportFilePathProvider(
...
-        ExportFilePathPlaceholder.SNAKE_LANGUAGE_TAG,
-      ).any { getTemplate().contains(it.placeholder) }
-
-    if (!containsLanguageTag) {
-      throw getMissingPlaceholderException(
-        ExportFilePathPlaceholder.LANGUAGE_TAG,
-        ExportFilePathPlaceholder.ANDROID_LANGUAGE_TAG,
-        ExportFilePathPlaceholder.SNAKE_LANGUAGE_TAG,
-      )
+    if (params.format !in listOf(ExportFormat.APPLE_XCSTRINGS)) {
```
JanCizmar, line 56: This is incorrect modifying of file which belongs to different layer. I understand you need to bypass this check for the the xcstrings. 

You can add new boolean property `multiLanguage` to the `ExportFormat` and check for it here.

JanCizmar (changes requested):

JanCizmar:
To make ktlint pass, you need to run `./gradlew ktlintFormat` task.

backend/data/src/main/kotlin/io/tolgee/formats/importCommon/ImportFormat.kt
```diff
@@ -125,6 +125,15 @@ enum class ImportFormat(
     messageConvertorOrNull = appleConvertor,
   ),
 
+  XCSTRINGS(
```
Anty0, line 128:
```suggestion
  APPLE_XCSTRINGS(
```

To make the import and export format name consistent.
There is an inconsistency between the `APPLE_STRINGSDICT` export format and the `STRINGSDICT` import format, and I can see the current name is probably trying to match that, but let's keep the new format names consistent.

ftp27:
@JanCizmar, you're right. Without ICU, the translation has been imported as it was supposed to be. But there's still a problem with the default language during export. Also, the whole JSON isn't exported as Xcode saves it, so the git diff will look messy. It's fixable, though, by making a change in the exported document and re-saving the file.

JanCizmar:
> Also, the whole JSON isn't exported as Xcode saves it

What is the difference?

ftp27:
@JanCizmar At first glance, the JSON files from Xcode appear to be beautified, with locale keys sorted alphabetically.

I'm not sure if it's possible to replicate the original file after import. Each translation can also have a `state` with values like `needs_review` (similar to fuzzy translation) or `translated`. Some strings may be marked with `shouldTranslate: false` and can include a `comment`.

neo773:
@JanCizmar 
Could I get a re-review? pushed some new commits and tested successfully with a real world file with 65 keys 

https://github.com/SokoloffA/radiola/blob/main/Radiola/Localizable.xcstrings

backend/data/src/main/kotlin/io/tolgee/formats/apple/in/xcstrings/XcstringsFileProcessor.kt
```diff
@@ -41,33 +44,77 @@ class XcstringsFileProcessor(
...
       }
     }
   }
 
+
+  private fun convertXCtoTolgeeTranslationState(xcState: String?): TranslationState {
+    return when (xcState) {
+      "needs_review" -> TranslationState.TRANSLATED
+      "translated" -> TranslationState.REVIEWED
+      null -> TranslationState.UNTRANSLATED
```
JanCizmar, line 84: I don't think you need the null branch, when it goes to the same state in the else branch.

backend/data/src/main/kotlin/io/tolgee/formats/apple/in/xcstrings/XcstringsFileProcessor.kt
```diff
@@ -41,33 +44,77 @@ class XcstringsFileProcessor(
...
+        convertPlaceholders = context.importSettings.convertPlaceholdersToIcu,
+        isProjectIcuEnabled = context.projectIcuPlaceholdersEnabled
+      )
+
+      context.addTranslation(
+        keyName = key,
+        languageName = sourceLanguage,
+        value = converted.message,
+        convertedBy = importFormat,
+        state = TranslationState.TRANSLATED,
```
JanCizmar, line 60: Here you're are not converting the state.

backend/data/src/main/kotlin/io/tolgee/formats/apple/in/xcstrings/XcstringsFileProcessor.kt
```diff
@@ -41,33 +44,77 @@ class XcstringsFileProcessor(
...
     localization: JsonNode,
   ) {
     val stringUnit = localization.get("stringUnit")
+    val xcState = stringUnit?.get("state")?.asText()
+    var translationState = convertXCtoTolgeeTranslationState(xcState)
+
     val translationValue = stringUnit?.get("value")?.asText()
 
     if (translationValue != null) {
+      val converted = messageConvertor.convert(
```
JanCizmar, line 101: It seems like the conversion and adding translation logic can be extracted to separate function for this case and also for the case when the sourceLanguage is missing.

backend/data/src/main/kotlin/io/tolgee/formats/apple/in/xcstrings/XcstringsFileProcessor.kt
```diff
@@ -78,9 +125,16 @@ class XcstringsFileProcessor(
...
     val forms = mutableMapOf<String, String>()
+    var translationState = TranslationState.UNTRANSLATED
 
     variations.fields().forEach { (form, content) ->
-      val value = content.get("stringUnit")?.get("value")?.asText()
+      val stringUnit = content.get("stringUnit")
+      val value = stringUnit?.get("value")?.asText()
+      val currentState = convertXCtoTolgeeTranslationState(stringUnit?.get("state")?.asText())
+
+      if (form == "other" || form == "zero" || translationState == TranslationState.UNTRANSLATED) {
```
JanCizmar, line 135: Can you please add comment why this is happening here? I have no idea...

JanCizmar (changes requested):
Good job!

I was thinking more about the state importing and I think, we should completely remove it. 

Tolgee is designed to be the source of truth, so I don't think one would ever want to import the state to Tolgee. 

If they have all reviewed in he xcstrings, they can always run batch operation and make everything reviewed.

We can keep it for export, but for import, I would just ignore it, because It doesn't look to me that we would need a checkbox to enable/disable the state import and that doesn't worth the hustle.

neo773:
@JanCizmar 
Removed translation state and added the changes.

JanCizmar:
It looks fine to me, if the tests pass, we can merge it!

JanCizmar:
Tell me if you are open to other opportunities to work on Tolgee! Here is the link to our community slack https://tolg.ee/slack where you can DM me. 🚀

PR #2896, opened 2025-02-06
fix: OpenApi docs for exporting endpoints

### LINKED ISSUE #2871: Minor API spec issue: StreamingResponseBinary is object, not binary
As a follow on for #1450 , the current REST API spec has `type: object` for `StreamingResponseBody`. It's more precise and more helpful to generated API clients if it is `type: binary` (in which case the generated client will expect a `File` or `Path`). I encountered this when exporting a project's translations to a zip file via API.

Image

JanCizmar:
Closes #2871


JanCizmar:
@jdimeo Hope this helps!

StaNov (approved):

PR #2908, opened 2025-02-12
feat: Notifications, visual fixes

StaNov:


webapp/src/component/layout/Notifications/NotificationItem.tsx
```diff
@@ -75,34 +87,72 @@ export const NotificationItem: React.FC<NotificationItemProps> = ({
...
+                name: originatingUser.name,
+                avatar: originatingUser.avatar,
+                type: 'USER',
+                id: originatingUser.id || 0,
+              }}
+              size={32}
+            />
+          )
+        ) : (
+          <LoadingSkeleton
```
JanCizmar, line 103: Can we move the notification skeleton to separate component so it doesn't mix with the actual code?

webapp/src/component/layout/Notifications/NotificationsPopup.tsx
```diff
@@ -167,11 +167,12 @@ export const NotificationsPopup: React.FC<NotificationsPopupProps> = ({
             <T keyName="notifications-empty" />
           </ListItem>
         )}
-        {(notificationsLoadable.isFetching ||
-          notificationsLoadable.hasNextPage) && (
-          <ListItem>
-            <BoxLoading width="100%" />
-          </ListItem>
+        {[1, 2, 3, 4, 5].map(
```
JanCizmar, line 170: This is not very readable. I guess it's there because of the skeletons, right? 
Move it to separate component named `NotificationSkeletons` or something.

Then move it to separate component and don't mix the skeleton with the real notification item. 

This check for fetching, should happen before the `.map` function. Rendering 5x nothing makes no sense.

JanCizmar (changes requested):

webapp/src/component/layout/Notifications/NotificationsPopup.tsx
StaNov: I am adding the skeletons to the end of the list in case there are some more notifications left to be loaded.

When a user scrolls to them, it loads more.

image

I updated the condition in the front of the map function, thanks for the catch.

JanCizmar (approved):

PR #2911, opened 2025-02-13
feat: Notification settings

StaNov:


webapp/src/hooks/useTimeDistance.ts
```diff
@@ -0,0 +1,30 @@
+import { useTranslate } from '@tolgee/react';
+
+export function useTimeDistance() {
+  const { t } = useTranslate();
+
+  return (timeInPast: number | string | Date) => {
+    const differenceInSeconds =
```
JanCizmar, line 7: If it's more then a month I would just print the date in short variant.

webapp/src/views/userSettings/notifications/SettingsRow.tsx
```diff
@@ -0,0 +1,101 @@
...
+        query: {
+          group: group,
+          channel: channel,
+          enabled: enabled,
+        },
+      },
+      {
+        onSuccess() {
+          messageService.success(
+            <T keyName="User data - Successfully updated!" />
```
JanCizmar, line 51: This is not very nice key name.

webapp/src/views/userSettings/notifications/SettingsRow.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <Typography
+            variant="body2"
+            color={(theme) => theme.palette.text.secondary}
+          >
+            {subdescription}
+          </Typography>
+        </Box>
+      </Box>
+      <Box>
+        <Tooltip title={disabledInApp && 'Cannot be turned off'}>
```
JanCizmar, line 78: not translated.

webapp/src/views/userSettings/notifications/SettingsRow.tsx
```diff
@@ -0,0 +1,101 @@
...
+            <Switch
+              checked={inAppEnabled}
+              disabled={disabledInApp}
+              onClick={() => saveSettings('IN_APP', !inAppEnabled)}
+            />
+          </StyledSwitch>
+        </Tooltip>
+      </Box>
+      <Box>
+        <Tooltip title={disabledEmail && 'Cannot be turned off'}>
```
JanCizmar, line 89: Not translated. Btw you should never input the untranslated raw key to the code. Use the Tolgee Dev tools immediately.

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationSettingService.kt
```diff
@@ -0,0 +1,62 @@
...
+    notification: Notification,
+    channel: NotificationChannel,
+  ) = notificationSettingRepository
+    .findByUserIdAndGroupAndChannel(
+      notification.user.id,
+      notification.type.group,
+      channel,
+    )?.enabled ?: true
+
+  private fun defaultSettings(
```
JanCizmar, line 35: the methods should be named with some verb. "getDefaultSettings"

backend/data/src/main/kotlin/io/tolgee/repository/notification/NotificationSettingRepository.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.context.annotation.Lazy
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+
+@Repository
+@Lazy
+interface NotificationSettingRepository : JpaRepository<NotificationSetting, Long> {
+  fun findByUserId(userId: Long): List<NotificationSetting>
+
+  fun findByUserIdAndGroupAndChannel(
```
JanCizmar, line 15: Head up. The user can be deleted or disabled. Write custom query everywhere you're working with user and filter our disabled and deteled users.

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationSettingService.kt
```diff
@@ -0,0 +1,62 @@
...
+    user: UserAccount,
+    group: NotificationTypeGroup,
+    channel: NotificationChannel,
+  ) = NotificationSetting().apply {
```
JanCizmar, line 39: We want to make the default for each channel configurable via server config. Please add such option and test it.

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationService.kt
```diff
@@ -33,15 +35,19 @@ class NotificationService(
...
-    notificationRepository.save(notification)
-    applicationEventPublisher.publishEvent(
-      OnNotificationsChangedForUser(
-        notification.user.id,
-        notification,
-      ),
-    )
+  fun notify(notification: Notification) {
+    if (notificationSettingService.getSettingValue(notification, NotificationChannel.EMAIL)) {
+      emailNotificationsService.sendEmailNotification(notification)
```
JanCizmar, line 40: Prevent sending e-mail notifications to deleted users and disabled users (please test it)

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationService.kt
```diff
@@ -33,15 +35,19 @@ class NotificationService(
...
-  fun save(notification: Notification) {
-    emailNotificationsService.sendEmailNotification(notification)
-    notificationRepository.save(notification)
-    applicationEventPublisher.publishEvent(
-      OnNotificationsChangedForUser(
-        notification.user.id,
-        notification,
-      ),
-    )
+  fun notify(notification: Notification) {
```
JanCizmar, line 38: Please check that notifications are not created for deleted users. (and test it)

You can create notifications for disabled users, but don't send them e-mails.

JanCizmar (changes requested):

webapp/src/hooks/useTimeDistance.ts
StaNov: I put there medium which is `23. 1. 2025` but short is `23.01.25` which is imho less readable and we have space for the medium one.

webapp/src/views/userSettings/notifications/SettingsRow.tsx
StaNov: True, thanks for the catch.

Btw fun fact: I copy-pasted it from [an existing place](https://github.com/tolgee/tolgee-platform/blob/main/webapp/src/views/userSettings/userProfile/UserProfileView.tsx#L43).

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationService.kt
StaNov: I would do this in a separate PR, I made a note about it.

backend/data/src/main/kotlin/io/tolgee/service/notification/NotificationService.kt
StaNov: Will be done in another PR.

backend/data/src/main/kotlin/io/tolgee/repository/notification/NotificationSettingRepository.kt
StaNov: Will be done in another PR.

JanCizmar (approved):

PR #2917, opened 2025-02-17
Slack refactoring

fprochazka:


ee/backend/app/src/main/kotlin/io/tolgee/ee/component/slackIntegration/notification/SlackAutomationMessageSender.kt
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    val messagesDto = slackTranslationChangeMessageFactory.createTranslationChangeMessages(context)
+    val savedMessagesByKey = savedSlackMessageService.findAll(messagesDto, context.slackConfig.id).groupBy { it.keyId }
+
+    val messagesToUpdate = mutableMapOf<Long, MutableList<SlackMessageDto>>()
+    val messagesToCreate = mutableListOf<SlackMessageDto>()
+
+    for (message in messagesDto) {
+      if (!isExplicitlySubscribedToAnyUpdatedLanguage(message, context.slackConfig)) {
```
JanCizmar, line 64: I guess some tests don't pass because of this part.

The logic should be like this:
 - When I am subscribed to some languages explicitely, I always get new message on change
 - When I am subscribed with non-specific language subscription, I get messages updated only when something change
   - That's because specific language translators/reviewers want to see all the changes. Product managers and others just need to check it was changed but they don't need to be always notified
 - When there is more then treshold messages I am notified about count only and I get a button to see it in Tolgee

JanCizmar (changes requested):
Good job! It looks like you can indeed write some good clean code. 🚀

PR #2927, opened 2025-02-20
fix: Adjust height threshold for Firefox compatibility

### LINKED ISSUE #2924: Incorrect overflow in Firefox/Ubuntu
Image

**Describe the bug**
Seems like incorrect measurement of the element height, causing overflow shadow being visible when it's not supposed to.

**To Reproduce**
Seems like only happens in Firefox (135.0.1) on Ubuntu (24.04).

1. Open tolgee platform
2. Open Demo project (or any other project containing translations)
3. Go to "Translations" tab

**Expected behavior**
Translations (and keys) that fit into one line, should not contain overflow, only translations longer than 2 lines should have the fade-out effect, so it's clear that we don't display the full translation
Image

**Additional context**
Please make sure to test the functionality in all major browsers (chrome, safari, firefox) after the fix.

neo773:
https://github.com/user-attachments/assets/b829ef5e-a6ac-4bb4-ae8e-85fbd9bc3a37

/claim #2924
/closes #2924

### DIFF
--- a/webapp/src/component/LimitedHeightText.tsx
+++ b/webapp/src/component/LimitedHeightText.tsx
@@ -86,9 +86,10 @@ export const LimitedHeightText: React.FC<Props> = ({
   const detectExpandability = () => {
     const textElement = textRef.current;
     if (textElement != null) {
-      // values should be the same, however firefox applies some weird rounding
-      // so by adding one we eliminate that
-      setExpandable(textElement.clientHeight + 1 < textElement.scrollHeight);
+      // Use a 2px threshold to account for browser rounding differences
+      const heightDifference =
+        textElement.scrollHeight - textElement.clientHeight;
+      setExpandable(heightDifference > 2);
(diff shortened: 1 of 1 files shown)


JanCizmar:
This looks to me like it will break the gradient ultimately on the KeyCell. I cannot see, how this would fix the issue specifically for the Ubuntu Firefox.

neo773:
I don't think we're using the gradient

This is what Chrome shows

image


Firefox on Ubuntu

image

neo773:
@JanCizmar 
Pushed a new commit


https://github.com/user-attachments/assets/581ea80a-5b47-40b9-85d9-b7a7d878ee2d

stepan662:
Ok, cool. Seems like it fixed the issue.

JanCizmar (approved):

stepan662 (approved):

PR #2929, opened 2025-02-20
feat: third party authentication provider change

### LINKED ISSUE #2913: Changing third-party authentication provider
This is a follow-up issue for #2619. We need to allow users to switch between all supported auth providers, not just SSO:

- [ ] Refine the UX design of the flow
- [ ] Allow switching between all supported auth providers
- [ ] Implement tests

Anty0:
Fixes #2913


webapp/src/component/layout/DashboardPage.tsx
```diff
@@ -49,7 +49,9 @@ export const DashboardPage: FunctionComponent<Props> = ({
     (c) => Boolean(c.auth.jwtToken) && Boolean(c.auth.adminToken)
   );
 
-  const rightPanelWidth = useGlobalContext((c) => c.layout.rightPanelWidth);
+  const rightPanelWidth = !hideQuickStart
+    ? useGlobalContext((c) => c.layout.rightPanelWidth)
+    : 0;
```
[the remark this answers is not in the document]
Anty0: Nice catch! Overlooked it! Introducing bugs when trying to fix bugs... Classic. 😁

backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
```diff
@@ -34,11 +34,19 @@ data class UserAccount(
...
   override var id: Long = 0L,
   @field:NotBlank
-  var username: String = "",
+  override var username: String = "",
   var password: String? = null,
   var name: String = "",
   @Enumerated(EnumType.STRING)
   var role: Role? = Role.USER,
+  /**
+   * This property is redundant; kept for legacy reasons.
```
JanCizmar, line 43: What are the legacy reasons specifically? -> Add it to the comment.

backend/data/src/main/kotlin/io/tolgee/service/security/AuthProviderChangeService.kt
```diff
@@ -1,117 +1,269 @@
...
+    throw AuthenticationException(Message.THIRD_PARTY_SWITCH_INITIATED)
   }
 
   @Transactional(propagation = Propagation.REQUIRES_NEW)
-  fun saveProviderChange(data: AuthProviderChangeData) {
-    val req =
-      authProviderChangeRequestRepository.findByUserAccountId(data.userAccount.id)
-        ?: AuthProviderChangeRequest()
-    authProviderChangeRequestRepository.save(req from data)
+  protected fun forceSave(
```
JanCizmar, line 79: Name it `saveInNewTransaction` instead and explain why you're doing so in the comment.

backend/data/src/main/kotlin/io/tolgee/service/security/SignUpService.kt
```diff
@@ -28,86 +26,104 @@ class SignUpService(
...
-      throw AuthenticationException(Message.REGISTRATIONS_NOT_ALLOWED)
+    return SignUpProcessor(entity, invitationCode, organizationName, userSource).process()
+  }
+
+  fun dtoToEntity(request: SignUpDto): UserAccount {
+    val encodedPassword = passwordEncoder.encode(request.password!!)
+    return UserAccount(name = request.name, username = request.email, password = encodedPassword)
+  }
+
+  inner class SignUpProcessor(
```
JanCizmar, line 72: -> To separate file

backend/development/src/main/kotlin/io/tolgee/controllers/internal/e2eData/AuthProviderChangeE2eDataController.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.transaction.annotation.Transactional
+import org.springframework.web.bind.annotation.CrossOrigin
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@CrossOrigin(origins = ["*"])
+@Hidden
+@RequestMapping(value = ["internal/e2e-data/authProviderChange"])
```
JanCizmar, line 17:
```suggestion
@RequestMapping(value = ["internal/e2e-data/auth-provider-change"])
```

backend/security/src/main/kotlin/io/tolgee/security/authentication/EmailValidInterceptor.kt
```diff
@@ -0,0 +1,60 @@
...
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.context.annotation.Lazy
+import org.springframework.core.Ordered
+import org.springframework.stereotype.Component
+import org.springframework.web.method.HandlerMethod
+import org.springframework.web.servlet.HandlerInterceptor
+
+@Component
+class EmailValidInterceptor(
```
JanCizmar, line 17:
```suggestion
class EmailValidationInterceptor(
```

backend/security/src/test/kotlin/io/tolgee/security/authentication/SsoAuthenticationInterceptorTest.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.mockito.Mockito
+import org.mockito.kotlin.any
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get
+import org.springframework.test.web.servlet.setup.MockMvcBuilders
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RestController
+
+class SsoAuthenticationInterceptorTest {
```
JanCizmar, line 21: These are unit tests, move them to unit folder.

JanCizmar:
It's a great job BTW! 🚀 Hope it will work without issues.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/PatController.kt
```diff
@@ -50,6 +52,8 @@ class PatController(
 ) : IController {
   @GetMapping(value = [""])
   @Operation(summary = "Get PAKs")
+  @BypassEmailVerification
+  @BypassForcedSsoAuthentication
```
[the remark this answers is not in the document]
Anty0: I would like the user to be able to manage existing tokens at any time—so he can delete a token that got exposed in a leak, for example. Tokens shouldn't work anyway while the account isn't verified or SSO is forced, if I didn't miss anything, but this way, the user can remove the token before restoring access to his account.
It made sense to me to allow a user in this state to change security-related settings if, for some reason, the user can't finish verification/migration but still needs to change the password/active tokens.

backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
Anty0: It is mainly because, as outlined in the comment, its value is determined by other existing properties.
I'll modify the comment to specify the reason explicitly. :)

Technically, we don't need to make it a legacy; we can just keep the rules for the value, but I think at this point, it is redundant.

backend/data/src/main/kotlin/io/tolgee/service/security/AuthProviderChangeService.kt
```diff
@@ -1,117 +1,269 @@
...
   fun getRequestedChange(userAccount: UserAccount): AuthProviderDto? {
     return getActiveAuthProviderChangeRequest(userAccount)?.asAuthProviderDto()
   }
 
-  fun initiateProviderChange(data: AuthProviderChangeData) {
-    if (data.accountType == UserAccount.AccountType.MANAGED) {
-      self.saveProviderChange(data)
-      throw AuthenticationException(Message.THIRD_PARTY_SWITCH_INITIATED)
+  fun tryInitiate(
+    resolvedUser: UserAccount?,
```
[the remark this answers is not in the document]
Anty0: There are two users at play here—the user who is already authenticated, `currentUser`, and the user who would be authenticated by the third-party authentication process, `resolvedUser`. If they both match (and a few other things check out), we can initiate the third-party authentication provider change.
Do you have any ideas for a better name? Should I add a comment with a simplified version of this explanation?

backend/data/src/main/kotlin/io/tolgee/service/security/SignUpService.kt
Anty0: I don't think it makes sense in this context. The `SignUpService` doesn't contain any logic itself—it only creates and calls the `SignUpProcessor`. The `SignUpProcessor` also needs access to all the services and components. Is there some spring way of handling this? I wanted to make it a separate file, but wouldn't using a non-inner class make the `SignUpService` just a one big constructor call passing over all the services to the `SignUpProcessor`?

backend/security/src/test/kotlin/io/tolgee/security/authentication/SsoAuthenticationInterceptorTest.kt
Anty0: I'll move all three (inc. `EmailValidationInterceptor` and `AuthenticationInterceptorTest`) tests there. `SsoAuthenticationInterceptorTest` and `EmailValidationInterceptor` are based on the `AuthenticationInterceptorTest`.

Anty0:
Thank you! I'll finish up with fix-ups tomorrow.
Hope it'll work flawlessly!

- I'm bypassing the verification regarding account security settings and required endpoints, but we can change that. Before this PR, the email verification would block project and organization endpoints, so the current setup is stricter, even though it might not look like it from the patchset.
- I didn't test with old data. Do you have any concerns about what might fail when deployed on a legacy database? There might be some quirks I haven't encountered yet.
- I'll create separate tests for a few endpoints and see if they get blocked when the user is in the email verification/SSO migration state. That should cover a situation where the interceptor stops working globally for whatever reason while still working in the unit test.

JanCizmar:
> I'm bypassing the verification regarding account security settings and required endpoints, but we can change that. Before this PR, the email verification would block project and organization endpoints, so the current setup is stricter, even though it might not look like it from the patchset.

OK, but I still wonder, why do we need to Bypass it for so many endpoint which look unrelated to this case. Why I may edit my PAT tokens when I don't have verified email?

backend/api/src/main/kotlin/io/tolgee/security/thirdParty/ThirdPartyUserHandler.kt
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    if (userAccount != null) {
+      if (data.thirdPartyAuthType in arrayOf(ThirdPartyAuthType.SSO, ThirdPartyAuthType.SSO_GLOBAL)) {
+        userAccountService.updateSsoSession(userAccount, data.refreshToken)
+      }
+      return userAccount
+    }
+
+    return createUser(data)
```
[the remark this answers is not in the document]
Anty0: Yes, because it calls `signUpService.signUp`, and that function checks whether signing up is allowed. However, it allows signing up with an SSO account regardless.

JanCizmar (approved):

PR #2933, opened 2025-02-22
Fix Japanese static data load

Rapilias:
## Summary
Thank you for releasing the Japanese translation.

In v3.103.0, Japanese is selectable, but it is not displayed. 
From my understanding, I think that the static data for i18n included in this pull request needs to be loaded.

ref: https://github.com/tolgee/tolgee-platform/pull/2932
### v3.103.0
スクリーンショット 2025-02-22 135048
### tolgee:main
スクリーンショット 2025-02-22 140327
### Rapilias:feature/fix-japanese-static-data-load
スクリーンショット 2025-02-22 135759

## Checked
- [x] Even if I select Japanese in v3.103.0, Japanese is not displayed
- [x] If I make changes to the PR, Japanese is displayed

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -55,6 +55,7 @@ const tolgee = Tolgee()
       de: () => import('./i18n/de.json').then((m) => m.default),
       pt: () => import('./i18n/pt.json').then((m) => m.default),
       da: () => import('./i18n/da.json').then((m) => m.default),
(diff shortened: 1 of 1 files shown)


JanCizmar:
Thanks a lot! We should add some tests, testing that all languages are working!

JanCizmar (approved):

Rapilias:
I checked that it is displayed correctly in 3.103.1
Thank you!

PR #2936, opened 2025-02-23
Escaping Tags in XLIFF

### LINKED ISSUE #2912: Escaping Tags in XLIFF export/import
**Describtion**
Hi there, 
we use Tolgee in our company in a project. 

We use HTML tags to format our text (e.g. line breaks, a list of points, paragraphs).
Right now we are facing the problem that HTML Tags are not escaped in the XLIFF files. They are written in plain text

Due to the fact that XLIFF is also an XML based format, we are not able to consume the XLIFF file with other software properly, because other XLIFF parser implementations expect the HTML Tags to be escaped.

**Expected behavior**
I would expect that html tags are escaped with ```&lt;TAGNAME&gt;``` for opening tags and ```&lt;/TAGNAME&gt;&#13;``` for closing tags respectively.

For instance ```<p> some content </p>``` would be escaped with ```&lt;/P&gt; some content &lt;/P&gt;&#13;```, for import and for export as well.

Thank you for any reply.

neo773:
This PR adds support for escaping HTML tags in XLIFF

Check list
- [x] Import
- [x] Export
- [x] Tests
- [x] Docs [PR](https://github.com/tolgee/documentation/pull/857)
- [x] CDN (Complete but unable to test it due to local dev environment issues)

/claim #2912
/closes #2912


backend/data/src/main/kotlin/io/tolgee/formats/xliff/util/XliffHtmlEscaper.kt
```diff
@@ -0,0 +1,17 @@
+package io.tolgee.formats.xliff.util
+
+object XliffHtmlEscaper {
+    fun escapeHtmlTags(text: String): String {
+        var result = text.replace("\"", "&quot;")
+        
+        val regex: Regex = "<(/?[^>]+)>".toRegex()
+        return result.replace(regex) { matchResult ->
+            val tag = matchResult.groupValues[1]
+            when {
```
cardvvs, line 10: this `when` clause is unnecessary since all the branches return same value

backend/data/src/main/kotlin/io/tolgee/formats/xliff/util/XliffHtmlEscaper.kt
```diff
@@ -0,0 +1,17 @@
+package io.tolgee.formats.xliff.util
+
+object XliffHtmlEscaper {
+    fun escapeHtmlTags(text: String): String {
+        var result = text.replace("\"", "&quot;")
+        
+        val regex: Regex = "<(/?[^>]+)>".toRegex()
```
cardvvs, line 7: imho no need to compile regex every function call since it's always same, I'd extract it to private variable. Also maybe question of tag validity is on the table??? because this regex will also match for example `<<<p> `which is not a valid tag according to specification (correct me if I'm wrong) @JanCizmar what is your take on this?

neo773: Validity of the markup is guaranteed by the Parser, Escaping takes place after Parser has successfully parsed it

If you try it with invalid markup it will throw an error

image

backend/data/src/main/kotlin/io/tolgee/formats/xliff/in/Xliff12FileProcessor.kt
```diff
@@ -50,7 +51,13 @@ class Xliff12FileProcessor(
     transUnitId: String,
     language: String,
   ) {
-    val converted = convertMessage(text, language)
+    val textToConvert =
```
JanCizmar, line 54: I don't think it makes any sense to escape html when importing. This wasn't in the scope of the task.

backend/data/src/main/kotlin/io/tolgee/dtos/request/export/ExportParams.kt
```diff
@@ -82,6 +83,10 @@ data class ExportParams(
     description = ExportParamsDocs.FILE_STRUCTURE_TEMPLATE_DESCRIPTION,
   )
   override var fileStructureTemplate: String? = null,
+  @field:Parameter(
+    description = ESCAPE_HTML_DESCRIPTION,
+  )
+  override var escapeHtml: Boolean? = false,
```
JanCizmar, line 89: Please add documentation comment.

```
/** Enabling or disabling HTML escaping for XLIFF format. Some tools expect the XML/HTML tags escaped and don't accept raw unescaped tags. */

File: backend/data/src/main/kotlin/io/tolgee/model/contentDelivery/ContentDeliveryConfig.kt [hunk 5]
```diff
@@ -129,4 +129,7 @@ class ContentDeliveryConfig(
 
   @ActivityLoggedProp
   override var fileStructureTemplate: String? = null
+
+  @ActivityLoggedProp
+  override var escapeHtml: Boolean? = false
```
JanCizmar, line 134: There is no reason to make it nullable. Make it false by default and add `@ColumnDefault` annotation so tests won't fail.

JanCizmar (changes requested):

backend/data/src/main/kotlin/io/tolgee/formats/xliff/util/XliffHtmlEscaper.kt
```diff
@@ -0,0 +1,13 @@
+package io.tolgee.formats.xliff.util
+
+object XliffHtmlEscaper {
```
JanCizmar, line 3: I believe we already have apache commons package dependency, which we can use for XML escaping:
```
import org.apache.commons.text.StringEscapeUtils

fun escapeXmlContent(input: String): String {
    return StringEscapeUtils.escapeXml10(input)
}

fun main() {
    val rawText = """<tag attribute="value">Content & more</tag>"""
    val escapedText = escapeXmlContent(rawText)
    println(escapedText) // Output: &lt;tag attribute=&quot;value&quot;&gt;Content &amp; more&lt;/tag&gt;
}
```

I believe they have it tested well.

JanCizmar (changes requested):

backend/data/src/main/kotlin/io/tolgee/formats/xliff/in/Xliff12FileProcessor.kt
neo773: Will remove it, I only added it since the author of the issue mentioned it

image

backend/data/src/main/kotlin/io/tolgee/formats/xliff/util/XliffHtmlEscaper.kt
JanCizmar: It would make sense to replace all the possible xml syntax, but I am proposing the Apache commons solution in my review.

@cardvvs Thanks for the review anyway.

neo773:
Apologies for the spam, the branch ran into several issues and I had to reset it and do a merge from a clean branch.
@JanCizmar  It has addressed the changes you requested.

JanCizmar:
Hey! So does it currently also work with the CDN and is it tested?

neo773:
@JanCizmar 
I’ve been trying to enable CDN to test it but haven’t had any luck so far.
Will look into it again today.

neo773:
@JanCizmar 
Got CDN working with Cloudflare and it seems to be working as expected

https://github.com/user-attachments/assets/d0d801ab-19d0-4203-8dd9-21831aef8f39

JanCizmar:
Hello! Thanks a lot it's great. I would like to see a cypress test, which tests that the CD configuration is actually saved with the the new checkbox value.

That's my last input before I jump on the final Code Review. 

Thanks a lot for the persistence on this.

backend/data/src/main/kotlin/io/tolgee/dtos/ExportParamsDocs.kt
```diff
@@ -66,4 +66,8 @@ The `{snakeLanguageTag}` placeholder is the same as `{languageTag}` but in snake
 The Android specific `{androidLanguageTag}` placeholder is the same as `{languageTag}` 
 but in Android format. (e.g., en-rUS)
 """
+
+  const val HTML_ESCAPE_DESCRIPTION = """If true, HTML tags are escaped in the exported file.
```
JanCizmar, line 70:
```suggestion
  const val HTML_ESCAPE_DESCRIPTION = """If true, HTML tags are escaped in the exported file. (Supported in the XLIFF format only).
```

backend/data/src/main/kotlin/io/tolgee/dtos/request/ContentDeliveryConfigRequest.kt
```diff
@@ -71,4 +71,9 @@ class ContentDeliveryConfigRequest() : IExportParams {
   override var messageFormat: ExportMessageFormat? = null
 
   override var fileStructureTemplate: String? = null
+
+  @Schema(
+    description = "If true, HTML tags are escaped in the exported file.",
```
JanCizmar, line 76: You might want the description constant, don't you?

e2e/cypress/e2e/projects/contentDelivery.cy.ts
```diff
@@ -146,6 +146,34 @@ describe('Content delivery', () => {
     );
   });
 
+  it('tests XLIFF format with HTML escaping', () => {
```
JanCizmar, line 149:
```suggestion
  it('stores content delivery configuration for XLIFF format with HTML escaping', () => {
```

webapp/src/views/projects/developer/contentDelivery/CdEscapeHtml.tsx
```diff
@@ -0,0 +1,66 @@
...
+const StyledHelpIcon = styled(HelpCircle)`
+  width: 17px;
+  height: 17px;
+`;
+
+type Props = {
+  className?: string;
+};
+
+export const CdEscapeHtml: FC<Props> = ({ className }) => {
```
JanCizmar, line 28: Nice that you added it as separate component.

webapp/src/views/projects/export/ExportFormContent.tsx
```diff
@@ -49,6 +51,11 @@ export const ExportFormContent = ({
           </StyledOptions>
         </>
       )}
+      {getFormatById(values.format).showEscapeHtml && (
+        <StyledOptions className="options">
+          <EscapeHtmlSelector />
```
JanCizmar, line 56: You should use the same field for both CD na export. As for the other fields.

JanCizmar (changes requested):

JanCizmar (approved):

JanCizmar:
You will have to fix the rest of the issues.

PR #2937, opened 2025-02-24
feat: Fetching more notifications by cursor

StaNov:


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/notification/NotificationController.kt
```diff
@@ -54,16 +57,35 @@ class NotificationController(
     pageable: Pageable,
     @ParameterObject
     filters: NotificationFilters = NotificationFilters(),
-  ): PagedModel<NotificationModel> {
+    cursor: String? = null,
+  ): PagedModelWithNextCursor<NotificationModel> {
+    if (cursor != null &&
+      pageable.pageNumber > 0 &&
+      pageable.sort.toString() != "id: DESC"
```
JanCizmar, line 64: I would not rely on the toString method, rather I check that the sort is default sort.

Also! I recommend to use `createdAt: desc, id: desc` for sorting. Because in some cases the date can be higher when id can be lower, because the id is allocated before the actual insert so there can be a race condition.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/notification/NotificationController.kt
```diff
@@ -54,16 +57,35 @@ class NotificationController(
...
+      )
+    val model =
+      pagedResourcesAssembler.toModel(
+        notifications,
+        NotificationModelAssembler(enhancers, notifications),
       )
-    return pagedResourcesAssembler.toModel(
-      notifications,
-      NotificationModelAssembler(enhancers, notifications),
+    return PagedModelWithNextCursor(
```
JanCizmar, line 81: This is hard to read. Add the parameter names so it's clear what parameter is assigned where. 

Instead of using ID as cursor, follow the same structure as it is in the cursor util.

io.tolgee.service.queryBuilders.CursorUtil.Companion#getCursor

JanCizmar: You might also generalize the cursor util. I believe the `io.tolgee.service.queryBuilders.CursorUtil.Companion#getCursor` is almost there, just the `item` parameter should be generic and the `io.tolgee.service.queryBuilders.CursorUtil.Companion#getCursorValue` should be moved to different class specific for translations.

JanCizmar (changes requested):

JanCizmar (approved):

PR #2941, opened 2025-02-24
feat: publish to s3 subfolder

cardvvs:
## Summary by CodeRabbit

- **New Features**
  - Added support for specifying an optional path prefix for S3 content storage, allowing files to be stored within a subfolder in the S3 bucket.
  - Updated web interface to include a new field for entering the S3 path prefix in storage configuration forms.

- **Improvements**
  - Enhanced validation for S3 bucket names and added validation for the new path field.
  - Expanded API and configuration models to support the new S3 path property.

- **Bug Fixes**
  - Ensured correct storage and retrieval of files when a path prefix is configured.

- **Tests**
  - Added and updated tests to verify correct handling of the S3 path prefix in storage operations.

- **Documentation**
  - Updated API schema descriptions to reflect the new S3 path option.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/ContentDeliveryConfigTestData.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/ContentDeliveryConfigTestData.kt
@@ -29,20 +29,6 @@ class ContentDeliveryConfigTestData : BaseTestData() {
         }
     }
 
-  val s3ContentStorageWithCustomPath =
-    projectBuilder.addContentStorage {
-      name = "S3"
-      this.s3ContentStorageConfig =
-        S3ContentStorageConfig(this).apply {
-          bucketName = "fake"
-          accessKey = "fake"
-          secretKey = "fake"
-          endpoint = "fake"
-          signingRegion = "fake"
-          path = "fake"
-        }
-    }
-
   val defaultServerContentDeliveryConfig =
     projectBuilder.addContentDeliveryConfig {
       name = "Default server"
@@ -63,14 +49,6 @@ class ContentDeliveryConfigTestData : BaseTestData() {
   val s3ContentDeliveryConfigWithCustomSlug =
     projectBuilder.addContentDeliveryConfig {
       contentStorage = s3ContentStorage.self
-      name = "Custom Slug"
-      slug = "my-slug"
-      customSlug = true
-    }
-
-  val s3ContentDeliveryConfigWithCustomSlugAndPath =
-    projectBuilder.addContentDeliveryConfig {
-      contentStorage = s3ContentStorageWithCustomPath.self
       name = "Custom Slug"
       slug = "my-slug"
       customSlug = true


webapp/src/ee/developer/storage/StorageFormS3.tsx
```diff
@@ -1,140 +1,150 @@
-import { DialogContent } from '@mui/material';
-import { useTranslate } from '@tolgee/react';
-import { Form, Formik, setNestedObjectValues } from 'formik';
+import {DialogContent} from '@mui/material';
```
StaNov, line 1: You can use the Prettier plugin to format the file in the way we format it.

backend/data/src/main/kotlin/io/tolgee/model/contentDelivery/S3Config.kt
```diff
@@ -6,6 +6,10 @@ interface S3Config : StorageConfig {
   val secretKey: String?
   val endpoint: String?
   val signingRegion: String?
+  /**
+   * Specifies optional subfolder structure within s3 bucket to which content will be stored
```
StaNov, line 10: I would remove it from here and keep it only at the API level. This one may get obsolete because it is not that much seen.

backend/data/src/main/kotlin/io/tolgee/dtos/contentDelivery/S3ContentStorageConfigDto.kt
```diff
@@ -6,7 +6,7 @@ import jakarta.validation.constraints.Size
 
 class S3ContentStorageConfigDto : S3Config {
   @field:NotBlank
-  @field:Size(max = 255)
+  @field:Size(min = 3, max = 63)
```
StaNov, line 9: minor: I would put this change into a separate commit with description why it was done.

Or maybe even a separate PR if appropriate.

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/S3Settings.kt
```diff
@@ -40,4 +40,6 @@ class S3Settings(
   override var signingRegion: String? = null,
   @DocProperty(description = "Name of the bucket where Tolgee will store its files.")
   override var bucketName: String? = null,
+  @DocProperty(description = "Optional subfolder structure within s3 bucket to which content will be stored")
```
StaNov, line 43: I would add an example here, eg. to show if heading slash or trailing slash is ok.

backend/app/src/main/kotlin/io/tolgee/configuration/FileStorageConfiguration.kt
```diff
@@ -25,9 +25,7 @@ class FileStorageConfiguration(
       return InMemoryFileStorage()
     }
     if (s3config.enabled) {
-      val amazonS3 = S3ClientProvider(s3config).provide()
-      val bucketName = properties.fileStorage.s3.bucketName ?: throw RuntimeException("Bucket name is not set")
-      return S3FileStorage(bucketName, amazonS3)
+      return s3FileStorageFactory.create(s3config)
```
StaNov, line 28: Good catch that we weren't using the factory here!

.run/Backend localhost.run.xml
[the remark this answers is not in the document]
cardvvs: yes, that was by mistake as I was creating PR in a rush, will revert this ofc

backend/data/src/main/kotlin/io/tolgee/model/contentDelivery/S3Config.kt
cardvvs: you mean field or just a javadoc comment?

StaNov: The javadoc.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/contentDelivery/ContentDeliveryConfigControllerTest.kt
```diff
@@ -89,6 +91,16 @@ class ContentDeliveryConfigControllerTest : ProjectAuthControllerTest("/v2/proje
     assertPruned(mocked)
   }
 
+  @Test
+  @ProjectJWTAuthTestMethod
+  fun `publishes to custom server content delivery config`() {
+    tolgeeProperties.contentDelivery.storage.s3.bucketName = "my-bucket"
```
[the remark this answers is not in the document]
JanCizmar: Yeah. this should be changed back in the aftereach.

backend/data/src/main/kotlin/io/tolgee/component/fileStorage/S3FileStorageFactory.kt
```diff
@@ -8,6 +8,20 @@ class S3FileStorageFactory {
   fun create(config: S3Config): S3FileStorage {
     val client = S3ClientProvider(config).provide()
     val bucketName = config.bucketName ?: throw RuntimeException("Bucket name for S3 storage is not set")
-    return S3FileStorage(bucketName, client)
+    val path = getCanonicalPath(config.path)
+
+    return S3FileStorage(bucketName = bucketName, path = path, s3 = client)
+  }
+
+  private fun getCanonicalPath(path: String?): String {
```
JanCizmar, line 16: It's good you handled this, but...

- The code is very long and ugly (actually there already is a method for this)
   ```io.tolgee.fixtures.RemoveSlashSuffixKt#removeSlashSuffix```
- It does not handle the cases when the path is invalid
- It does not handle duplicated `/` at the end
- I am not sure whether it should be placed here. I would be afraid that someone will create the S3FileStorage will be created without this factory and so it would not handle the path correctly.

JanCizmar (changes requested):

PR #2954, opened 2025-02-27
fix: add sso user to organization explicitly - fixes accepting project invitations during signup

Anty0:


backend/data/src/main/kotlin/io/tolgee/service/security/SignUpService.kt
```diff
@@ -82,11 +82,15 @@ class SignUpService(
         }
       }
       invitationService.accept(invitation.code, user)
-    } else if (organizationForced != null) {
-      organizationRoleService.grantRoleToUser(
+    }
```
JanCizmar, line 85: this function is not readable anymore and will need refactoring. I will approve it now, but for the next iteration, we will need refactor it after this.

- If I didn't have the context I wouldn't know whats the organizationForced
- This whole method needs to me moved to special class with single method where all the params are constructor  params
- The ifs are too nested and should be split to different methods with clean semantic names and comments

JanCizmar (approved):

backend/data/src/main/kotlin/io/tolgee/service/security/SignUpService.kt
Anty0: I agree; it got a bit out of hand with all the SSO-related invitation checks. I have added this to my notes.

PR #2955, opened 2025-02-27
feat: Cypress tests

StaNov:


backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/NotificationsTestData.kt
```diff
@@ -3,13 +3,19 @@ package io.tolgee.development.testDataBuilder.data
 import io.tolgee.model.notifications.Notification
 
 class NotificationsTestData : BaseTestData() {
+  private var currentTaskNumber = 1L
+
+  init {
+    user.username += "@example.com"
```
JanCizmar, line 9: Afaik BaseTestData do have the username as constructor param.

JanCizmar (approved):

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/NotificationsTestData.kt
StaNov: Ah, they do! Thanks, I'll change it.

Issue #2778, opened 2024-12-13
Task improvements


ZuzanaOdstrcilova:
We need to do small improvements after user testing.

https://www.figma.com/design/wVmz098ZrSxswOI0nbIjQ2/Design-system-Tolgee?node-id=3310-62794&t=WDuUpTtIb4zSav5r-1

Change of behaviour (problem found with Market)
- [x] If I am assigned to the task and am editing translation in tran.view, I will se "Save and done"
   - this requires more research
- [x] The same for review, but in case that the task is blocked, there will be just "review" and not marked as"done"
   - this also requires more research + translate task can also be blocked, not just review task
- [x] Add filter - "not done"

Task board: 

- [x] - Buttons: Show all, task report, cancel task
    - show all - ok
    - task report - it's hard to translate to czech ("report/zpráva o úkolu" - sounds weird) - for me "download report" ( "stáhnout report") is better
    - cancel task - problematic (as described below)
- [x] - Change tooltip to show all / readable + cancel
   - readable + cancel ?
- [x] - add tooltip - menu (task actions)
- [x] - Change Icons (I), report to download icon
- [x] - Change dialog cancel task
  - move the actions from the top right corner to the bottom and put an "x" icon to close the dialog instead
  - think of adding icons to the change state actions
- [x] - Change #keys to #words
- [x] - Change color of review badge

Translations in task

- [x] - Add dialog after opening blocked task
   - add task info above editor, telling user if the task is blocked or there is some problem (unassigned, blocked etc.)
- [x] - Change icons (as in task board)
- [x] - Add now floating component - small info about task
   - task component could be floating instead of the search bar

Task detail

- [x] - Change actions Cancel task
- [x] - Change general button cancel to close

  - task states will be renamed to "finished" and "cancelled", dialog will be closed by "x" button

PR #2956, opened 2025-02-27
feat: task improvements

stepan662:
Closing #2778


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/TaskService.kt
```diff
@@ -559,8 +559,8 @@ class TaskService(
   private fun createNotificationIfApplicable(task: Task) {
     val notificationType =
       when (task.state) {
-        TaskState.DONE -> NotificationType.TASK_COMPLETED
-        TaskState.CLOSED -> NotificationType.TASK_CLOSED
+        TaskState.FINISHED -> NotificationType.TASK_COMPLETED
```
JanCizmar, line 562: Let's migrate also the notification types. 

Also if you are in the migrating this, the `TASKS_CREATE` should be renamed to `TASK_CREATE`.

JanCizmar (changes requested):

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/TaskService.kt
stepan662: There are to activity events, `TASK_CREATE` and `TASKS_CREATE`, because you can create multiple tasks at once, I think it's correct that it's separated, or not?

JanCizmar: OK then.

JanCizmar (approved):

PR #2988, opened 2025-03-13
Feature/add cloudfront cache purger

kingcdavid:


backend/data/src/test/kotlin/io/tolgee/unit/cachePurging/AWSCloudFrontStorageConfigCachePurgingTest.kt
```diff
@@ -0,0 +1,68 @@
...
+      whenever(cloudFrontClientBuilderMock.build()).thenReturn(cloudFrontClientMock)
+
+      whenever(
+        cloudFrontClientMock.createInvalidation(any<CreateInvalidationRequest>()),
+      ).thenReturn(createInvalidationResponseMock)
+
+      val contentDeliveryConfig = mock<ContentDeliveryConfig>()
+      whenever(contentDeliveryConfig.slug).thenReturn("fake-slug")
+
+      purging.purgeForPaths(
```
JanCizmar, line 62: The test will need some asserts that the cloudFrontClientMock was called with right parameters.

JanCizmar (changes requested):
Great job! Thanks a lot for the contribution 🚀 ❤️ 👍

backend/data/src/test/kotlin/io/tolgee/unit/cachePurging/AWSCloudFrontStorageConfigCachePurgingTest.kt
kingcdavid: Have added the verify's for this

qq: and probably i should have established prior to doing this work!
Is the CDN cache purging also available in the cloud version of Tolgee?

JanCizmar: So you can either add the UI in separate PR or you can create feature request issue.

kingcdavid: I think will need to open the feature request for the UI if thats OK?

JanCizmar: It is. :)

PR #2991, opened 2025-03-13
fix: translation sorting

stepan662:


backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/CursorPredicateProvider.kt
```diff
@@ -28,33 +28,60 @@ class CursorPredicateProvider(
...
+      val expression = selected as Expression<Comparable<Any>>
+
       val strongCondition: Predicate
       val condition: Predicate
+
       if (value.direction == Sort.Direction.ASC) {
         condition =
           if (isUnique) {
-            cb.greaterThan(expression, value.value!!)
+            cb.greaterThan(expression, typedValue!!)
```
JanCizmar, line 49: Instead of the `!!` I would rather do something like
```
?: throw IllegalStateException("keyId in the cursor cannot be null")
```

Because the isUnique is only there for the key.

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/CursorPredicateProvider.kt
```diff
@@ -28,33 +28,60 @@ class CursorPredicateProvider(
...
+    javaType: Class<*>,
+    raw: String?,
+  ): Comparable<Any>? {
+    if (raw == null) return null
+
+    return when (javaType) {
+      String::class.java -> raw as Comparable<Any>
+      java.lang.Long::class.java -> raw.toLong() as Comparable<Any>
+      java.sql.Timestamp::class.java -> java.sql.Timestamp(raw.toLong()) as Comparable<Any>
+      // If you use LocalDateTime, do:
```
JanCizmar, line 81: This comment looks unrelated.

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/CursorPredicateProvider.kt
```diff
@@ -28,33 +28,60 @@ class CursorPredicateProvider(
...
   }
+
+  @Suppress("UNCHECKED_CAST")
+  private fun parseValue(
+    javaType: Class<*>,
+    raw: String?,
+  ): Comparable<Any>? {
+    if (raw == null) return null
+
+    return when (javaType) {
```
JanCizmar, line 77: I don't get what this function is supposed to do.

It looks to me like it just converts some types to Comparable, which it already is, right? So what's the point?

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/CursorPredicateProvider.kt
```diff
@@ -28,33 +28,60 @@ class CursorPredicateProvider(
...
+      String::class.java -> raw as Comparable<Any>
+      java.lang.Long::class.java -> raw.toLong() as Comparable<Any>
+      java.sql.Timestamp::class.java -> java.sql.Timestamp(raw.toLong()) as Comparable<Any>
+      // If you use LocalDateTime, do:
+      // LocalDateTime::class.java -> LocalDateTime.parse(raw)
+      // or any other date/time you use
+      else -> throw Error("Cannot parse value for type $javaType")
```
JanCizmar, line 84: Throwing Error means, that the program should die, because it's not possible to recover from this. Throw IllegalArgumentException instead.

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/CursorPredicateProvider.kt
JanCizmar: I think you can just do 

```
raw ?: return null
return rawNotNull as? Comparable<Any> ?: throw IllegalArgumentException("Only comparable values can be used for sorting") 
```

JanCizmar (changes requested):
I will approve it without, but this functionality is so crucial, we should give it special care with tests in next iteration.

Including the tests in the UI that the scrolling works correctly with larger dataset for each of the sort options.

JanCizmar (approved):

PR #3016, opened 2025-03-26
feat: Add maintainer organization permission

sedyjaku:
Missing translations for all languages with codes
`organization_role_type_MAINTAINER`
`organization_role_type_MAINTAINER_hint`

also regenerate code to match new enum value for roles


JanCizmar:
@Anty0 will probably use it in his task.

Issue #3008, opened 2025-03-25
Enhancement of the API with the /languages to add updatedAt per language


stefanschaller:
**Is your feature request related to a problem? Please describe.**

Related to my Slack message in the support channel of "Tolgee community", I would love to extend the /language endpoint from the REST API.

Slack message: 

> Hey,
> First of all, thanks for the product—it’s amazing!
> I have a question regarding the REST API.
> My use case: I manage multiple languages, each with around 300 keys. I want to fetch translations in the mobile app every time the user starts the app, but only if the translations have changed.
> Currently, I request:
> /languages
> to retrieve the list of available languages. After receiving a list like ["de", "en", "es", "it", "fr", "pl"], I’d love to know if any of these languages have been updated.
> Is there a way to get the last updated timestamp for each language?
> Right now, I only see the “Last-Modified” date in the response header when requesting /translations/en. However, I noticed that modifying a language like "es" also changes the “Last-Modified” date for /translations/en, which is a bit confusing. Could you clarify how this works?
> Currently, I avoid updating all languages if the first requested language hasn’t changed, but this approach feels inefficient since I always need to request at least one language to determine if updates are necessary.
> Any insights would be greatly appreciated!
> TL;DR:
> I want to fetch the actual translations, when there are new one after a specific date.

TL;DR

I would like to extend the /languages endpoint with an updatedAt per language.

e.g.

```
{
    "_embedded": {
        "languages": [
            {
                "id": XXXX,
                "updatedAt": "2007-12-24T18:21Z",
                "name": "German",
                "tag": "de",
                "originalName": "Deutsch",
                "flagEmoji": "🇩🇪",
                "base": false
            },
            {
                "id": XXX,
                "updatedAt": "2007-12-24T18:21Z",
                "name": "English",
                "tag": "en",
                "originalName": "English",
                "flagEmoji": "🇬🇧",
                "base": true
            },
        ]
    },
    "_links": {
        "self": {
            "href": "XY"
        }
    },
    "page": {
        "size": 20,
        "totalElements": 6,
        "totalPages": 1,
        "number": 0
    }
}
```

**Additional context**
I need the updatedAt timestamp to ensure that translations are updated each time the mobile app starts, but only if there are new updates.

jdimeo:
[...]

PR #3018, opened 2025-03-27
feat: Add translation update time to the stats endpoint

JulianBissekkou:
We now expose a timestamp of the last modification as explain in #3008 
I planned to write tests, however I wasn't sure where and what kind of tests you prefer. I would like to have some guidance, and then I can add tests as well.

I tested this manually by updating translations for a language, and then I expected the timestamp to change.
In the video, you can see that I was checking the network tab in the devtools which also query the stats endpoint.

The interesting part is the new timestmap which changes from 
`1743062340802` which is `Thu Mar 27 2025 07:59:00.802` to
`1743063177609`  which is `Thu Mar 27 2025 08:12:57.609`

https://github.com/user-attachments/assets/40cf8990-c8da-4751-a79e-f526b5e3e54c

## Summary by CodeRabbit

- **New Features**
	- Language statistics now include a “Last Translation Update” timestamp, giving you clear insight into when translations were last refreshed.
  
- **Bug Fixes**
	- Improved data retrieval for language statistics to accurately reflect the last modified date of translations.

- **Tests**
	- Enhanced test coverage confirms the new timestamp is correctly reflected in language statistics responses.
	- Additional validation ensures updates to language statistics reflect the most recent translation update time.
	- New assertions added to verify the presence of the last translation update timestamp in responses.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
@@ -37,10 +37,7 @@ class LanguageStats(
 
   override var reviewedPercentage: Double = 0.0
 
-  override val translationsUpdatedAt: Date?
-    get() = language.translations
-      .mapNotNull { it.updatedAt ?: it.createdAt }
-      .maxByOrNull { it }
+  override var translationsUpdatedAt: Date? = null
 
   override val languageId: Long
     get() = language.id
--- a/backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
@@ -37,8 +37,10 @@ class LanguageStats(
 
   override var reviewedPercentage: Double = 0.0
 
-  override val lastModified: Date
-    get() = updatedAt ?: Date()
+  override val translationsUpdatedAt: Date?
+    get() = language.translations
+      .mapNotNull { it.updatedAt ?: it.createdAt }
+      .maxByOrNull { it }
 
   override val languageId: Long
     get() = language.id
@@ -56,6 +58,6 @@ class LanguageStats(
       untranslatedPercentage = untranslatedPercentage,
       translatedPercentage = translatedPercentage,
       reviewedPercentage = reviewedPercentage,
-      lastModified = lastModified
+      translationsUpdatedAt = translationsUpdatedAt
     )
 }
--- a/backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
@@ -37,8 +37,8 @@ class LanguageStats(
 
   override var reviewedPercentage: Double = 0.0
 
-  override val translationsUpdatedAt: Date?
-    get() = updatedAt
+  override val lastModified: Date
+    get() = updatedAt ?: Date()
 
   override val languageId: Long
     get() = language.id
@@ -56,6 +56,6 @@ class LanguageStats(
       untranslatedPercentage = untranslatedPercentage,
       translatedPercentage = translatedPercentage,
       reviewedPercentage = reviewedPercentage,
-      translationsUpdatedAt = translationsUpdatedAt
+      lastModified = lastModified
     )
 }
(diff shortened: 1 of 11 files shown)


backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
```diff
@@ -35,6 +36,10 @@ class LanguageStats(
   override var translatedPercentage: Double = 0.0
 
   override var reviewedPercentage: Double = 0.0
+
+  override val lastModified: Date
+    get() = updatedAt ?: Date()
+
```
JulianBissekkou, line 42: `updatedAt` is nullable in the parent class. We have the option of force unwrapping the value or we make the field nullable in all other DTOs or classes.

JanCizmar (changes requested):
Hey!

Thanks for the PR! 

I don't think this is the correct implementation. Since the language stats are updated asynchronously, the `updatedAt` in languageStats doesn't actually equal the last modification of the specific translation. 

We store the `updatedAt` for every standard entity, which might be confusing for your case.

I guess you will have to create a query returning the last modified for each language and store it here:

```io.tolgee.service.project.LanguageStatsService#refreshLanguageStats```

The new field should be named `dataUpdatedAt`.

**Comment by jdimeo** [comment]
Wow, this is amazing! One question is what terminology we want to standardize on. When I see
```
  override val lastModified: Date
    get() = updatedAt ?: Date()
```

I'm thinking we want to always/only use "last modified". I can't think of a good reason for one over the other, but it's less confusing if we only ever use one.

**Comment by stefanschaller** [comment]
@JanCizmar I'm unsure if we need a new field, since we could still use the `updatedAt` / `createdAt` from the actual translation list. 

I like the implementation of @JulianBissekkou, since it's super straightforward. 
Could you please take another look?

File: backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt [hunk 2]
```diff
@@ -35,6 +36,12 @@ class LanguageStats(
   override var translatedPercentage: Double = 0.0
 
   override var reviewedPercentage: Double = 0.0
+
+  override val translationsUpdatedAt: Date?
+    get() = language.translations
+      .mapNotNull { it.updatedAt ?: it.createdAt }
```
stefanschaller, line 42: Awesome, in case a new translatipn is added, the value of `translationsUpdatedAt` is still correct

JanCizmar: This is performance hell. What happens that all the translations for all languages will be fetched from database to memory. 

Also, maybe I am blind, but I don't think this will compile. You need max value of updated at, but this will return list of all translation `updatedAt` values.

What you ned to do is to run a single SQL returning max value for all languages and then assign it in the place I have suggested last time.

stefanschaller: How about not storing a new value in the DB, and we could execute something like: `SELECT MAX(updatedAt) latest_update FROM translations`;

It's simple for the database and straightforward implementation-wise. 

We're not backend experts since we're primarily mobile devs, which is why we'd prefer to avoid adding a new field and handling migrations 😆

stefanschaller:
@JanCizmar  I'm getting the following error: 

```
Caused by: org.postgresql.util.PSQLException: ERROR: column ls1_0.translations_updated_at does not exist
  Position: 261
        at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2725)
        at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2412)
        at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:371)
        at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:502)
        at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:419)
        at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:194)
        at org.postgresql.jdbc.PgPreparedStatement.executeQuery(PgPreparedStatement.java:137)
        at com.zaxxer.hikari.pool.ProxyPreparedStatement.executeQuery(ProxyPreparedStatement.java:52)
        at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeQuery(HikariProxyPreparedStatement.java)
        at org.hibernate.sql.results.jdbc.internal.DeferredResultSetAccess.executeQuery(DeferredResultSetAccess.java:246)
        ... 245 more
```

How do I write the migrations? Do you have some documentation on what is needed there?

backend/data/src/main/kotlin/io/tolgee/model/LanguageStats.kt
stefanschaller: ✅

stefanschaller:
@JanCizmar  Could you take a look?

JanCizmar (approved):

PR #3069, opened 2025-05-13
build: split standardTests to 2 jobs

laz-001:
This brings the runtime of the BT Matrix down from 10 to <7min. 

A runtime of 5min. should be doable with some more effort. 

But the current blocker is the cypress based E2E matrix (10min.). this could be brought down to around 6min.

In total, the ci runtime could be brought down from 15min. to <10min.

This PR would be the 1st step.

## Testrun

https://github.com/laz-001/tolgee-platform/actions/runs/14994330776

## Summary by CodeRabbit

- **Chores**
	- Improved test execution in the backend by splitting standard tests into two separate jobs for more efficient CI runs. No impact on application features or user experience.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,7 +74,8 @@ jobs:
         command:
           [
             "server-app:runContextRecreatingTests",
-            "server-app:runStandardTests",
+            "server-app:runStandardJob1Tests",
+            "server-app:runStandardJob2Tests",
             "server-app:runWebsocketTests",
             "server-app:runWithoutEeTests",
             "ee-test:test",
--- a/backend/app/build.gradle
+++ b/backend/app/build.gradle
@@ -228,6 +228,87 @@ tasks.register('runWithoutEeTests', Test) {
     rootProject.setTestRetry(it)
 }
 
+// Register a new Gradle test task that supports splitting test execution across CI jobs
+tasks.register('standardJobTests', Test) {
+    outputs.upToDateWhen { false }
+    useJUnitPlatform {
+        // These tests run in separate tasks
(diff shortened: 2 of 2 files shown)


JanCizmar:
Hey! Thanks for the PR! 🙏 

To be honest, I don't think this leads us to the right direction. We need to look more into why the actual tests run so slow instead of trying to optimize the run in the pipeline. 

I described this in the problem section of the issue:

> 1. Problems
> - It's time consuming to write integration test for Tolgee
> - Because of the build time might be lower
> - Because the context loads in 8 seconds
> - The time to run all integration tests is ~7mins

So the long CI/CR workflow run is just a secondary issue. The most important is to optimize it so it runs fast when we run a single test when writing the tests, because that's what slows us down every day.

I don't say this PR is totally off, I just say I would be happy if the individual tests run that fast, we wouldn't need splitting the groups.

laz-001:
I understand your concerns.

The CI run became important because your requirement to reduce the overall test-time (`gradlew test`) requires... test-runs. The iterations take horribly long.

CI is faster than my local machine, so I thought "OK, can't be so difficult, invest one or two days to parallelize them, so we have faster iterations, at least remote on CI". Well, overconfidence. It's a bit more complex. I'm back to the main problem.

Your prioritize "Single Test Runtimes". Though i was aware of this requirement, I thought that the `gradlew test` run was the top priority. In any way, the one radiates to the other.

At this point I'm deep in your test-code re 'full suite runtime'. This domain knowledge will allow me to fix the single test problem more efficient.

Priorities:

- single test
- gradlew test (currently working on this)
- ci run (future, but MUST be done, really... )

Issue #3000, opened 2025-03-18
500 while importing (Probably with meta)


JanCizmar:
**Describe the bug**
org.hibernate.TransientPropertyValueException: object references an unsaved transient instance - save the transient instance before flushing : io.tolgee.model.dataImport.ImportKey.keyMeta -> io.tolgee.model.key.KeyMeta

**To Reproduce**
Don't know how to reproduce, but there should be enough info in Sentry

**Expected behavior**
Should import successfully.

**Stack trace**
```
org.hibernate.TransientPropertyValueException: object references an unsaved transient instance - save the transient instance before flushing : io.tolgee.model.dataImport.ImportKey.keyMeta -> io.tolgee.model.key.KeyMeta
    at org.hibernate.engine.spi.CascadingActions$8.noCascade(CascadingActions.java:372)
    at org.hibernate.engine.internal.Cascade.cascade(Cascade.java:173)
    at org.hibernate.event.internal.AbstractFlushingEventListener.cascadeOnFlush(AbstractFlushingEventListener.java:169)
    at org.hibernate.event.internal.AbstractFlushingEventListener.prepareEntityFlushes(AbstractFlushingEventListener.java:152)
    at org.hibernate.event.internal.AbstractFlushingEventListener.preFlush(AbstractFlushingEventListener.java:102)
    at org.hibernate.event.internal.DefaultAutoFlushEventListener.onAutoPreFlush(DefaultAutoFlushEventListener.java:103)
    at org.hibernate.event.service.internal.EventListenerGroupImpl.fireEventOnEachListener(EventListenerGroupImpl.java:127)
    at org.hibernate.internal.SessionImpl.autoPreFlush(SessionImpl.java:1387)
    at org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan.withCacheableSqmInterpretation(ConcreteSqmSelectQueryPlan.java:322)
    at org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan.performList(ConcreteSqmSelectQueryPlan.java:302)
    at org.hibernate.query.sqm.internal.QuerySqmImpl.doList(QuerySqmImpl.java:526)
    at org.hibernate.query.spi.AbstractSelectionQuery.list(AbstractSelectionQuery.java:423)
    at org.hibernate.query.Query.getResultList(Query.java:120)
[...]

PR #3113, opened 2025-06-09
fix: import with meta without creating new keys  (#3000)

bdshadow:
When `shouldImportKey` is called in the `FileProcessorContext.processTranslations()` method, Hibernate does the flush before executing a select query. At this flush, it discovers that `ImportKey` has `KeyMeta`, which is newly created and needs to be flushed, however, there is no cascading. 

So cadacding would be a good solution. It is also the first one which is recommended in many articles, e.g. https://www.baeldung.com/hibernate-unsaved-transient-instance-error, as well as cascading is usually (sure, not always) applied in OneToOne relations: https://vladmihalcea.com/the-best-way-to-map-a-onetoone-relationship-with-jpa-and-hibernate/. But when I started applying cascading, I ended up modifying too many entities, which led to many unpredictable problems

So I think this solution of saving a `KeyMeta` alongside the `ImportKey` is a good compromise here.

Probably the test can be moved to a more suitable place. (I'm just getting acquainted with the code base). It can easily be replicated also with the `strings.xml` file mentioned in [the comment](https://github.com/tolgee/tolgee-platform/issues/3000#issuecomment-2865753203)


backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
```diff
@@ -297,7 +301,9 @@ class CoreImportFilesProcessor(
       this.keys.computeIfAbsent(name) {
         ImportKey(name = name, this.fileEntity)
       }.also {
+        it.keyMeta?.also { keyMeta -> importDataManager.prepareKeyMeta(keyMeta) }
```
JanCizmar: I am afraid that these changes might affect performance. In the case of imports the performance is crucial because we are handling huge amount of saves. Are you sure this is fine?

It looks to me like we are saving key metas even if they are already exist, which might cost us some resources...

backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
```diff
@@ -312,10 +318,8 @@ class CoreImportFilesProcessor(
       }
       keyEntity.shouldBeImported = shouldImportKey(keyEntity.name)
     }
-    importDataManager.prepareKeyMetas()
```
JanCizmar, line 315: this methods are unused now, right? So they should be removed.

backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
```diff
@@ -312,10 +318,8 @@ class CoreImportFilesProcessor(
...
     if (saveData) {
       importDataManager.saveAllStoredKeys()
-      importDataManager.saveAllKeyMetas()
```
JanCizmar, line 318: this methods are unused now, right? So they should be removed.

backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
bdshadow: `repository.save` method doesn't execute an `update`query if entity is not marked dirty. For example the following snippet doesn't issue "update" to db:
```java
Optional<Entity> entity = reposRepository.findById(id);
entity.ifPresent(existingEntity -> {
            existingEntity.setUrl(existingEntity.getUrl());
            reposRepository.save(existingEntity);
            });
```

When saving keys, isn't it correct to save KeyMeta? What's done in the fix is actually just applying the cascading manually.

Regarding the performance, of course, it needs to be tested, but I suppose that it should not be affected since the number of requests to db must not change. Is there an example of a such huge files for imports?

backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
bdshadow: done

backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
bdshadow: done

backend/data/src/main/kotlin/io/tolgee/service/dataImport/CoreImportFilesProcessor.kt
bdshadow: i've added another change. Now it works correctly. I've tested it on the "example.po". This fix generates two additional "update" queries for key_meta. I will try to get rid of it. Meanwhile will move this pr do drafts

JanCizmar: Thanks a lot! I believe I have separated all the saves to groups, so hibernate can execute the updates in single statement... However, I am not hibernate expert so if you are sure it's not harming the performance, we can merge it. However, it's a good idea to test it with some larger dataset like 10000 keys and see how fast it is.

JanCizmar: Also, as stated in this article, it's not always that straightforward with the save() method. Read more in the "The redundant save anti-pattern" section. https://vladmihalcea.com/jpa-persist-and-merge/

bdshadow: @JanCizmar  it's ready for review. Now, the number of queries to the database is the same as it was before, plus it fixes the issue.

It was not only about the `save` method, but more about the order of saving entities.  

My opinion is that current mapping can lead to unexpected bugs like this. However, as they say, if you criticize, offer a solution :) I'm sure there is a reason behind such relations between entities, that I don't see at the moment. I will get more familiar with it and with the code in general. Then I'll be able to offer a solution... or will admit that it's ok :)

JanCizmar: OK, thanks a lot for the PR. I will merge it and let's see what happens.

PR #3115, opened 2025-06-09
fix: conditional glossaries load only if the feature is available

Anty0:
## Summary by CodeRabbit

- **Bug Fixes**
  - Loading indicators for glossaries are now only shown when the glossary feature is enabled, preventing unnecessary loading states when the feature is disabled.  
- **Chores**
  - Improved clarity by renaming the glossary feature flag variable throughout the glossary-related components and hooks.

### DIFF
--- a/webapp/src/ee/glossary/views/GlossariesListView.tsx
+++ b/webapp/src/ee/glossary/views/GlossariesListView.tsx
@@ -45,6 +45,7 @@ export const GlossariesListView = () => {
       sort: ['id,desc'],
     },
     options: {
+      enabled: glossaryFeature,
       keepPreviousData: true,
     },
   });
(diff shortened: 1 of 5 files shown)


webapp/src/ee/glossary/views/GlossariesListView.tsx
```diff
@@ -45,6 +45,7 @@ export const GlossariesListView = () => {
       sort: ['id,desc'],
     },
     options: {
+      enabled: glossaryFeature,
```
JanCizmar: `glossaryFeature` can mean a lot of things. `isFeatureEnabled` would tell more. `glossaryFeatureEnabled` would tell even more.

Anty0: True, sounds much better that way.
I probably copied it from somewhere like `MyTasksList`, so I didn't pay much attention to it.

PR #3122, opened 2025-06-11
feat: add glossary_id for deepl usage

swellner:
Added the ability to configure a glossary_id in the server config for deepl. This is needed to translate texts with a custom configured glossory from deepl.

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a DeepL glossary ID when using machine translation, allowing users to leverage custom glossaries for improved translation accuracy.
  - Introduced the ability to include optional DeepL API parameters to customize translation requests further.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/DeeplApiService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/DeeplApiService.kt
@@ -34,7 +34,9 @@ class DeeplApiService(
     requestBody.add("text", text)
     requestBody.add("source_lang", sourceTag.uppercase())
     requestBody.add("target_lang", targetTag.uppercase())
-
+    deeplMachineTranslationProperties.glossaryId?.let {
+      requestBody.add("glossary_id", it)
+    }
     addFormality(requestBody, formality)
 
     val response =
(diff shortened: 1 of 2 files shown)


swellner:
i changed the implementation to a map of generic parametes which can be used for the deepl api

backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/DeeplApiService.kt
```diff
@@ -34,7 +34,9 @@ class DeeplApiService(
     requestBody.add("text", text)
     requestBody.add("source_lang", sourceTag.uppercase())
     requestBody.add("target_lang", targetTag.uppercase())
-
+    deeplMachineTranslationProperties.optionalParameters?.map {
```
JanCizmar: since we don't need to return anything we should go for`forEach` rather then map.

JanCizmar (changes requested):

JanCizmar:
Nice! Thanks a lot for the PR!

backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/DeeplApiService.kt
swellner: Yey you are right. I changed it. Thanks.

swellner:
One test is failing
`CreditLimitTest > correctly propagates credit spending limit exceeded() FAILED
    java.lang.AssertionError at CreditLimitTest.kt:110`
 The tests are running locally correctly.
 
 Is there something i missed?

swellner:
Tests seems flickering. Now another test case failed without any code change.

`SeatCountLimitTest > does not throw when removing users() FAILED
    io.tolgee.util.RepeatedlyThrowingConstraintViolationException at SeatCountLimitTest.kt:89
        Caused by: org.hibernate.exception.GenericJDBCException at SeatCountLimitTest.kt:89
            Caused by: org.postgresql.util.PSQLException at SeatCountLimitTest.kt:89
                Caused by: org.postgresql.util.PSQLException at SeatCountLimitTest.kt:89`

JanCizmar:
You're right. It's an intermittent test! Thanks a lot for the PR!

PR #3147, opened 2025-06-24
fix: project hard delete must also remove import settings and glossaries

Anty0:
## Summary by CodeRabbit

- **New Features**
  - Project deletion now also removes all related glossary assignments and import settings, ensuring cleaner data removal.
- **Bug Fixes**
  - Glossary unassignment during project deletion now includes glossaries regardless of their previous deletion status.

### DIFF
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/glossary/GlossaryService.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/glossary/GlossaryService.kt
@@ -149,7 +149,7 @@ class GlossaryService(
     glossaryRepository.unassignProject(organizationId, glossaryId, projectId)
   }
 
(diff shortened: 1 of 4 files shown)


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/glossary/GlossaryService.kt
```diff
@@ -144,4 +148,8 @@ class GlossaryService(
   ) {
     glossaryRepository.unassignProject(organizationId, glossaryId, projectId)
   }
+
+  override fun unassignProjectFromAll(project: Project) {
+    glossaryRepository.unassignProjectFromAll(project.id)
```
stepan662, line 153: This name is kinda weird, shouldn't it be `unassignFromAllProjects`?

stepan662 (approved):

PR #3155, opened 2025-06-27
feat: subscription name consistency

dkrizan:
https://github.com/tolgee/billing/issues/157


backend/data/src/main/kotlin/io/tolgee/constants/Message.kt
```diff
@@ -295,6 +295,7 @@ enum class Message {
   GLOSSARY_NON_TRANSLATABLE_TERM_CANNOT_BE_TRANSLATED,
   LLM_CONTENT_FILTER,
   LLM_PROVIDER_EMPTY_RESPONSE,
+  STRIPE_PRODUCT_CREATION_FAILED,
```
JanCizmar: This should be deleted (view the comment in billing repo)

JanCizmar (changes requested):

PR #3164, opened 2025-07-06
feat: codeql workflow

transparentChange:
Helps identify security vulnerabilities in the webapp ts code through CodeQL static security scanning. Results will be uploaded to the Github Security tab that will show potential security flaws and help monitor code health. 

## Summary by CodeRabbit

* **Chores**
  * Added automated security analysis using CodeQL, which runs on key branches and on a weekly schedule to help identify potential vulnerabilities in TypeScript code.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,32 @@
+name: CodeQL Scan
+
+on:
+  push:
+    branches: [main, test-codeql]
+    paths: ['**/*.ts', '**/*.tsx']
+  pull_request:
+    branches: [main]
+    paths: ['**/*.ts', '**/*.tsx']
+  schedule:
+    - cron: '0 4 * * 1'  # Weekly on Monday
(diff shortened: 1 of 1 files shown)


JanCizmar:
The issues won't probably be on the typescript side, but on the kotlin side. However, we are already using Snyk and we're happy with that.

transparentChange:
@JanCizmar There are some issues on the typescript side. You can see them be looking at the SARIF file generated as an artifact by the run here: [https://github.com/hydrofluorich/cs489-tolgee-platform/actions/runs/16203273005](url).

Also I understand you have Snyk setup through a github integration. If you are interested in giving the vulnerabilities more visibility then using the Snyk CLI could be a good solution. An example run can be found here: [https://github.com/hydrofluorich/cs489-tolgee-platform/actions/runs/16103198618/job/45435024145 ](url) with the output under "Run Snyk and capture result"

PR #3180, opened 2025-07-25
feat: shows a "VAT taxable" message with the required checkbox in Stripe

dkrizan:


JanCizmar (changes requested):
There is already a localization utility: io.tolgee.util.I18n

We should use the stanrdard way.

Is there any reason why to make theses changes in the public repo?

dkrizan:
> There is already a localization utility: io.tolgee.util.I18n
> 
> We should use the stanrdard way.
> 
> Is there any reason why to make theses changes in the public repo?

Thanks, didn't know about i18n util. I extended i18n util to support mutliple locales.

dzienisz:
I think we should update this Stripe dep.

JanCizmar:
We decided not to proceed with this as it looks pretty bad, we decieded to go with other approach.

PR #3184, opened 2025-07-30
fix: Server configuration property generation

JanCizmar:
## Summary by CodeRabbit

* **Documentation**
  * Improved code comments to clarify handling of simple property types, including Map types and enums, in configuration documentation.

* **Refactor**
  * Enhanced recognition of Map types as simple properties for consistent processing.

* **Tests**
  * Added integration tests to verify the configuration properties endpoint returns expected structured data.


JanCizmar:
> Could we add a test to the configuration enpdoint that it doesn't fail?

Yes, good idea.

PR #3193, opened 2025-08-07
feat: Optimize Dockerfile

transparentChange:
Changes the Dockerfile to use a base image without postgresql. Instead, the external postgres is used when running with docker-compose. This change significantly reduces build size and performance.

docker images local output:
image

## Summary by CodeRabbit

* **New Features**
  * Added Docker Compose setup including app, database, and SMTP services.
  * Introduced a lightweight application image with built-in health checks and a persistent data volume.
  * Enabled local image builds via Docker Compose instead of pulling a prebuilt image.

* **Chores**
  * Added an environment variable to control Postgres autostart (disabled by default).
  * Standardized the app startup script to use sh for broader compatibility.


transparentChange:
@JanCizmar @Anty0 Requesting your review

Anty0:
@transparentChange Hi! Thanks for the PR!

We discussed the possibility of a similar change internally. Unfortunately, removing the bundled Postgres instance wouldn't be straightforward, as users already rely on it. Implementing this change would require some manual migration steps for them.

One idea would be to create a separate Docker image without the bundled Postgres, while keeping the existing one with the bundled Postgres. New self-hosters could then use the new image, and we could keep the old one for existing deployments.

JanCizmar:
@transparentChange However, we can create separate "lite" image without postgres. But I would not change the default `tolgee/tolgee`.

transparentChange:
@JanCizmar Thank you for the review. Like you suggested I added the new Dockerfile.lite file and a comment for how it can be hooked up to the docker-compose.

JanCizmar:
@Anty0 can you look into this when you have some time? It stays in my backlog for a long time already and I don't want to ghost @transparentChange

PR #3194, opened 2025-08-08
feat: React Emails > default template

JanCizmar:


backend/data/src/main/kotlin/io/tolgee/email/EmailGlobalVariablesProvider.kt
```diff
@@ -34,7 +34,7 @@ class EmailGlobalVariablesProvider(
     return mapOf(
       "isCloud" to isCloud,
       "instanceQualifier" to if (isCloud) tolgeeProperties.appName else tolgeeProperties.frontEndUrl.intoQualifier(),
-      "instanceUrl" to tolgeeProperties.frontEndUrl,
+      "instanceUrl" to "http://localhost:8080",
```
cyyynthia: Seeing that it makes me wonder if the issues you were experiencing are that in development the frontEndUrl points to the Vite server and it isn't configured to proxy `/static/*` to the backend... That'd explain it

email/components/Var.ts
```diff
@@ -27,9 +27,16 @@ export default function Var({
...
+  // eslint-disable-next-line react/no-danger-with-children
   return React.createElement(
     'span',
-    { [injectHtml ? 'th:utext' : 'th:text']: `\${${variable}}` },
-    demoValue
+    {
+      [injectHtml ? 'th:utext' : 'th:text']: `\${${variable}}`,
+      dangerouslySetInnerHTML: { __html: html },
+    },
+    children
```
cyyynthia:
```suggestion
    injectHtml
        ? {
          ['th:utext']: `\${${variable}}`,
          dangerouslySetInnerHTML: { __html: demoValue },
        }
        : {
          ['th:text']: `\${${variable}}`,
          children: demoValue
        }
```

backend/data/src/main/kotlin/io/tolgee/component/email/TolgeeEmailSender.kt
```diff
@@ -2,50 +2,28 @@ package io.tolgee.component.email
...
-      { ClassPathResource("tolgee-logo.png").inputStream },
-      "image/png",
+    val properties = mapOf<String, Any>()
+      .let { if (params.text != null) it.plus("content" to params.text!!) else it }
+      .let { if (params.header != null) it.plus("header" to params.header!!) else it }
+      .let { if (params.recipientName != null) it.plus("recipientName" to params.recipientName!!) else it }
+    emailService.sendEmailTemplate(
+      recipient = params.to,
+      subject = params.subject,
+      template = params.templateName ?: "default",
```
cyyynthia, line 20: FWIW, I recon my original idea was to eventually deprecate that component; i.e. new code would use the new email system while legacy code can still use this until it is eventually migrated to use a dedicated template. Not sure templateName is needed 😅

backend/data/src/main/kotlin/io/tolgee/email/EmailGlobalVariablesProvider.kt
```diff
@@ -27,14 +28,16 @@ class EmailGlobalVariablesProvider(
...
   operator fun invoke(): Map<String, Any?> {
     val isCloud = billingConfigProvider().enabled
+    val backendUrl = tolgeeProperties.backEndUrl ?: frontendUrlProvider.url
 
     return mapOf(
       "isCloud" to isCloud,
-      "instanceQualifier" to if (isCloud) tolgeeProperties.appName else tolgeeProperties.frontEndUrl.intoQualifier(),
-      "instanceUrl" to tolgeeProperties.frontEndUrl,
+      "instanceQualifier" to if (isCloud) tolgeeProperties.appName else backendUrl.intoQualifier(),
+      "backendUrl" to backendUrl
```
cyyynthia, line 40: Unsure about that... Especially wrt the instance qualifier. The fallback to using the domain name people are using to access Tolgee is likely a lot more recognizable than whatever internal URL is used.

Problems in development environment are likely related to the fact I completely overlooked that and didn't add the appropriate proxy config to Vite 😅

backend/data/src/main/kotlin/io/tolgee/email/EmailService.kt
```diff
@@ -63,7 +64,7 @@ class EmailService(
     context.setVariable(ThymeleafEvaluationContext.THYMELEAF_EVALUATION_CONTEXT_CONTEXT_VARIABLE_NAME, tec)
 
     val html = templateEngine.process(template, context)
-    val subject = extractEmailTitle(html)
+    val subject = subject ?: extractEmailTitle(html)
```
cyyynthia, line 69: Note: Setting a subject would break internationalization of emails. I guess this is needed for the backcompat layer, but... not ideal as it makes the service contract loose here :(

I don't think the template allows the subject being a variable directly but that could be mitigated with a passthrough i18n string; i.e. `email.legacy.subject-passthrough = "{subject}"`

email/emails/registration-confirm.tsx
```diff
@@ -40,9 +41,18 @@ export default function RegistrationConfirmEmail() {
...
       </Text>
       <Text>
+        <If condition="${isSignup}">
+          <If.Then>
+            <LocalizedText
+              keyName="registration-welcome-text"
+              defaultValue="Welcome and thank you for creating an account!"
+            />{' '}
+          </If.Then>
+          <If.Else></If.Else>
```
cyyynthia: Nit: `If.Else` isn't strictly mandatory; looking back at the HACKING.md note I see where it comes from but I more or less wanted to express that bare JSX as children of `If` isn't great because it's hard to tell what part of an `if-then-else` we're working with.

Empty else doesn't hurt, but omitting it is fine imho.

email/emails/registration-confirm.tsx
```diff
@@ -40,9 +41,18 @@ export default function RegistrationConfirmEmail() {
         />
       </Text>
       <Text>
+        <If condition="${isSignup}">
```
cyyynthia, line 44: Not sure using the same template for registration and email changes is advisable (given the advanced templating system now); subject of the email might be ambiguous and it limits the extensibility of the registration email (e.g. adding additional sections such as getting started guides and what not).

The wording of the `confirm-email-text` is a bit confusing too; people aren't "starting to use" Tolgee but rather "changing their email" and I think the email body should reflect that.

Related to the 1st point email change confirmation (and other important emails) should perhaps contain warnings and advise caution regarding phishing campaigns and what to do in case of suspicious activity (use MFA, change passwords, etc).

cyyynthia (commented):
Leaving some feedback for consideration 😄

backend/data/src/main/kotlin/io/tolgee/component/email/TolgeeEmailSender.kt
dkrizan: We agreed to wrap all existing emails in a default template and just drop the legacy email text into it. Since the new email designs aren’t ready yet, this lets us still use the feature in the meantime 😀

email/emails/registration-confirm.tsx
dkrizan: i am afraid the reason i put it there is, that it actually didn't work without `<If.Else></If.Else>` :man_shrugging: (I guess `<If.Else />` shoud be good too) .. I don't know why, I was too lazy to find out 😁 

btw this template is not used anyway, so it might be deleted

cyyynthia: I initially included it as a showcase/"real-world" demo (and was the template I used while prototyping back then); but it can indeed be ditched 😅

Weird that the Else branch is required; I know for sure it didn't use to be but I might've enforced it when refactoring at some point. Not very important anyway... :)

dzienisz:
@dkrizan can you add screenshot of those emails before and after the change?

dkrizan:
**Sign-up email (verify email)**

image


**Password reset**

image

**Project invitation**

image

dkrizan:
**Email Notification**

image


**Invoice email**

image


**EE license key**

image

PR #3199, opened 2025-08-12
fix: change base for UnsupportedXliffVersionException to avoid 500

Anty0:
Changed base of the `UnsupportedXliffVersionException` to Exception. This way, it should get caught and converted to `ImportCannotParseFileException`, avoiding a 500 error.

## Summary by CodeRabbit

* **Refactor**
  * Standardized internal error handling to use platform-standard exceptions for a specific XLIFF version error. No changes to behavior, APIs, or settings; users will see unchanged functionality. Improves consistency, logging clarity, and future maintainability. No action required by administrators or end users.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/exceptions/UnsupportedXliffVersionException.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/exceptions/UnsupportedXliffVersionException.kt
@@ -1,5 +1,4 @@
 package io.tolgee.exceptions
 
-class UnsupportedXliffVersionException(version: String) : Exception() {
-  override val message: String = "XLIFF version $version not supported."
-}
+class UnsupportedXliffVersionException(version: String) :
+  Exception("XLIFF version $version not supported.")


backend/data/src/main/kotlin/io/tolgee/exceptions/UnsupportedXliffVersionException.kt
```diff
@@ -1,5 +1,4 @@
 package io.tolgee.exceptions
 
-class UnsupportedXliffVersionException(version: String) : Throwable() {
-  override val message: String = "XLIFF version $version not supported."
-}
+class UnsupportedXliffVersionException(version: String) :
```
JanCizmar, line 3: Afaik, you would have to go with BadRequestException?

JanCizmar: Or am I wrong?

Anty0: It has only one use within the codebase, where it's changed to `ImportCannotParseFileException` almost immediately:
https://github.com/tolgee/tolgee-platform/blob/8c5ca2a35085fd8b700a4ca2dbc9df93d748ecc2/backend/data/src/main/kotlin/io/tolgee/formats/xliff/in/XliffFileProcessor.kt#L26-L34

It might be a bit confusing like this - maybe it should be based on `ImportCannotParseFileException` instead.

The issue was that the exception was based on `Throwable` while the try-catch was only catching an `Exception`, so I changed it to an `Exception`. ^^

JanCizmar (approved):

PR #3201, opened 2025-08-13
feat: tolgee platform client

stepan662:
## Summary by CodeRabbit

* **New Features**
  - Introduced the @tginternal/client package: a typed API client with API-key and user-token auth, improved error messages, utilities for key info and unresolved-translation reporting, and a consolidated public API.
  - Automated publish workflow to release the client package on demand and after releases.

* **Chores**
  - Added build/config tooling, version generation, and package manifest for the client.
  - Expanded CI to include client build/tests.

---

## Additions on top of the original client package

This branch also lands **typed webhooks** so anything consuming `@tginternal/client` (apps, plugins, custom webhook receivers) can type-check webhook payloads end-to-end without hand-maintaining shapes.

### Backend — OpenAPI 3.1 `webhooks` section

- `application.yaml` now emits **OpenAPI 3.1** (`springdoc.api-docs.version=openapi_3_1`).
- New `AppWebhookOpenApiCustomizer` injects a `webhooks` entry per visible `ActivityType` into the "All Internal" group. Per-event payload schemas honour the three size-control flags on `ActivityType`:
  - `onlyCountsInList=true` → `modifiedEntities` is constrained to `{}`; signal moves to `counts` and `params`.
  - `restrictEntitiesInList=[X::class, …]` → `modifiedEntities` is restricted to those entity classes.
  - `paramsProvider=…` → a typed `params` property appears on `activityData`.
- Per-entity `<EntityClass>Modifications` schemas (`TranslationModifications`, `KeyModifications`, …) are reflectively built from `@ActivityLoggedProp` annotations on each `@ActivityLoggedEntity`.
- `OpenApiUnusedSchemaCleaner` now also walks `openAPI.webhooks` so synthesized payload schemas survive the cleanup pass.

### Tests

- **`WebhookSpecGenerationTest`** — 214 parameterized cases prove the *spec* is well-formed: every visible `ActivityType` has a webhook entry with a `$ref` to a payload schema, payload schemas honour the three flag-driven shape rules, and per-entity Modifications schemas mirror their `@ActivityLoggedProp` annotations.
- **`WebhookContractFixture`** + 4 contract test classes — prove the *wire payload* matches the spec. Pattern: per-event test class with one trigger method calling the real controller/service path, then `fixture.waitForWebhookWithType(event)` + `fixture.assertConformsTo(payload, event)`. **No abstract base class** (team convention) — every helper is called via an explicit `fixture.X(…)` so origin is grep-able from the call site.
  - `SetTranslationsWebhookContractTest` — full-data, single entity.
  - `KeyNameEditWebhookContractTest` — full-data + asserts the `modifications.name.{old,new}` diff.
  - `CreateLanguageWebhookContractTest` — full-data, Language only.
  - `DeleteLanguageWebhookContractTest` — proves `restrictEntitiesInList = [Language::class]` is honoured even when the cascade deletes translations.

### Client (`@tginternal/client`)

- `openapi-typescript` is already on `^7.9.1` here (vs the webapp's 5.x), so the generated `src/schema.generated.ts` carries the `webhooks` namespace automatically.
- README documents the consumer pattern:
  ```ts
  import type { webhooks } from "@tginternal/client";
  type SetTranslationsPayload =
    webhooks["SET_TRANSLATIONS"]["post"]["requestBody"]["content"]["application/json"];
  ```
- Discriminated-union usage via `activityData.type` is documented too, plus the HMAC verification recipe for the `Tolgee-Signature` header.

### Deferred (clean follow-ups)

- Contract tests for `IMPORT` (counts-only — needs `ImportService.import(...)` driver), `BATCH_MACHINE_TRANSLATE` (counts-only + `paramsProvider` — `@Import(BatchJobBaseConfiguration::class)` interacts oddly with `@MockitoBean` wiring; needs its own debugging session), and `BRANCH_MERGE` (EE-only). The pattern is established — each is a copy-paste-and-swap of an existing class.
- Webapp's `openapi-typescript` 5.4.2 → 7.x bump (~40 `useApiQuery`/`useApiMutation` call sites need updates due to operation-type shape changes). Webapp's 5.x still tolerates 3.1 input for the schema it consumes; bump can ship as its own scope.

### Versions published

- `@tginternal/client@0.0.0-prerelease.c8329b552` (workflow_dispatch on this branch)
- Additional docs commit on top — next `workflow_dispatch` will publish `…4ca8a9a4`.


dzienisz (approved):

JanCizmar:
@Anty0, @dkrizan can you please get familiar with this and maybe do a CR? I think this will affect our future integration development. 

Thanks @stepan662 for finishing this! 🙏

.github/workflows/publish-client.yml
```diff
@@ -0,0 +1,91 @@
+name: Publish client
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: ["Test"]
+    branches: ["main"]
+    types:
+      - completed
```
JanCizmar, line 9: Probably should be released when release (not the "Test" one) workflow is completed successfully.

.github/workflows/test.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: Setup environment
         uses: ./.github/actions/setup-env
         with:
-          node: 'false'
+          node: "false"
```
JanCizmar, line 69: Probably should be reverted. Looks like we have different formatting preferences.

JanCizmar (changes requested):

client/src/ApiClient.ts
```diff
@@ -0,0 +1,174 @@
...
+  autoThrow?: boolean;
+  verbose?: boolean;
+  userToken?: string;
+  printError?: (loadable: LoadableData) => string;
+};
+
+export function createApiClient(props: ApiClientProps) {
+  let {
+    projectId,
+    baseUrl = "https://app.tolgee.io",
```
dkrizan: this could be in `constants.ts` too

client/src/ApiClient.ts
```diff
@@ -0,0 +1,174 @@
...
+        request.headers.set("Authorization", "Bearer " + userToken);
+      } else if (apiKey) {
+        request.headers.set("x-api-key", apiKey);
+      }
+    },
+    onResponse: async ({ response, options }) => {
+      let responseText = `[HTTP] Response: ${response.url} [${response.status}]`;
+      const apiVersion = response.headers.get("x-tolgee-version");
+      if (apiVersion) {
+        responseText += ` [${response.headers.get("x-tolgee-version")}]`;
```
dkrizan:
```js
responseText += ` [${apiVersion}]`;
```

dkrizan (commented):
Code looks good! Left a few minor notes.

JanCizmar:
@Anty0, can you please finish this, so we have it done?

client/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "erasableSyntaxOnly": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUncheckedSideEffectImports": true,
+
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "noEmit": false,
+    "outDir": "./lib"
+  },
+  "include": ["src"]
```
zomp: Does this work? Should not it be `"src/**/*"`?

JanCizmar:
Picking this up to address the open review feedback (so we can use this client from the upcoming Tolgee Apps SDK). The new commit (`4d6add8`) applies:

**From @JanCizmar's review**
- `publish-client.yml`: trigger on the **Release** workflow (not Test), so publishing only happens after a successful release build.
- `test.yml`: reverted Stepan's diff against `main` — restores `buildTestEmails`, the sharded matrix, `SKIP_EMAIL_BUILD`, CTRF reporting, etc.

**From @dkrizan's review**
- Default base URL `https://app.tolgee.io` moved to `constants.ts` as `DEFAULT_BASE_URL`.
- The verbose response log reuses the captured `apiVersion` variable instead of re-reading the header.

**Outstanding CodeRabbit items**
- `errorFromLoadable.ts`: real bug — was passing the full `params` array to `getUnresolvedConflictsMessage` instead of `params[0]` (the conflicts array). Fixed.
- `ApiClient.ts#getApiKeyInfo`: replaced the `apiKey!` non-null assertion with an explicit error when no API key is configured.
- `publish-client.yml`: the version-existence check now uses `npm view ...@<version> version` (singular) — `versions --json` succeeds whenever **any** version of the package exists, so the original check was always reporting "already exists" once the package was published once.
- `printFailedKeys.ts`: dropped the leading `""` in the result buffer that combined with the caller's prepended `\n` to produce double spacing.
- `tsconfig.json`: excluded `*.test.ts` from build emission so `vitest/globals` doesn't leak into published `.d.ts`. Removed the unused `"node"` types entry (no source imports `node:*`).
- `vite-env.d.ts`: deleted — nothing in `src/` uses Vite-specific globals (`import.meta.env` etc.), and the `/// <reference types="vite/client" />` would have leaked Vite ambient types to consumers.

**Items left as-is** (please push back if you want them addressed):
- CodeRabbit comments on `TolgeeClient.ts`, `ExportClient.ts`, `ImportClient.ts`, `pathToPosix.ts` — those files were removed from the PR after Stepan's Aug 15 batch.
- The `updateVersion.js:11` "regex too broad" comment — the current implementation uses `JSON.parse`/`stringify` and no longer relies on a regex.
- The `vite.config.js:19` "invalid UMD global name" comment — the lib name was already changed from `@tolgee/client` to `client` (valid JS identifier).

**Verification**
- `cd client && npm run build` — green.
- `cd client && npm test` — 4/4 green.
- `git diff origin/main -- .github/workflows/test.yml` — empty (revert is complete).

@JanCizmar @dkrizan — could you take another look when convenient?

JanCizmar:
Force-pushed a rebased + cleaned-up version of the branch. The full PR diff against `main` is now just three paths: `client/`, `.github/workflows/publish-client.yml`, and a small additive change to `.github/workflows/test.yml` (one new `client-checks` job). All other diffs from the previous push (which had drifted with main moving forward) are gone.

Additional changes on top of the earlier review-feedback round:

1. **`base32-decode` moved to `dependencies`** — it was imported at runtime but declared as a devDependency. Worked only because Vite's lib mode inlines deps; would have broken the moment `rollupOptions.external` was added.
2. **`client/README.md`** — short install + usage doc covering `createApiClient`, the main helpers, and schema regeneration.
3. **`schema` script now reads `TOLGEE_API_PORT`** (default `8080`, so it works against a local `bootRun`). The publish workflow exports `TOLGEE_API_PORT=8201` to match `runDockerE2e`.
4. **Test coverage broadened**: 13 tests (up from 4) — covers `getProjectId`/`getApiKeyInfo` guard paths, `errorFromLoadable` status-code mapping, and `getUnresolvedConflictsMessage` formatting (including the `params[0]` bug fix from the previous round).
5. **`getSettings()` now returns `printError`** so the snapshot matches `ApiClientProps`.
6. **`exports` map now includes `default`** for forward-compat with stricter package resolvers.
7. **New `client-checks` job in `test.yml`** runs `npm ci && npm run build && npm test` on every PR; added to `everything-passed.needs` so it gates the green check.

The schema port via env var is bash-only (`${TOLGEE_API_PORT:-8080}`), which matches the rest of tolgee's npm scripts.

Versioning is unchanged: client publishes as `@tginternal/client@<platform-version>` on release, `0.0.0-prerelease.<sha>` otherwise (via the `verifyReleaseCmd` hook that writes `.VERSION`).

Verified locally:
- `git diff origin/main -- .github/workflows/test.yml` — only the new `client-checks` job
- `npm run build` — green
- `npm test` — 13/13 green

PR #3203, opened 2025-08-13
feat: Allow websocket subscription with PAK/PAT

JanCizmar:


JanCizmar:
I am working on it!

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/batch/V2ExportController.kt
```diff
@@ -46,39 +48,86 @@ class V2ExportController(
...
+        languageService
+          .getLanguagesForExport(params.languages, projectHolder.project.id, authenticationFacade.authenticatedUser.id)
+          .toList()
+          .map { language -> language.tag }
+          .toSet()
+      val exported = exportService.export(projectHolder.project.id, params)
+      checkExportNotEmpty(exported)
+      val responseEntity = getExportResponse(params, exported)
+      headersBuilder.headers(responseEntity.headers)
+      responseEntity.body
```
Anty0: It's somewhat messy that we create a response entity just to extract headers and body from it here. It ignores the status code, for example, which could be confusing.

backend/api/src/main/kotlin/io/tolgee/component/ProjectLastModifiedManager.kt
```diff
@@ -0,0 +1,74 @@
...
+   *    (which translates to HTTP 304 Not Modified / or HTTP 412 when the method is POST)
+   * 4. If data has changed, executing the provided function and wrapping the result in a ResponseEntity
+   *    with appropriate cache control headers
+   *
+   * The response includes:
+   * - Last-Modified header set to the project's modification timestamp
+   * - Cache-Control header set to max-age=0 to ensure validation on each request
+   *
+   * This mechanism helps optimize performance by preventing export data computation and loading from the database when
+   * not modified.
```
Anty0: I'm not opposed to including a bit more comments, but the last paragraph feels like complete filler.

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
```diff
@@ -0,0 +1,132 @@
...
+            .onFailure {
+              logger.debug(
+                "PAT authentication failed",
+                it,
+              )
+            }.getOrNull()
+
+        else ->
+          runCatching { pakAuth(xApiKey) }
+            .onFailure { logger.debug("PAT authentication failed", it) }
```
Anty0:
```suggestion
            .onFailure { logger.debug("PAK authentication failed", it) }
```

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
```diff
@@ -0,0 +1,132 @@
...
+    val prefix = "Bearer "
+    return if (value.startsWith(prefix, ignoreCase = true)) value.substring(prefix.length).trim() else null
+  }
+
+  private fun pakAuth(key: String): TolgeeAuthentication {
+    val parsed = apiKeyService.parseApiKey(key) ?: throw AuthenticationException(Message.INVALID_PROJECT_API_KEY)
+    val hash = apiKeyService.hashKey(parsed)
+    val pak = apiKeyService.findDto(hash) ?: throw AuthenticationException(Message.INVALID_PROJECT_API_KEY)
+
+    if (pak.expiresAt?.before(java.util.Date()) == true) {
```
Anty0: Shouldn't this use `CurrentDateProvider`?

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
```diff
@@ -0,0 +1,132 @@
...
+      actingAsUserAccount = null,
+      isReadOnly = false,
+    )
+  }
+
+  private fun patAuth(key: String): TolgeeAuthentication {
+    val hash = patService.hashToken(key.substring(PAT_PREFIX.length))
+    val pat = patService.findDto(hash) ?: throw AuthenticationException(Message.INVALID_PAT)
+
+    if (pat.expiresAt?.before(java.util.Date()) == true) {
```
Anty0: Shouldn't this use `CurrentDateProvider`?

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
```diff
@@ -0,0 +1,132 @@
...
+   * Supports:
+   * - Authorization: Bearer <jwt>
+   * - X-API-Key: tgpat_<token> (PAT) or tgpak_<...> (PAK, incl. legacy/raw)
+   * - jwtToken: <jwt> (legacy header)
+   */
+  fun resolve(
+    authorizationHeader: String?,
+    xApiKeyHeader: String?,
+    legacyJwtHeader: String?,
+  ): TolgeeAuthentication? {
```
Anty0: Is it a bad idea to pass `StompHeaderAccessor` instead and extract the headers here? I would do it this way, since this function is specific to WebSocket, and it would bring the header names here.

backend/api/src/main/kotlin/io/tolgee/websocket/WebSocketConfig.kt
```diff
@@ -77,19 +78,30 @@ class WebSocketConfig(
...
     }
 
+    val creds = authentication.credentials
+    if (creds is ApiKeyDto) {
+      val matchesProject = creds.projectId == projectId
+      val hasScope = creds.scopes.contains(Scope.KEYS_VIEW)
+      if (!matchesProject || !hasScope) {
+        throw MessagingException("Forbidden")
+      }
+      return
```
Anty0: Shouldn't we still verify users' permissions? I believe these API key scopes are not filtered based on what the user has access to. If that's the case, when user scopes change, the user might no longer have enough scopes to view the project, even though the API key says otherwise.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/batch/V2ExportController.kt
JanCizmar: Makes sense, fixed.

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
JanCizmar: Good point. Thx.

backend/api/src/main/kotlin/io/tolgee/websocket/WebSocketConfig.kt
JanCizmar: Very good catch! ❤️

JanCizmar:
Thanks for the review @Anty0 . I have added support fro eTag and updated some stuff in the security service. Can you please look into that too?

backend/api/src/main/kotlin/io/tolgee/websocket/WebSocketConfig.kt
```diff
@@ -82,22 +81,22 @@ class WebSocketConfig(
...
-    } catch (e: Exception) {
-      throw MessagingException("Forbidden")
+      securityService.checkProjectPermission(
+        projectId = projectId,
+        requiredPermission = Scope.KEYS_VIEW,
+        user = user,
+        apiKey = apiKey,
+      )
+    } catch (e: PermissionException) {
+      logger().debug("API key does not have required scopes", e)
```
Anty0: This could be a bit misleading when `apiKey` is `null`.

Anty0 (approved):
There are some failing tests, otherwise lgtm ^^

PR #3226, opened 2025-08-28
chore: update @sentry/browser and posthog-js dependencies to latest versions

dzienisz:


dzienisz:
@stepan662 I have a lot of issues with the repo package versions; let's have a call to handle this.

Issue #3175, opened 2025-07-16
Improve importing with namespaces


JanCizmar:
### 1. Problem
When users import many namespaces using the UI
- it does not preselect the namespaces when the format is different to `{namespace}/{languageTag}.{extension}`, so the user has to select it manually, which takes time (https://github.com/tolgee/tolgee-platform/issues/3172)
- it's slow

**Bussines impact**
The import is usually the first thing users interact with. We don't want to lose them, so import UI is extremely important.

---

### 2. Appetite
_How much time do we want to spend on this?_

**3 days**

---

### 3. Solution
_How do we imagine solving this problem?_
- Automatically preselect the namespaces according to the file structure (by the folder name or according to the convention for the imported format)
  - You will need to somehow detect what structure it is 
- Use profiler to inspect what's making importing with namespaces so slow. There will be probably some n+1 problem

PR #3232, opened 2025-09-05
feat(import): add pagination to ImportResult table for better perform…

melmathari:
##Summary

Fixes performance of #3175 – Add pagination to ImportResult

https://www.loom.com/share/0aad1864952547f8bc6ba259fc3c834a

Added pagination (20 rows per page) to the import results table.

Prevents rendering all rows at once, improving performance for large imports.

Verified that with correct i18next structure, Tolgee preselects namespaces and languages.

Testing

✅ Tested with ~50 namespaces → works, but still feels slow/laggy.

✅ Checked namespace-related issues but could not reproduce.

✅ Used test files will be added in a comment.

Screenshots
ImportResult Pagination Namespaces Import

## Summary by CodeRabbit

* **New Features**
  * Client-side pagination for import results (20 rows per page) with right-aligned page selector and navigation controls.
  * Pagination shown only when multiple pages exist; page selection updates displayed rows and stays in sync with data changes.

* **Bug Fixes**
  * Empty result sets treated as no data, showing a clear empty state and avoiding invalid page views.

### DIFF
--- a/webapp/src/views/projects/import/component/ImportResult.tsx
+++ b/webapp/src/views/projects/import/component/ImportResult.tsx
@@ -73,7 +73,7 @@ export const ImportResult: FunctionComponent<ImportResultProps> = (props) => {
     setCurrentPage(page);
   };
 
-  // Keep previous behavior: render nothing when there are no rows
+  // Render nothing when there are no rows
   if (totalItems === 0) {
     return null;
   }


melmathari:
## Test files

[locales-500.zip](https://github.com/user-attachments/files/22169865/locales-500.zip)
[locales-nons.zip](https://github.com/user-attachments/files/22169868/locales-nons.zip)
[locales-300.zip](https://github.com/user-attachments/files/22169869/locales-300.zip)
[locales-big.zip](https://github.com/user-attachments/files/22169870/locales-big.zip)
[i18next-flat.zip](https://github.com/user-attachments/files/22169871/i18next-flat.zip)
[locales.zip](https://github.com/user-attachments/files/22169872/locales.zip)

melmathari:
@coderabbit review this

melmathari:
@JanCizmar Open to any feedback. The latest changes seem to fully resolve the performance issues. Previously my machine was crashing, but now I’m able to import 10,000 records without any problems.

I also wasn’t able to reproduce the namespace issues and have included test files for verification.

melmathari:
@CodeRabbit review this please. I've addressed the Early return before memo, you were spot on!

melmathari:
@CodeRabbit review this please. I've addressed the Early return before memo, you were spot on!

Anty0:
@melmathari Hi! Thanks for the PR! I reviewed it, but I'm not sure it addresses the problems mentioned in the issue. The "it's slow" comment mainly relates to "how long it takes to import a file" — importing files takes longer than it should. The front end is already fast enough.

Regarding namespace detection, it works well when the namespace name matches the first directory in the structure, but it fails if it's located elsewhere. There might be a better way to detect them. We should also make it easier for users to fix detected namespaces — currently, they have to go through each file and set the namespace, which isn't ideal. If you have any suggestions for fixing this, ask before implementing them. ^^

melmathari:
@Anty0 Makes sense, I’m following you. When I tested it, the upload process was quite smooth and eventually timed out because of the file size trying to load the table. Adding pagination makes the experience much smoother. I’m open to suggestions, but without the changes from my last PR my entire Tolgee instance freezes when trying to load more than 1,000 items.

As for the namespaces, i'll re-test!

Thanks for reviewing this.

Anty0:
@melmathari I'm not sure if there are actually people importing over 200 translation files. Most apps don't need more than 20 languages. To reach 1000 files, you'd need 50 namespaces with 20 languages each.

If we were to implement pagination for that list, it would have to be done on the server side, and the frontend would use the `PaginatedHateoasList` component, which we use for all other paginated views. However, I still don't see the point in implementing it. It would only add more code to maintain without providing real benefits for normal use cases.

melmathari:
@Anty0 Alright, makes sense and appreciate you taking the time. I'm going to close this and will create a new PR for addressing the core issue. Apologies, didn't realize that this would not be a normal use case as i'm new to Tolgee.

Anty0:
@melmathari Sure! Thank you for taking the time to create a PR. Looking forward to seeing your next one ^^

PR #3233, opened 2025-09-05
fix ConcurrentModificationException in V2ExportAllFormatsTest

bdshadow:
It looks like there is a bug in spring security which causes ConcurrentModificationException when using MockMvc.
There are multiple reports and some of them are pretty old. Here are the most relevant ones:
https://github.com/spring-projects/spring-framework/issues/31543
https://github.com/spring-projects/spring-security/issues/9175
It doesn't seem to be fixed soon. The best explanation of what is going on is here by Rossen Stoyanchev:
https://github.com/spring-projects/spring-security/issues/11452#issuecomment-1172491187
So replacing MockMvc with TestRestTemplate for this test, until these spring bugs are fixed.


bdshadow:
this one weirdly failed with `ResourceAccessException`, which i've never had locally. Anyway, it's better (in my opinion), than ConcurrentModificationException, as this is caused by some timeout or network inconsistency.  Please, run the pipeline again. I'll invetigate it, if it continues failing

Anty0:
Same test (Flutter) both times 🤔

bdshadow:
pushed it with a `ByteArrayResource` option. Maybe it will help. There is such difference between local run and github actions...

bdshadow:
replicated it in the workflow in my fork of the repo: https://github.com/bdshadow/tolgee-platform/actions/runs/17519346008/job/49761163077 But it's very unpredictable. Though i still think it's better than the ConcurrentModificationExcpeption. 

I will investigate it further with builds in my fork

bdshadow:
Replicated locally. And it failed on the "Cache-Control" header. Check the screenshot, you'll see that it was not set correctly.
Screenshot of the normal execution.
Screenshot 2025-09-07 at 0 20 51

Image of the failed execution:
Screenshot 2025-09-07 at 0 37 56

And it means that we're back to the same bug in spring. Just the same links i used in the code :)
https://github.com/spring-projects/spring-security/issues/9175#issuecomment-2017714531
and that comment from Rossen: https://github.com/spring-projects/spring-security/issues/11452#issuecomment-1172491187

So it's not a bug in the test, it's a bug in the way spring processes `StreamingResponseBody` response. For example, here it is summarized: https://github.com/spring-projects/spring-security/issues/9175#issuecomment-2096978502. And it's not good news, since users of tolgee can also experience this bug (not often, but it can happen). So this endpoint must be rewritten without `StreamingResponseBody` in the `ResponseEntity`

bdshadow:
@JanCizmar please, don't merge by now. It will just replace one exception with another one. The problem of both is the same. The controller mus be rewritten

bdshadow:
@Anty0 @JanCizmar I need your advice here. I'll try to summarize it shortly.

The problem is with all async requests. To reproduce it, you just need to repeat it many time in a loop or use the `@RepeatedTest`. So it's not only about the `V2ExportAllFormatsTest`. You can, for example, take the ``TranslationSuggestionControllerStreamingTest.`it works`()`` test and replace `@Test` annotation with `@RepeatedTest(100)` - then 2-3 out of 100 will fail with the `ConcurrentModificationException`.

So `V2ExportAllFormatsTest` is just "lucky" to catch this problem sometimes because of the `@ParameterizedTest` and many values in the `@ExportFormat` enum, or you can try the following test:
```kotlin
  @RepeatedTest(100)
  @ProjectJWTAuthTestMethod
  fun `check for ConcurrentModificationException - spring bug`() {
    // This test is to ensure that no ConcurrentModificationException is thrown during export
    val mvcResult = performProjectAuthGet("export?format=${ExportFormat.JSON}")
      .andIsOk
      .andDo { obj: MvcResult -> obj.asyncResult }
      .andReturn()

    val responseContent = mvcResult.response.contentAsByteArray
    assertThat(responseContent).isNotEmpty()
  }
```

It is a spring bug: https://github.com/spring-projects/spring-security/issues/9175. It exists for a very long time. I tried on of the described workarounds: `headerWriterFilter.setShouldWriteHeadersEagerly(true)`, but it doesn't allow to work with headers conveniently (for example, in the `ImageStorageController` when setting the `Cache-Control` header).

So here are 2 options:
1. Wait till the bug is fixed, knowing that sometimes users can experience this problem. It's also irritating in the pipelines - this can be dirty-fixed with try-catch or with the [junit assumptions](https://docs.junit.org/5.12.1/api/org.junit.jupiter.api/org/junit/jupiter/api/Assumptions.html)
2. Rewrite these async responses from returning `StreamingResponseBody` to `ByteArray`, like in the `ImageStorageController`

Since i'm not yet well aware of the usage of these endpoints and how critical they are, i need your opinion on this.

JanCizmar:
@bdshadow None of the proposed solutions sound optimal to me. We cannot get rid of the streaming response body, because we really need it in some cases for streaming of `json-nd` files in multiple cases.

However, can we as a workaround use the try-catch retries to handle that with some proper comment why is this happening?

Please tell me if I am missing something. I am currently super busy and I don't remember the whole context.

The goal is basically to have the functionality fixed and in the same time don't be annoying in the pipeline.

Anty0:
@bdshadow Is the issue only related to the mocked requests?

Looking at our Sentry dashboard, I see only a few `ConcurrentModificationExceptions` events caught, all from `TagService` (our own service), and they are unrelated to this issue. I recall you mentioned it could affect users outside the test environment. I'm checking if we still believe this is true. The chance of this happening in production must be very low, as I haven't seen any reports like this in Sentry.

bdshadow:
@Anty0 it took me some time to test and reply to you.

Yes, this can happen not only on test env. The chance is very-very low. You can try to launch tolgee locally and start bombing it with async requests: 
```java
suspend fun main() {
    for (i in 1..10000) {
        reactor.netty.http.client.HttpClient.create().baseUrl("http://localhost:8080")
            .headers { h -> h.add("Authorization", "Bearer <JWT_TOKEN_HERE>") }
            .get()
            .uri("/v2/projects/1/export?format=FLUTTER_ARB")
            .responseSingle { resp, bytes ->
                print("Response status ${i}: ${resp.status().code()}\n")
                bytes
            }
            .block()
        delay(200)
    }
}
```
Usually within 2000 iterations you will have an exception in the tolgee logs:
Screenshot 2025-09-12 at 11 39 50

And it is not always a `ConcurrentModificationExceptions`. You can check the spring tickets i mentioned above, and somtimes it's a `NullPointerException` inside jetty or `IOException` with `Cannot invoke "org.apache.tomcat.util.http.MimeHeaderField.getName()"`, which is more closer to the screenshot. It is always about dealing with headers.

You can try to check sentry for such problems. But as i said, the probability is very-very low, yet it exists. In my opinion, it's enough to just fix in tests to not fail in case of `ConcurrentModificationExceptions`. I think this thing doesn't worth time to workaround properly. But it's not my call here :)

Anty0:
@bdshadow Thank you for all the testing! I agree; let's just add a retry to the affected tests. We should try to make sure other `ConcurrentModificationExceptions` won't get ignored—there is a chance we might hit those during export tests, and it would be good if we could still detect these while ignoring those triggered by the bug.

bdshadow:
@Anty0 as you said, i tried the best to exclude exactly that `ConcurrentModificationException`.
Also I updated it the way that the test is ignored, when this exception is caught. In my opinion repeated request isn't good. Described it in the comments. I think this on can be merged now

Anty0 (approved):
Good; I'd prefer to retry it ten times. Hopefully, the issue won't occur that many times, and we won't skip any tests that way, but let's go ahead. We'll merge it like this.

PR #3241, opened 2025-09-09
chore: upgrade Sentry packages to v10.11.0 and related dependencies

dzienisz:
## Summary by CodeRabbit

* **Improvements**
  * Enhanced error monitoring with performance tracing and session replay; reduced replay sampling and include user identifiers when available.

* **Removals**
  * Stopped collecting Web Vitals performance metrics; removed the related performance utility.

* **Documentation**
  * Minor formatting fix in the Unreal integration guide for improved readability.

* **Chores**
  * Updated monitoring-related dependencies.


webapp/src/component/MandatoryDataProvider.tsx
```diff
@@ -36,11 +36,18 @@ export const MandatoryDataProvider = (props: any) => {
...
+          Sentry.browserTracingIntegration(),
+          Sentry.replayIntegration({
+            maskAllText: false,
+            blockAllMedia: false,
+          }),
+        ],
+        tracesSampleRate: 1.0,
+        replaysSessionSampleRate: 0.1,
         replaysOnErrorSampleRate: 1.0,
+        sendDefaultPii: true,
```
dkrizan: I am sure this would really help us debug the issues, but I would recommend we discuss if it’s compliant with our TOS and legal/privacy obligations before enabling it.

dzienisz: ok, let's remove it for now

webapp/src/index.tsx
```diff
@@ -110,8 +109,3 @@ ReactDOM.render(
   </StrictMode>,
   document.getElementById('root')
 );
-
-// If you want to start measuring performance in your app, pass a function
-// to log results (for example: reportWebVitals(console.log))
-// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
-reportWebVitals();
```
dkrizan, line 117: Just a quick question - do we really want to do this? I don't have any background information on it, so I'm just asking.

dzienisz: It's not working it's empty function that do nothing. Just a cleanup. We have sentry and postbox it's enough.

dkrizan: I guess it's like an interface for developers who want to measure performance. You can pass in `console.log` or anything similar. But by default, its not enabled. But I'm not familiar with how to use it, that's what I just read in the documentation.

https://create-react-app.dev/docs/measuring-performance/

dzienisz: I understand, but we don't need it. It's for special reasons, but we don't have any use case for it right now. Sentry collects this data by default. It's why I made this change in the first place. The current implementation is not correct. We should use the React package, not browser (for Vanilla JS)

dzienisz:
@dkrizan can we somehow fix those flaky backend tests?

webapp/src/index.tsx
dkrizan: Alright, got it 👍

dkrizan:
> @dkrizan can we somehow fix those flaky backend tests?

that one failing is tricky, it is hard to reproduce it locally .. I know how to modifty the test so it won't fail, but I am afraid it might affect the expected result.

dzienisz:
@dkrizan @JanCizmar I think you are authorized to merge this branch.

Issue #3218, opened 2025-08-22
Supporter user role - read-only admin role


Anty0:
**Introduce a new role SUPPORTER that:**
- Can view everything globally like ADMIN.
- Cannot perform modifying actions on data that is not their own (read-only on others’ resources).
- Can use the platform normally for their own data (profile, their projects/orgs, etc.).
- Can impersonate any user, with a read-only impersonation token that guarantees no changes can be made during impersonation.

---

### High-level Approach

- Add SUPPORTER as a user role alongside USER and ADMIN.
- Introduce a "read-only mode" flag in the security context that can be triggered either because:
  - The authenticated platform role is SUPPORTER and is acting on resources they don’t own, or
  - The session is an impersonation session initiated with a read-only token.
- Enforce read-only:
  - API layer (HTTP method guards + explicit annotations for safe POST endpoints).
  - Method-level security (service-level guard in case endpoints circumvent HTTP verb semantics).
- Provide an endpoint to generate read-only impersonation tokens that encode “read-only” in claims; ensure the authentication pipeline respects this and blocks mutation.
- *Update UI to reflect read-only mode and disable modifying controls.

### Acceptance Criteria

- A user with role SUPPORTER:
  - Can retrieve any resource (GET/HEAD/OPTIONS) across the platform.
  - Cannot create/update/delete resources they do not own; receives 403 with code `read_only_mode`.
  - Can perform full CRUD on resources they own (their profile, their organizations/projects).
- Read-only impersonation token:
  - Can be issued by ADMIN and SUPPORTER.
  - If issued by SUPPORTER, the token is always read-only; mutation is blocked even on the impersonated user’s own account.
- Frontend:
  - Read-only banner visible during read-only impersonation.
  - *Mutating controls are disabled.
- Tests:
  - Unit, integration, and E2E cover the above and pass in CI.
- Documentation:
  - OpenAPI reflects read-only rules.
  - Docs reflect the new SUPPORTER role.

PR #3248, opened 2025-09-18
feat: supporter user role

Anty0:
Implements #3218

Known limitations:
- When impersonating in read-only mode, the frontend does not respect the read-only mode. (UI navigation works fine, but the frontend allows the user to trigger write operations, which only fail on the server side afterward.)
- When accessing a project or organization as a supporter, I rely on a server reporting the user has read-only permissions to the frontend. However, the actual check doesn't consider permissions but instead looks at the HTTP method—this could lead to discrepancies between the UI and actual API access. The issue isn't as pronounced, though. Also, we may be able to mitigate the first issue by doing something similar for read-only authenticated requests—only returning a subset of permissions the user really has.

## Summary by CodeRabbit

* **New Features**
  * Added "Supporter" server role with read-only access and a consolidated "all.view" permission scope.
  * Initial app load now includes auth info (read-only) so the UI adapts messaging and behavior.

* **Improvements**
  * Administration, Billing, organization and project UIs now treat Supporters as read-only admins where appropriate.
  * Read-only mode blocks write actions and updates announcements/messages.

* **API Changes**
  * Invitation accept now supports PUT; endpoint for removing direct project permissions renamed.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/V2UserControllerTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/V2UserControllerTest.kt
@@ -246,7 +246,7 @@ class V2UserControllerTest : AuthorizedControllerTest() {
     ).andIsOk.andAssertThatJson {
       node("accessToken").isString.satisfies { token: String ->
         val authentication = jwtService.validateToken(token)
-        authentication.isSuperToken
+        assertThat(authentication.isSuperToken).isTrue
       }
     }
   }
@@ -264,7 +264,7 @@ class V2UserControllerTest : AuthorizedControllerTest() {
     ).andIsOk.andAssertThatJson {
       node("accessToken").isString.satisfies { token: String ->
         val authentication = jwtService.validateToken(token)
-        authentication.isSuperToken
+        assertThat(authentication.isSuperToken).isTrue
       }
     }
   }


backend/app/src/main/kotlin/io/tolgee/configuration/WebSecurityConfig.kt
```diff
@@ -94,7 +100,7 @@ class WebSecurityConfig(
           },
         )
         it.requestMatchers(*PUBLIC_ENDPOINTS).permitAll()
-        it.requestMatchers("/v2/administration/**", "/v2/ee-license/**").hasRole("ADMIN")
+        it.requestMatchers(*ADMIN_ENDPOINTS).hasRole("SUPPORTER")
         it.requestMatchers("/api/**", "/v2/**").authenticated()
```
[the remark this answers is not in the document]
Anty0: What is this piece of code in `TolgeeAuthentication` doing then? Wouldn't it happen that it also grants the SUPPORTER role to ADMIN? :3

``` kotlin
  override fun getAuthorities(): Collection<GrantedAuthority> {
    return when (userAccount.role) {
      UserAccount.Role.USER ->
        listOf(
          SimpleGrantedAuthority(ROLE_USER),
        )
      UserAccount.Role.SUPPORTER ->
        listOf(
          SimpleGrantedAuthority(ROLE_USER),
          SimpleGrantedAuthority(ROLE_SUPPORTER),
        )
      UserAccount.Role.ADMIN ->
        listOf(
          SimpleGrantedAuthority(ROLE_USER),
          SimpleGrantedAuthority(ROLE_SUPPORTER),
          SimpleGrantedAuthority(ROLE_ADMIN),
        )
      null -> emptyList()
    } + authorityFromIsReadOnly
  }
```

Or is there a reason not to do this?

backend/data/src/main/kotlin/io/tolgee/constants/Message.kt
```diff
@@ -307,8 +307,10 @@ enum class Message {
   SUGGESTION_CANT_BE_PLURAL,
   SUGGESTION_MUST_BE_PLURAL,
   DUPLICATE_SUGGESTION,
-
   UNSUPPORTED_MEDIA_TYPE,
```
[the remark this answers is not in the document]
Anty0: The change removes an empty line above the `UNSUPPORTED_MEDIA_TYPE`. The `UNSUPPORTED_MEDIA_TYPE` is left unchanged.

backend/data/src/main/kotlin/io/tolgee/model/enums/Scope.kt
```diff
@@ -90,19 +101,19 @@ enum class Scope(
...
+              promptsView,
+              HierarchyItem(PROJECT_EDIT),
+              HierarchyItem(LANGUAGES_EDIT),
+              translationsView,
+              screenshotsView
+            )
           ),
-          HierarchyItem(ACTIVITY_VIEW),
+          activityView,
           HierarchyItem(LANGUAGES_EDIT, listOf(HierarchyItem(PROMPTS_VIEW))),
```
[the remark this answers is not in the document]
Anty0: It was already like this. Yes, it is intentional—you can't modify prompts unless you have project edit permission. The user can clearly see in the UI that granting the permission gives them project edit permission, too.

backend/data/src/main/kotlin/io/tolgee/security/authentication/JwtService.kt
```diff
@@ -80,6 +97,17 @@ class JwtService(
...
+  fun emitTokenRefreshForCurrentUser(
+    isSuper: Boolean? = false,
+  ): String {
+    return emitToken(
+      userAccountId = authenticationFacade.authenticatedUser.id,
+      actingAsUserAccountId = authenticationFacade.actingUser?.id,
+      isReadOnly = authenticationFacade.isReadOnly,
+      isSuper = isSuper ?: authenticationFacade.isUserSuperAuthenticated,
+    )
+  }
```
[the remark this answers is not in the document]
Anty0: That's the idea. It's much safer not to propagate it by default. Still, I should document this quirk.

backend/data/src/main/kotlin/io/tolgee/security/authentication/JwtService.kt
```diff
@@ -133,13 +161,29 @@ class JwtService(
...
     val steClaim = jws.body[SUPER_JWT_TOKEN_EXPIRATION_CLAIM] as? Long
     val hasSuperPowers = steClaim != null && steClaim > currentDateProvider.date.time
 
+    val roClaim = jws.body[JWT_TOKEN_READ_ONLY_CLAIM] as? Boolean ?: false
+
+    if (roClaim && account.isAdmin()) {
+      // we don't allow read-only admin impersonation to make our lives easier
+      throw AuthenticationException(Message.INVALID_JWT_TOKEN)
+    }
+
```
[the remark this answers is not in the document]
Anty0: Why would I check if the actor is an admin? I don't want to allow the impersonation of an admin using a read-only token. I'll improve the comment about the intentions of this check.

At this point, it doesn't matter who is impersonating. (That was already checked before the token was created, though there might be an argument we should check to "revoke" the token if the actor is no longer an admin or supporter. I'll add that, too.)

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationService.kt
```diff
@@ -153,7 +154,7 @@ class OrganizationService(
         tolgeeProperties.authentication.userCanCreateOrganizations &&
           userAccount.thirdPartyAuthType !== ThirdPartyAuthType.SSO
 
-      if (canCreateOrganizations || userAccount.role == UserAccount.Role.ADMIN) {
+      if (canCreateOrganizations || userAccount.isAdmin()) {
```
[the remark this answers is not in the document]
Anty0: Yes, the supporter role is read-only - they can't create organizations in the name of other users.

webapp/src/component/layout/TopBar/announcements/DebuggingCustomerAccountAnnouncement.tsx
```diff
@@ -15,11 +18,19 @@ export const DebuggingCustomerAccountAnnouncement: FC = () => {
...
+  ) : (
+    <T keyName="administration-debugging-customer-account-message" />
+  );
+
   return (
     <TopBarAnnouncementWithAlertIcon data-cy="administration-debug-customer-account-message">
-      <Box>
-        <T keyName="administration-debugging-customer-account-message" />
-      </Box>
+      <Box>{message}</Box>
```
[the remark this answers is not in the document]
Anty0: How many times do we need to say that Tolgee is using a CI job to update the translation keys before the release? There is no need to update them manually. Funni rabbit doing funni mistakes :3

backend/data/src/main/kotlin/io/tolgee/security/authentication/JwtService.kt
```diff
@@ -133,13 +161,29 @@ class JwtService(
...
     }
 
+    val actor = validateActor(jws.body)
+
+    // to avoid mass sign-out on update, we allow tokens without a device id
+    // maybe we should consider changing this later on
+    val deviceId = jws.body[JWT_TOKEN_DEVICE_ID_CLAIM] as? String
+
     val steClaim = jws.body[SUPER_JWT_TOKEN_EXPIRATION_CLAIM] as? Long
     val hasSuperPowers = steClaim != null && steClaim > currentDateProvider.date.time
```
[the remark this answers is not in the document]
Anty0: This is a correct observation, but the code has been like this for a long time, so I assume the Date gets automatically converted to a number when stored in Jwt token.

backend/security/src/test/kotlin/io/tolgee/security/authentication/ReadOnlyModeInterceptorTest.kt
```diff
@@ -0,0 +1,201 @@
...
+  private val authenticationFacade = Mockito.mock(AuthenticationFacade::class.java)
+
+  private val authentication = Mockito.mock(TolgeeAuthentication::class.java)
+
+  private val interceptor = ReadOnlyModeInterceptor(authenticationFacade)
+
+  private val mockMvc =
+    MockMvcBuilders.standaloneSetup(TestController::class.java)
+      .addInterceptors(interceptor)
+      .build()
```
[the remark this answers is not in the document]
Anty0: What handler invocation do you mean?

backend/security/src/main/kotlin/io/tolgee/security/authentication/ReadOnlyModeInterceptor.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    if (handler.isReadOnly(request.method)) {
+      // These methods should be read-only - safe to call from read-only mode
+      return true
+    }
+
+    logger.debug(
+      "Rejecting access for user#{} - Write operation is not allowed in read-only mode",
+      authenticationFacade.authenticatedUser.id,
+    )
```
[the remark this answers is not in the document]
Anty0: That is only reachable when `isAuthenticated` is true, at which point the `authenticatedUser` is guaranteed to be available.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
```diff
@@ -115,51 +116,56 @@ class OrganizationRoleService(
...
+    this.checkUserIsOwner(authenticationFacade.authenticatedUser.id, organizationId, isReadOnlyAccess)
   }
 
-  fun checkUserIsOwnerOrMaintainer(organizationId: Long) {
-    this.checkUserIsOwnerOrMaintainer(authenticationFacade.authenticatedUser.id, organizationId)
+  fun checkUserIsOwnerOrMaintainer(organizationId: Long, isReadOnlyAccess: Boolean) {
+    this.checkUserIsOwnerOrMaintainer(authenticationFacade.authenticatedUser.id, organizationId, isReadOnlyAccess)
   }
 
   fun checkUserIsMember(
```
JanCizmar: This method seems to be unused.

backend/data/src/main/kotlin/io/tolgee/dtos/cacheable/UserAccountDto.kt
```diff
@@ -43,3 +43,19 @@ data class UserAccountDto(
...
+
+fun UserAccountDto.isSupporter(): Boolean {
+  return role == UserAccount.Role.SUPPORTER
+}
+
+fun UserAccountDto.isSupporterOrAdmin(): Boolean {
+  return role == UserAccount.Role.SUPPORTER || role == UserAccount.Role.ADMIN
+}
+
+fun UserAccountDto.hasAdminAccess(isReadonlyAccess: Boolean): Boolean {
```
JanCizmar: Here we have a method by which name I would guess we are specifically checking for admin access. I would not guess that we are checking also supporter role. 

What I would suggest here is not to deal with supporer in this method, but create another one called hasSupporterRole. Then I would refactor the rest of the code to be clear in this. I believe that endless passing of `isReadonlyAccess` is clean. It would just make everyone asking "Why the hell is it here and what does it mean".

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/AdministrationController.kt
```diff
@@ -147,7 +149,24 @@ class AdministrationController(
...
+    if (isAlreadyImpersonating) {
+      // We don't want to recreate the Inception movie here
+      throw BadRequestException(Message.ALREADY_IMPERSONATING_USER)
+    }
+
+    val actingUser = authenticationFacade.authenticatedUser
     val user = userAccountService.get(userId)
-    return jwtService.emitToken(user.id, true)
+    val isAdmin = actingUser.isAdmin()
+    if (user.isAdmin() > actingUser.isAdmin()) {
```
JanCizmar: This condition is not readable to me is this equivalent to `user.isAdmin() && !actingUser.isAdmin()`?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/AdministrationController.kt
```diff
@@ -147,7 +149,24 @@ class AdministrationController(
...
+      // We don't allow impersonation of admin by supporters
+      throw BadRequestException(Message.IMPERSONATION_OF_ADMIN_BY_SUPPORTER_NOT_ALLOWED)
+    }
+    return jwtService.emitToken(
+      userAccountId = user.id,
+      actingAsUserAccountId = actingUser.id,
+      isReadOnly = !isAdmin,
```
JanCizmar: Here the passing of `isReadOnly` seems hacky and not readable. I don't know what's happening on this side.

backend/data/src/main/kotlin/io/tolgee/service/security/SecurityService.kt
```diff
@@ -544,12 +551,8 @@ class SecurityService(
     }
   }
 
-  private fun isCurrentUserServerAdmin(): Boolean {
-    return isUserAdmin(activeUser)
-  }
-
-  private fun isUserAdmin(user: UserAccountDto): Boolean {
-    return user.role == UserAccount.Role.ADMIN
+  private fun hasCurrentUserServerAdminAccess(isReadonlyAccess: Boolean): Boolean {
```
JanCizmar: Instead of passing the boolean I would create a separate method `hasCurrentUserServerSupporterAccess`. I would also add documentation comments for both methods.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/AdministrationController.kt
JanCizmar: Reading it again I found out that only case when this can happen is when the user is supporter right? Anyway. I believe right aproach would be not using the boolean, but creating amehtod in jwtService called `emitSupporterToken` and use it when `actingUser.role == SUPPORTER`

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/AdministrationController.kt
```diff
@@ -147,7 +149,24 @@ class AdministrationController(
...
+    val isAlreadyImpersonating = authenticationFacade.actingUser != null
+    if (isAlreadyImpersonating) {
+      // We don't want to recreate the Inception movie here
+      throw BadRequestException(Message.ALREADY_IMPERSONATING_USER)
+    }
+
+    val actingUser = authenticationFacade.authenticatedUser
     val user = userAccountService.get(userId)
-    return jwtService.emitToken(user.id, true)
+    val isAdmin = actingUser.isAdmin()
```
JanCizmar: This is confusing, harder to read. I would not store it in variable if we are using it so far from the usage and between we are using other isAdmin functions so often.

backend/data/src/main/kotlin/io/tolgee/service/security/SecurityService.kt
Anty0: I've just switched to using `isAdmin` and `isAdminOrSupporter` directly on `activeUser`.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationController.kt
```diff
@@ -156,7 +155,7 @@ class AdministrationController(
     }
 
     val actingUser = authenticationFacade.authenticatedUser
-    val user = userAccountService.get(userId)
```
Anty0, line 151: @JanCizmar Is there a reason to use the full `UserAccount` entity instead of just `UserAccountDto`?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2InvitationController.kt
```diff
@@ -53,6 +54,7 @@ class V2InvitationController(
   private val publicInvitationModelAssembler: PublicInvitationModelAssembler,
 ) {
   @GetMapping("/v2/invitations/{code}/accept")
+  @WriteOperation
```
JanCizmar, line 57: Oh crap! Why is the method GET... 🤦 My bad...

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2UserController.kt
```diff
@@ -117,7 +117,7 @@ class V2UserController(
   ): JwtAuthenticationResponse {
     userAccountService.updatePassword(authenticationFacade.authenticatedUserEntity, dto!!)
     return JwtAuthenticationResponse(
-      jwtService.emitToken(authenticationFacade.authenticatedUser.id, true),
+      jwtService.emitTokenRefreshForCurrentUser(isSuper = true),
```
JanCizmar: Do we need to emit superToken here?

backend/api/src/main/kotlin/io/tolgee/component/PreferredOrganizationFacade.kt
```diff
@@ -31,4 +31,16 @@ class PreferredOrganizationFacade(
     }
     return null
   }
+
+  private fun getCurrentUserPreferences(): UserPreferences? {
```
JanCizmar: I would add some documentation comment explaining what happens here.

backend/data/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFacade.kt
```diff
@@ -79,9 +79,18 @@ class AuthenticationFacade(
         return authentication.userAccountView
       }
 
+  // -- ACTING USER
```
JanCizmar, line 82: Please use documentation comments instead of this weirdo comments and explain what is the purpose of each property.

backend/api/src/main/kotlin/io/tolgee/controllers/PublicController.kt
```diff
@@ -102,7 +104,7 @@ class PublicController(
     @PathVariable("code") @NotBlank code: String,
   ): JwtAuthenticationResponse {
     emailVerificationService.verify(userId, code)
-    return JwtAuthenticationResponse(jwtService.emitToken(userId))
+    return JwtAuthenticationResponse(jwtService.emitToken(userId, isSuper = false))
```
JanCizmar: isSuper is false by default. No need to provide it.

backend/data/src/main/kotlin/io/tolgee/dtos/ComputedPermissionDto.kt
```diff
@@ -63,12 +63,24 @@ class ComputedPermissionDto(
     }
   }
 
-  val isAllPermitted = this.expandedScopes.toSet().containsAll(Scope.values().toList())
+  val isAllPermitted = this.expandedScopes.toSet().containsAll(Scope.entries)
 
   fun getAdminPermissions(userRole: UserAccount.Role?): ComputedPermissionDto {
```
JanCizmar: Can we rename this to getAdminOrSupporterPermissions() and maybe add some documentation comment?

backend/data/src/main/kotlin/io/tolgee/security/authentication/JwtService.kt
```diff
@@ -54,16 +58,20 @@ class JwtService(
...
    *
    * @param userAccountId The user account ID this token belongs to.
+   * @param actingAsUserAccountId The user account ID of the actor who initiated impersonation.
+   * @param isReadOnly Whether the token allows read-only access or full read-write access.
    * @param isSuper Whether to emit a super-powered token or not.
    * @return An authentication token.
    */
   fun emitToken(
     userAccountId: Long,
+    actingAsUserAccountId: Long? = null,
```
JanCizmar, line 68: This is just a nitpick, but I want to mention it so we follow it in the future: I think we should not pass booleans in this case. Instead we can refactor this method so we:

1. create a base method that generates the bas builder
2. we add the claims in the other methods

backend/data/src/main/kotlin/io/tolgee/security/authentication/TolgeeAuthentication.kt
```diff
@@ -73,6 +110,9 @@ class TolgeeAuthentication(
 
   companion object {
     const val ROLE_USER = "ROLE_USER"
+    const val ROLE_SUPPORTER = "ROLE_SUPPORTER"
     const val ROLE_ADMIN = "ROLE_ADMIN"
+    const val ROLE_RO = "ROLE_RO"
```
JanCizmar: Maube it would be more clear if we call this `ROLE_READ_ONLY ` and similarly the RW?

backend/development/src/main/kotlin/io/tolgee/controllers/internal/e2eData/SensitiveOperationProtectionE2eDataController.kt
```diff
@@ -33,14 +33,14 @@ class SensitiveOperationProtectionE2eDataController(
 
     val baseline = currentDateProvider.date
     currentDateProvider.forcedDate = Date(baseline.time - authenticationProperties.jwtSuperExpiration - 10_000)
-    val expiredToken = jwtService.emitToken(data.franta.id, true)
+    val expiredToken = jwtService.emitToken(data.franta.id, isSuper = true)
```
JanCizmar, line 29: I would prefer having method "emitSuperToken" without passing the boolean, but it's just a nitpick.

ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/AiPromptCustomizationController.kt
```diff
@@ -81,7 +81,7 @@ class AiPromptCustomizationController(
   }
 
   @GetMapping("projects/{projectId:[0-9]+}/language-ai-prompt-customizations")
-  @Operation(summary = "Sets project level prompt customization")
+  @Operation(summary = "Returns language level prompt customization")
```
JanCizmar, line 84: Nice catch <3

backend/security/src/main/kotlin/io/tolgee/security/authorization/ProjectAuthorizationInterceptor.kt
```diff
@@ -177,4 +183,27 @@ class ProjectAuthorizationInterceptor(
 
     return projectPerms?.scopes
   }
+
+  private val canUseAdminRights
```
JanCizmar: I would name it `canUseAdminPermissions` to keep it consistent. Or did you have some good reason to name it like this?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2InvitationController.kt
Anty0: Should I add the POST version of the endpoint and mark the GET version as deprecated? ^^

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2UserController.kt
Anty0: It was mostly me keeping the functionality the same.

Thinking about it, the user is changing his password, which technically means he has just entered his password, so asking for the password again to get the supertoken seems like an unnecessary step. There is one edge case worth mentioning. If the user is using 2FA, there might be an argument that giving out the supertoken here might give the user elevated permissions that shouldn't be available to the user.

For simplicity, I'll leave out the supertoken here. This seems like a safer way to handle it.

backend/api/src/main/kotlin/io/tolgee/component/PreferredOrganizationFacade.kt
Anty0: Looking at this again, I’m not sure if it’s even necessary. The plan was to avoid creating a preferred organization if the user is being accessed in read-only mode, but I think I'll just let it happen. I don't see much benefit in preventing this.

backend/data/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFacade.kt
Anty0: I'm just following the existing style there (as created by Cynthia). What would you recommend I use to split the file into sections? Or should I remove the sections and add a separate doc string for each property?

backend/data/src/main/kotlin/io/tolgee/security/authentication/JwtService.kt
Anty0: Builder for token. I like this. I might just do it here anyway. It will fix the whole mess of "if you set the `isSuper` of `emitTokenRefreshForCurrentUser` to `null` it will keep the super token same as it is rn" since now I can just make a function for it. I'll return to this one.

ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/controllers/AiPromptCustomizationController.kt
Anty0: 😊

backend/security/src/main/kotlin/io/tolgee/security/authentication/WriteOperation.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import org.springframework.security.access.prepost.PreAuthorize
+
+/**
+ * Marks a component method that requires read-write mode (i.e. current authentication must not be read-only).
+ */
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.ANNOTATION_CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+@PreAuthorize("hasRole('READ_WRITE')")
+annotation class WriteOperation
```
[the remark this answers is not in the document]
Anty0: Not needed. `ReadOnlyInterceptor` already handles this. The `PreAuthorize` is more like an extra, so it can also be used for service methods.

backend/data/src/main/kotlin/io/tolgee/security/authentication/JwtService.kt
Anty0: I'll leave it out for now, but it's a good idea. I'll keep it in mind for the future. ^^

backend/development/src/main/kotlin/io/tolgee/controllers/internal/e2eData/SensitiveOperationProtectionE2eDataController.kt
Anty0: I'll leave it as is for now, but we can remake it as a builder the next time we come across it.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationController.kt
JanCizmar: I don't think so. If it works with dto, use dto. :)

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/V2InvitationController.kt
JanCizmar: I think it should be put. And yes, please deprecate and create new... :)

backend/data/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFacade.kt
JanCizmar: We can just ignore it for now.

PR #3252, opened 2025-09-25
fix: KeyControllerComplexEditTest.`stores big meta` - db commits in parallel transactions happen simultaneously

bdshadow:
Changes summary:
1. don't launch new async threads for just checking for emptiness of arrays
2. add missed tracking of bigMetaDeletingAsyncTimer metric
3. fix KeyControllerComplexEditTest.`stores big meta` - surprisingly, within the test 2 paralallel transactions happen at the same time within 0.001 of a second, that's why the comment "the data is stored in 2 different transactions executed with different timestamps" isn't actually correct with the given precision of storing time. 

The test failure occurred frequently in the pipelines. You can easily replicate it locally too, by replacing the `@Test` with `@RepeatedTest(100)` for the `stores big meta` test. 

I was also thinking about wrapping the `PlatformTransactionManager` to track the created transactions or follow the number of created threads when calling `@Async`. However, all of this complicates the test significantly. Using metrics seems like a sensible solution (please, comment if not)

## Summary by CodeRabbit

- Bug Fixes
  - Prevents unnecessary background processing when there are no items to insert or delete, reducing load and potential no-op executions.
  - Improves reliability of distance updates by gating synchronous and asynchronous operations on actual work.

- Chores
  - Enhanced instrumentation for insert/delete operations with precise timing and counters to improve monitoring.
  - Updated tests to validate metrics-based behavior for asynchronous processing.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/KeyControllerComplexEditTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/KeyControllerComplexEditTest.kt
@@ -1,5 +1,6 @@
 package io.tolgee.api.v2.controllers.v2KeyController
 
+import io.tolgee.Metrics
 import io.tolgee.ProjectAuthControllerTest
 import io.tolgee.constants.Message
 import io.tolgee.development.testDataBuilder.data.KeysTestData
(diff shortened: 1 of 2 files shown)


JanCizmar (approved):

JanCizmar:
Thanks! Awesome fix.

It's a chore btw.

PR #3259, opened 2025-10-01
chore: apply readOnly transactions in batch jobs, where possible

bdshadow:
Postgres does some optimizations in the locks and logs, when dealing with readOnly transactions. Might be helpful in problems, which are currently experienced in the batch jobs.

## Summary by CodeRabbit

* **Refactor**
  * Improved transaction handling to run certain background checks in read-only mode, reducing contention and smoothing batch job completion under load. No visible behavior changes.

* **Chores**
  * Added a read-only option to internal transaction utilities to better align background tasks with best practices and improve reliability. No user action required.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobCancellationManager.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobCancellationManager.kt
@@ -136,7 +136,7 @@ class BatchJobCancellationManager(
 
   private fun tryWaitForBatchJobCompletedStatus(jobId: Long) {
     waitFor(pollTime = 200, timeout = 2000) {
-      executeInNewTransaction(transactionManager) {
+      executeInNewTransaction(transactionManager, readOnly = true) {
         batchJobService.getJobDto(jobId).status.completed
       }
     }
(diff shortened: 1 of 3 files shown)


backend/data/src/main/kotlin/io/tolgee/batch/BatchJobCancellationManager.kt
```diff
@@ -136,7 +136,7 @@ class BatchJobCancellationManager(
 
   private fun tryWaitForBatchJobCompletedStatus(jobId: Long) {
     waitFor(pollTime = 200, timeout = 2000) {
-      executeInNewTransaction(transactionManager) {
+      executeInNewTransaction(transactionManager, true) {
```
JanCizmar:
```suggestion
      executeInNewTransaction(transactionManager, readOnly = true) {
```

JanCizmar (approved):

JanCizmar:
I guess, this was a fix, but too late. 😀

bdshadow:
> I guess, this was a fix, but too late. 😀

i believe that there will be a real fix later. This one is just a little improvement

PR #3263, opened 2025-10-03
fix: allow disabling accounts of managed users when user cannot be removed - allows owners of organizations with sso to remove users

Anty0:


backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
```diff
@@ -260,18 +261,45 @@ class OrganizationRoleService(
...
       throw ValidationException(Message.USER_IS_MANAGED_BY_ORGANIZATION)
     }
+
+    forceRemoveUser(userId, organizationId)
+  }
+
+  @Transactional
+  fun removeOrDeactivateUser(
+    userId: Long,
+    organizationId: Long,
```
bdshadow, line 282: It seems like passing any random `organizationId` (even of a non-existent org) will lead to successful deactivation.

Is it ok? Just asking/nitpicking, maybe that's fine

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
```diff
@@ -252,14 +252,16 @@ class OrganizationController(
   }
 
   @DeleteMapping("/{organizationId:[0-9]+}/users/{userId:[0-9]+}")
-  @Operation(summary = "Remove user from organization")
+  @Operation(
+    summary = "Remove user from organization; if user is managed by the organization, their account is disabled instead"
```
JanCizmar: This summary is ugly / long. Move the description to the description field.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
```diff
@@ -252,14 +252,16 @@ class OrganizationController(
...
+  )
   @RequiresOrganizationRole(OrganizationRoleType.OWNER)
   @RequiresSuperAuthentication
   fun removeUser(
     @PathVariable("organizationId") organizationId: Long,
```
JanCizmar: We should not take the organizationId from URL, but from organizationHolder. I guess it's more secure, because there we are sure about the permissions were correctly checked.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
```diff
@@ -260,18 +261,45 @@ class OrganizationRoleService(
...
+      userAccountService.disable(userId)
+      return
+    }
+
+    removeUser(userId, organizationId)
+  }
+
+  private fun canRemoveUser(userId: Long, organizationId: Long): Boolean {
+    val managedBy = getManagedBy(userId)
+    return managedBy == null || managedBy.id != organizationId
```
JanCizmar: Please add some documentation comment here. I don't understand why this condition defines whether we can remove user.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
```diff
@@ -260,18 +261,45 @@ class OrganizationRoleService(
...
+  }
+
+  private fun forceRemoveUser(
```
JanCizmar: Again, I am missing documentation comment. What are we forcing and why.

e2e/cypress/e2e/organizations/organizationsMembers.cy.ts
```diff
@@ -73,6 +73,17 @@ describe('Organization Members', () => {
...
 
+  it('Can remove users managed by the organization', () => {
+    gcy('global-paginated-list').within(() => {
+      cy.contains('Lonely Developer')
+        .closestDcy('organization-member-item')
+        .findDcy('organization-members-remove-user-button')
+        .click();
+    });
+    confirmStandard();
+    assertMessage('User removed from organization');
```
JanCizmar, line 84: Please validate that the user is not in the list anymore.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
```diff
@@ -28,16 +28,17 @@ import org.springframework.transaction.annotation.Transactional
 import java.util.EnumSet
 
 @Service
+@Suppress("SelfReferenceConstructorParameter")
 class OrganizationRoleService(
   private val organizationRoleRepository: OrganizationRoleRepository,
   private val authenticationFacade: AuthenticationFacade,
   private val userAccountService: UserAccountService,
-  @Lazy
+  @param:Lazy
```
JanCizmar, line 36: Why we are adding this @param?

JanCizmar (changes requested):

JanCizmar:
I am also curios what happens when the company would like add the user back. Maybe they left, but after some time they come back to the company again.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
Anty0: Oops, I didn't realize I was putting that in a summary.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
Anty0: I agree. This being a hotfix PR, I didn't spend time thinking about existing code/decisions. Let me know if I should even for these PRs. ^^
I usually do check this stuff when implementing new features and bigger fix PRs. (Though I might still miss stuff—harder to notice small things in existing code.)

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
Anty0: Ye, I don't like the name of this method. It's more like "remove the user without checking if we are allowed to do it". I would rename it `removeUserForReal`, but that feels like a bit too much :D

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
JanCizmar: Sorry, it seemed to me that this line was changed. If not, just keep it as it is.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
Anty0: IDE was complaining. "The annotation is only applied to the parameter. An explicit annotation use-site target is recommended." I agree. It reads better. This way, I know where the annotation will be, even if I don't know the defaults. Do you think it's redundant?

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
Anty0: The `getManagedBy(userId)` method must return the organization with the same ID as provided in `organizationId` for us to decide to disable the user. Therefore, the organization must exist, and the user must be managed by it for us to allow disabling the user account. I do assume that the `organizationId` has been validated by the `organizationHolder`, making sure it isn't disabled, and that the controller has checked the user permissions. So, an arbitrary `organizationId` should not reach this method anyway. However, even if it does, it shouldn't allow disabling the user. Or am I missing something? ^^

I was wondering recently if I should prefer passing entities to services instead of raw IDs. From an optimization perspective, IDs seem better. However, I've encountered multiple cases where using IDs would require duplicating checks, which usually happens when the entity is fetched from the database. So, I'm inclined to prefer using entities unless speed is important for that endpoint. What do you think @bdshadow @JanCizmar?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
Anty0: I already updated it for this controller. ^^
But I take it as it was fine leaving it unchanged.

The method's name changed, but the parameters didn't, so I didn't think about them.

e2e/cypress/e2e/organizations/organizationsMembers.cy.ts
Anty0: Done. I was thinking about doing that, but checking "something doesn't exist" with Cypress is iffy, so I decided to leave it out initially. Let me know if there's a better way of checking it than what I'm doing.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
```diff
@@ -260,18 +261,60 @@ class OrganizationRoleService(
...
+    userId: Long,
+    organizationId: Long,
+  ) {
+    if (!canRemoveUser(userId, organizationId)) {
+      userAccountService.disable(userId)
+      return
+    }
+
+    removeUserForReal(userId, organizationId)
+  }
```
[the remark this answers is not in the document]
Anty0: That's intentional—removing the user does that. Disabling the user should be reversible, so the user stays unmodified.

Anty0:
> I am also curios what happens when the company would like add the user back. Maybe they left, but after some time they come back to the company again.

@JanCizmar Same. I was thinking it might be better to have a separate disable/enable endpoint that the organization owner could use to re-enable managed users. Additionally, the frontend could use a separate endpoint to disable users so we wouldn't clutter the leave endpoint. To implement this, I would keep disabled accounts on the member list, but they would be grayed out or something similar on the front end. I believe this is a better solution, although it would require more work on my side. However, it would still be semi-compatible with this PR—we can merge this PR first to have at least one way of getting rid of managed users, and then I can work on this idea.

Regarding what will actually happen, the `UserAccount` has a unique check on the username if it isn't deleted. However, most of the backend checks both `deletedAt` and `disabledAt` to determine if the account is "active," so I assume that if a user tries to sign up again after being disabled, it would cause a constraint failure on the database side and a 500 error from the server. I assumed that disabled accounts are already well-supported. I probably should review the codebase and add proper handling for disabled accounts so we can return a clear error message to the user, instructing them to contact the organization or server admin.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
bdshadow: now, when code is rewritten with using `organizationHolder`, i think it's fine by me.

I don't quite understand why `From an optimization perspective, IDs seem better.` I think passing ids or entities, in general, depends on the situation

JanCizmar:
OK, let's merge this one and please, create anew pitch for the optimal solution. Thx.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
```diff
@@ -241,25 +243,31 @@ class OrganizationController(
   @RequiresOrganizationRole(OrganizationRoleType.OWNER)
   @RequiresSuperAuthentication
   fun setUserRole(
-    @PathVariable("organizationId") organizationId: Long,
```
bdshadow, line 244: nitpicking: why is this line updated? and the next one isn't

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
```diff
@@ -241,25 +243,31 @@ class OrganizationController(
...
+    description = (
+      "Remove user from organization. " +
+        "If user is managed by the organization, their account is disabled instead."
+    )
+  )
   @RequiresOrganizationRole(OrganizationRoleType.OWNER)
   @RequiresSuperAuthentication
   fun removeUser(
-    @PathVariable("organizationId") organizationId: Long,
+    @PathVariable organizationId: Long,
```
bdshadow, line 267: same here

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
JanCizmar: 👍

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationController.kt
JanCizmar: 👍

JanCizmar:
Since @Anty0 is super busy now. I am merging this without adressing the last comments. But you're right @bdshadow, we can focus on that later.

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationRoleService.kt
Anty0: Oh, I believed the organization was checked even if I didn't use the `organizationHoler`, but you're right, it isn't. Must have remembered it wrong.

I mean, you can avoid database queries. If you never need the entity—only use it to query other entities by its ID—then there's one unnecessary query to get the entity.

PR #3265, opened 2025-10-06
fix: batch job processing

bdshadow:
## Summary by CodeRabbit

- New Features
  - New administration API endpoints to inspect project batch locks and the current batch job queue, including job status/details. Access restricted to super admins.

- Configuration
  - Added a setting to limit how many job chunks are added to the internal queue per scheduling run.

- Improvements
  - Smarter prioritization of running/started jobs and capped queue population for steadier processing.
  - Increased default retry window for the AI service to improve resilience.

- Tests
  - Added tests covering new admin endpoints and access control.

### DIFF
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationController.kt
@@ -1,7 +1,8 @@
-package io.tolgee.api.v2.controllers
+package io.tolgee.api.v2.controllers.administration
 
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.tags.Tag
+import io.tolgee.api.v2.controllers.IController
 import io.tolgee.constants.Message
 import io.tolgee.dtos.queryResults.organization.OrganizationView
 import io.tolgee.exceptions.BadRequestException
@@ -45,23 +46,23 @@ import org.springframework.web.bind.annotation.RestController
 )
 @OpenApiSelfHostedExtension
 class AdministrationController(
-  private val organizationService: OrganizationService,
-  private val pagedOrganizationResourcesAssembler: PagedResourcesAssembler<OrganizationView>,
-  private val organizationModelAssembler: OrganizationModelAssembler,
-  private val authenticationFacade: AuthenticationFacade,
-  private val userAccountService: UserAccountService,
-  private val pagedResourcesAssembler: PagedResourcesAssembler<UserAccount>,
-  private val userAccountModelAssembler: UserAccountModelAssembler,
-  private val jwtService: JwtService,
+    private val organizationService: OrganizationService,
+    private val pagedOrganizationResourcesAssembler: PagedResourcesAssembler<OrganizationView>,
+    private val organizationModelAssembler: OrganizationModelAssembler,
+    private val authenticationFacade: AuthenticationFacade,
+    private val userAccountService: UserAccountService,
+    private val pagedResourcesAssembler: PagedResourcesAssembler<UserAccount>,
+    private val userAccountModelAssembler: UserAccountModelAssembler,
+    private val jwtService: JwtService,
 ) : IController {
   @GetMapping(value = ["/organizations"])
   @Operation(summary = "Get all server organizations")
   @RequiresSuperAuthentication
   fun getOrganizations(
-    @ParameterObject
+      @ParameterObject
     @SortDefault(sort = ["name"])
     pageable: Pageable,
-    search: String? = null,
+      search: String? = null,
   ): PagedModel<OrganizationModel> {
     val organizations =
       organizationService.findAllPaged(
@@ -76,10 +77,10 @@ class AdministrationController(
   @Operation(summary = "Get all server users")
   @RequiresSuperAuthentication
   fun getUsers(
-    @ParameterObject
+      @ParameterObject
     @SortDefault(sort = ["name"])
     pageable: Pageable,
-    search: String? = null,
+      search: String? = null,
   ): PagedModel<UserAccountModel> {
     val users = userAccountService.findAllWithDisabledPaged(pageable, search)
     return pagedResourcesAssembler.toModel(users, userAccountModelAssembler)
@@ -89,7 +90,7 @@ class AdministrationController(
   @Operation(summary = "Delete user")
   @RequiresSuperAuthentication
   fun deleteUser(
-    @PathVariable userId: Long,
+      @PathVariable userId: Long,
   ) {
     if (userId == authenticationFacade.authenticatedUser.id) {
       throw BadRequestException(Message.CANNOT_DELETE_YOUR_OWN_ACCOUNT)
@@ -107,7 +108,7 @@ class AdministrationController(
   )
   @RequiresSuperAuthentication
   fun disableUser(
-    @PathVariable userId: Long,
+      @PathVariable userId: Long,
   ) {
     if (userId == authenticationFacade.authenticatedUser.id) {
       throw BadRequestException(Message.CANNOT_DISABLE_YOUR_OWN_ACCOUNT)
@@ -119,7 +120,7 @@ class AdministrationController(
   @Operation(summary = "Enable user", description = "Enables previously disabled user.")
   @RequiresSuperAuthentication
   fun enableUser(
-    @PathVariable userId: Long,
+      @PathVariable userId: Long,
   ) {
     userAccountService.enable(userId)
   }
@@ -128,8 +129,8 @@ class AdministrationController(
   @Operation(summary = "Set Role", description = "Set's the global role on the Tolgee Platform server.")
   @RequiresSuperAuthentication
   fun setRole(
-    @PathVariable userId: Long,
-    @PathVariable role: UserAccount.Role,
+      @PathVariable userId: Long,
+      @PathVariable role: UserAccount.Role,
   ) {
     val user = userAccountService.get(userId)
     user.role = role
@@ -145,7 +146,7 @@ class AdministrationController(
   )
   @RequiresSuperAuthentication
   fun generateUserToken(
-    @PathVariable userId: Long,
+      @PathVariable userId: Long,
   ): String {
     val user = userAccountService.get(userId)
     return jwtService.emitToken(user.id, true)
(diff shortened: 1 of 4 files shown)


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/ProjectBatchLockController.kt
```diff
@@ -0,0 +1,152 @@
+package io.tolgee.api.v2.controllers
```
bdshadow: i would put it into the "administration" package together with the `AdministrationController`

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/ProjectBatchLockController.kt
```diff
@@ -0,0 +1,152 @@
...
+      val jobInfo = if (lockedJobId != null && lockedJobId > 0L) {
+        try {
+          val jobDto = batchJobService.getJobDto(lockedJobId)
+          JobInfo(
+            jobId = jobDto.id,
+            status = jobDto.status,
+            type = jobDto.type,
+            createdAt = jobDto.createdAt
+          )
+        } catch (e: Exception) {
```
bdshadow: "try/catch" can be avoided here:
```kotlin
      val jobInfo = if (lockedJobId != null && lockedJobId > 0L) {
        val jobDto = batchJobService.findJobDto(lockedJobId)
        if (jobDto == null) {
          logger.warn("Locked job $lockedJobId in project $projectId not found")
          null
        } else {
          JobInfo(
            jobId = jobDto.id,
            status = jobDto.status,
            type = jobDto.type,
            createdAt = jobDto.createdAt
          )
        }
      } else {
        null
      }
```

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/ProjectBatchLockControllerTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc
+import org.springframework.boot.test.context.SpringBootTest
+
+@SpringBootTest
+@AutoConfigureMockMvc
+class ProjectBatchLockControllerTest : AuthorizedControllerTest() {
+
+  @Autowired
+  private lateinit var batchJobProjectLockingManager: BatchJobProjectLockingManager
```
bdshadow: none of the autowired fields are used in the test. Can be just removed

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
```diff
@@ -63,14 +63,21 @@ class BatchJobChunkExecutionQueue(
         from BatchJobChunkExecution bjce
         join bjce.batchJob bk
         where bjce.status = :executionStatus
-        order by bjce.createdAt asc, bjce.executeAfter asc, bjce.id asc
+        order by 
+          case when bk.status = 'RUNNING' then 0 else 1 end,
```
bdshadow: it's better to have the value `RUNNING` as a parameter too. At least, to be consistent with the existing "executionStatus" parameter + coderabbitai is also complaning that it won't work like this (though i don't believe him :) probably it read it somwhere on stackoverflow :), but let's make AI happy too :) )

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
```diff
@@ -63,14 +63,21 @@ class BatchJobChunkExecutionQueue(
...
         """.trimIndent(),
         BatchJobChunkExecutionDto::class.java,
       ).setParameter("executionStatus", BatchJobChunkExecutionStatus.PENDING)
         .setHint(
           "jakarta.persistence.lock.timeout",
           LockOptions.SKIP_LOCKED,
-        ).resultList
+        )
+        // Limit to get pending batches faster
+        .setMaxResults(1_000)
```
bdshadow: it limits how many chunks are added to the queue for processing every minute. I think such thing should be parametrized. I would put it to the `io.tolgee.configuration.tolgee.BatchProperties`

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobConcurrentLauncher.kt
```diff
@@ -101,7 +101,9 @@ class BatchJobConcurrentLauncher(
             return@repeatForever false
           }
 
-          logger.trace("Jobs to launch: $jobsToLaunch")
+          // This trace will spam the logging output
+          // (one log every 100ms), so it's commented out for now
+          // logger.trace("Jobs to launch: $jobsToLaunch")
```
bdshadow: i would definitely not leave the commented code like this. I think it can be just completely removed, because the `logItemsPulled` method already gives tracing about how many items will actually be processed.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -134,12 +134,32 @@ class BatchJobProjectLockingManager(
 
   private fun getInitialJobId(projectId: Long): Long? {
     val jobs = batchJobService.getAllIncompleteJobIds(projectId)
+
+    // First priority: Find actually RUNNING jobs from database
+    // This prevents phantom locks from PENDING jobs that have chunks in queue but aren't executing
+    val runningJob = jobs.find { job ->
+      val jobDto = batchJobService.getJobDto(job.jobId)
+      jobDto.status == io.tolgee.model.batch.BatchJobStatus.RUNNING
```
bdshadow: status can be easily added to the query inside the `getAllIncompleteJobIds` method. Also, it is used only here, so it can be safely updated. It will save additional trips to a database.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -134,12 +134,32 @@ class BatchJobProjectLockingManager(
...
         .groupBy { it.batchJobId }.map { it.key to it.value.count() }.toMap()
     // we are looking for a job that has already started and preferably for a locked one
-    return jobs.find { it.totalChunks != unlockedChunkCounts[it.jobId] }?.jobId ?: jobs.firstOrNull()?.jobId
+    val startedJob = jobs.find { it.totalChunks != unlockedChunkCounts[it.jobId] }
+    if (startedJob != null) {
+      logger.debug("Found started job ${startedJob.jobId} for project $projectId (fallback logic)")
+      return startedJob.jobId
+    }
+
+    logger.debug("No RUNNING or started jobs found for project $projectId, allowing new job to acquire lock")
```
bdshadow: i think more correct here is `No RUNNING or PENDING`

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/ProjectBatchLockController.kt
ajeans: I see AdministrationController in the same package, am I missing something?

see https://github.com/tolgee/tolgee-platform/blob/main/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/AdministrationController.kt

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
ajeans: Oh right sorry, you mentioned it earlier.

Will fix.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/ProjectBatchLockController.kt
bdshadow: i mean, create a new package named "administration" one level lower, and put them there, so it will be:
```kotlin
package io.tolgee.api.v2.controllers.administration
```

ajeans:
@bdshadow this should be ready for review now 🙏

bdshadow:
I'm ready to merge it.

just one thing
@JanCizmar could you please check the value of the property here: https://github.com/tolgee/tolgee-platform/pull/3265/files#diff-cb60057503ee621ae60183c4875b58d704711d179812f8f7bd079a4f9d25552cR13. I don't have access to logs or to the database to check how many chunks are actually managed per minute from the queue, but i believe that it is more than 1000 and it can be increased in this pr

bdshadow:
@JanCizmar i mean it would be interesting to check the following logs:
```kt
logger.debug("Removing job $jobId from queue, queue size: ${queue.size}")
```

To see how the `queue.size` is changing over time to choose some sensible default value

ajeans:
@bdshadow May I create the second PR (multi job per project) already by branching off from HEAD of this branch?
Or do I need to wait until this PR is merged?

I'd like to get the ball rolling on this subject soon, as we'll need to wait for a release with both PRs merged to unfork our production.

And we'll probably have more fixes / improvements coming once we confirm the issues are still there on the latest versions.

Thanks

bdshadow:
@ajeans I understand your concern. Let's clarify this last question, and I'll merge it. 

Could you please share your experience about the queue.size? Is 1000 enough, or can be increased, because it's emptied during that minute?

ajeans:
@bdshadow complex question :-) 

* It depends on the type of tasks. TAG_KEYS jobs are fast here but db intensive. MACHINE_TRANSLATE jobs take much longer (mean 30s) when calling OpenAI for the translations.
* It depends on job concurrency and whether all jobs target the same project (as we do) or not.
* On our production mix, with a concurrency of 5 jobs and 1_000 jobs added every minute, the batch queue keeps increasing and grows minute by minute. After a day, we may end up with a queue at 10k or 20k and degraded performance. This is not ideal and our current strategy is to roll out restart instances every 12 hours to start back at 1 000 and get back nice performance. However it is mostly due to our production having to go through the backlog when jobs were entirely stuck with the vanilla versions. 
I assume that 1 000 or 2 000 would be a nice default.
As the job queue size is exported with prometheus (key `tolgee_batch_job_execution_queue_size`), you could easily monitor and adjust this value on your SaaS production.

bdshadow:
@ajeans  thank you very much for the comment. I'm ready to merge it, but i can't approve it, since i'm the owner of the pr. @JanCizmar could you please approve and merge it?

PR #3262, opened 2025-10-03
fix: batch job processing

bdshadow:
## Summary by CodeRabbit

- New Features
  - Added admin endpoints to inspect project batch locks and the batch job queue.
  - Introduced configurable per-project batch concurrency (default 1).
  - Added optional JSON logging profile alongside standard console logging.

- Improvements
  - Optimized batch queue ordering and limited fetched items for better performance.
  - Adjusted retry strategy for LLM operations to improve resilience.

- Tests
  - Added coverage for admin endpoints, including authorization and queue/lock operations.

- Chores
  - Added a manual Release workflow to build, package, and publish Docker images.

PR #3271, opened 2025-10-11
feat: implement project locks containing multiple job ids

ajeans:
Replace old format Map<Long, Long> with new format Map<Long, Set<Long>> to support configurable concurrent jobs per project.

One-time Redis Migration automatically runs on startup and performs one-time data conversion.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

- **New Features**
  - Configure how many batch jobs can run in parallel per project.
  - Project locks can track multiple concurrent jobs and return detailed info for each.
  - Lock status now treats an empty set as unlocked and existing locks are migrated automatically.

- **Bug Fixes**
  - Missing locked jobs now emit a warning when job info is unavailable.

- **Tests**
  - Tests converted to parameterized concurrency checks and updated to verify collection-based lock state.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/batch/BatchJobManagementControllerCancellationTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/batch/BatchJobManagementControllerCancellationTest.kt
@@ -58,66 +58,68 @@ class BatchJobManagementControllerCancellationTest :
   @ValueSource(ints = [1, 2])
   @ProjectJWTAuthTestMethod
   fun `cancels a job`(projectConcurrency: Int) {
-    val keys = testData.addTranslationOperationData(100)
-    saveAndPrepare()
-
     batchProperties.projectConcurrency = projectConcurrency
 
-    val keyIds = keys.map { it.id }.toList()
+    batchDumper.finallyDumpAll {
+      val keys = testData.addTranslationOperationData(100)
+      saveAndPrepare()
 
-    val count = AtomicInteger(0)
+      val keyIds = keys.map { it.id }.toList()
 
-    doAnswer {
-      if (count.incrementAndGet() > 5) {
-        while (simulateLongRunningChunkRun) {
-          // this simulates long-running operation, which checks for active context
-          val context = it.arguments[2] as CoroutineContext
-          context.ensureActive()
-          Thread.sleep(10)
+      val count = AtomicInteger(0)
+
+      doAnswer {
+        if (count.incrementAndGet() > 5) {
+          while (simulateLongRunningChunkRun) {
+            // this simulates long-running operation, which checks for active context
+            val context = it.arguments[2] as CoroutineContext
+            context.ensureActive()
+            Thread.sleep(10)
+          }
+        }
+        it.callRealMethod()
+      }.whenever(machineTranslationChunkProcessor).process(any(), any(), any(), any())
+
+      performProjectAuthPost(
+        "start-batch-job/machine-translate",
+        mapOf(
+          "keyIds" to keyIds,
+          "targetLanguageIds" to
+            listOf(
+              testData.projectBuilder.getLanguageByTag("cs")!!.self.id,
+            ),
+        ),
+      ).andIsOk
+
+      waitFor {
+        batchJobConcurrentLauncher.runningJobs.size >= 5
+      }
+
+      val job = util.getSingleJob()
+
+      // Verify the job is locked
+      val lockedJobs = batchJobProjectLockingManager.getLockedForProject(testData.project.id)
+      lockedJobs.assert.contains(job.id)
+
+      performProjectAuthPut("batch-jobs/${job.id}/cancel")
+        .andIsOk
+
+      waitForNotThrowing(pollTime = 100) {
+        executeInNewTransaction {
+          util.getSingleJob().status.assert.isEqualTo(BatchJobStatus.CANCELLED)
+          verify(batchJobActivityFinalizer, times(1)).finalizeActivityWhenJobCompleted(any())
+
+          // assert activity stored
+          entityManager.createQuery("""from ActivityRevision ar where ar.batchJob.id = :id""")
+            .setParameter("id", job.id).resultList
+            .assert.hasSize(1)
         }
       }
-      it.callRealMethod()
-    }.whenever(machineTranslationChunkProcessor).process(any(), any(), any(), any())
 
-    performProjectAuthPost(
-      "start-batch-job/machine-translate",
-      mapOf(
-        "keyIds" to keyIds,
-        "targetLanguageIds" to
-          listOf(
-            testData.projectBuilder.getLanguageByTag("cs")!!.self.id,
-          ),
-      ),
-    ).andIsOk
-
-    waitFor {
-      batchJobConcurrentLauncher.runningJobs.size >= 5
+      // Verify the job lock was released
+      val lockedJobsAfterCancel = batchJobProjectLockingManager.getLockedForProject(testData.project.id)
+      lockedJobsAfterCancel.assert.doesNotContain(job.id)
     }
-
-    val job = util.getSingleJob()
-
-    // Verify the job is locked
-    val lockedJobs = batchJobProjectLockingManager.getLockedForProject(testData.project.id)
-    lockedJobs.assert.contains(job.id)
-
-    performProjectAuthPut("batch-jobs/${job.id}/cancel")
-      .andIsOk
-
-    waitForNotThrowing(pollTime = 100) {
-      executeInNewTransaction {
-        util.getSingleJob().status.assert.isEqualTo(BatchJobStatus.CANCELLED)
-        verify(batchJobActivityFinalizer, times(1)).finalizeActivityWhenJobCompleted(any())
-
-        // assert activity stored
-        entityManager.createQuery("""from ActivityRevision ar where ar.batchJob.id = :id""")
-          .setParameter("id", job.id).resultList
-          .assert.hasSize(1)
-      }
-    }
-
-    // Verify the job lock was released
-    val lockedJobsAfterCancel = batchJobProjectLockingManager.getLockedForProject(testData.project.id)
-    lockedJobsAfterCancel.assert.doesNotContain(job.id)
   }
 
   @Test
--- a/backend/app/src/test/kotlin/io/tolgee/util/BatchDumper.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/util/BatchDumper.kt
@@ -78,7 +78,19 @@ class BatchDumper(
     }
   }
 
+  fun <T> finallyDumpAll(fn: () -> T): T {
+    return try {
+      fn()
+    } finally {
+      getAllJobs().forEach {
+        this.dump(it.id)
+      }
+    }
+  }
+
   fun getSingleJob(): BatchJob = entityManager.createQuery("""from BatchJob""", BatchJob::class.java).singleResult
+
+  fun getAllJobs(): List<BatchJob> = entityManager.createQuery("""from BatchJob""", BatchJob::class.java).resultList
 
   private fun dumpQueuedItems(
     jobId: Long,


backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -51,18 +55,20 @@ class BatchJobProjectLockingManager(
     jobId: Long,
   ) {
     projectId ?: return
-    getMap().compute(projectId) { _, lockedJobId ->
+    getMap().compute(projectId) { _, lockedJobIds ->
       logger.debug("Unlocking job: $jobId for project $projectId")
-      if (lockedJobId == jobId) {
+      val currentJobs = lockedJobIds ?: emptySet()
+      if (currentJobs.contains(jobId)) {
         logger.debug("Unlocking job: $jobId for project $projectId")
```
bdshadow, line 61: i think logging should be cleaned up here as well. Currently lines 59 and 62 have it the same. I believe the first one can be removed

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -161,11 +186,47 @@ class BatchJobProjectLockingManager(
     return null
   }
 
-  private fun getRedissonProjectLocks(): RMap<Long, Long> {
-    return redissonClient.getMap("project_batch_job_locks")
+  private fun getRedissonProjectLocks(): ConcurrentMap<Long, Set<Long>> {
+      return redissonClient.getMap(REDIS_PROJECT_BATCH_JOB_LOCKS_KEY)
+    }
+
+  override fun afterPropertiesSet() {
```
bdshadow, line 192: i believe the migration is not needed, because `io.tolgee.component.CacheCleaner` clears all caches on `ApplicationReadyEvent`

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -71,65 +77,84 @@ class BatchJobProjectLockingManager(
...
+          batchJobDto.id,
+          batchJobDto.projectId,
+          newLockedJobIds
+        )
+        newLockedJobIds
+      } ?: emptySet()
+    return computedJobIds.contains(batchJobDto.id)
   }
 
   private fun computeFnBody(
```
bdshadow: maybe we can rename the function to smth more meaningful? Or am i missing the idea behind it?

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -71,65 +77,84 @@ class BatchJobProjectLockingManager(
...
-        logger.debug("No job found, locking ${toLock.id}")
-        return toLock.id
+      val initialJobId = getInitialJobId(projectId)
+      logger.info("Initial locked job $initialJobId for project ${toLock.projectId}")
+      if (initialJobId == null) {
+        logger.debug("No initial job found, locking only ${toLock.id}")
+        return setOf(toLock.id)
       }
+      val newLockedJobIds = mutableSetOf<Long>(initialJobId)
+      if (newLockedJobIds.size < batchProperties.projectConcurrency) {
```
bdshadow, line 138: won't `newLockedJobIds.size` be always equal to 1 here?

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
ajeans: I didn't know about `io.tolgee.component.CacheCleaner` 

Looking at it, it enumerates caches with `io.tolgee.constants.Caches#caches`
I don't see the project batch locks in there.

And theoretically, it would not make sense to synchronize n instances with a (key, value) in Redis (`project_batch_job_locks`), and then drop it any time any instance starts.

In our production and qa, we've not seen the locks disappear on reboots. 
We had to add scripts to clear the locks quite often when we debugged the batch problems.

The migration of data in Redis is possibly the biggest complexity in this PR, but I didn't find a way that was less intrusive. My first takes on this were to transparently handle Long? and Set<Long> data in Redis and map them to Set<Long> magically.
This proved very intrusive in a lot of methods (reading and writing), and because of type erasure, forced me to duplicate many methods.

This how I ended up a with redis value migration on start. This requires a reboot of all the instances, as the ones with version n-1 don't like the new redis value format.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
ajeans: I didn't dare to change it, but I agree the naming is so generic it is next to useless.
I'm going with `computeLockedJobIdsForProject`, feel free to rename it before merging the PR, I'm happy with any naming.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
ajeans: It seemed I handled the case where `projectConcurrency` could be 0...
Or prepared for `getInitialJobId` to return multiple ids... 😄 

YAGNI, will remove

ajeans:
@bdshadow I took care of 3 comments, and replied on the big comment regarding redis data migration.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -71,65 +76,78 @@ class BatchJobProjectLockingManager(
...
+      logger.info("Initial locked job $initialJobId for project ${toLock.projectId}")
+      if (initialJobId == null) {
+        logger.debug("No initial job found, locking only ${toLock.id}")
+        return setOf(toLock.id)
       }
+      val newLockedJobIds = mutableSetOf<Long>(initialJobId)
+      logger.debug("Locking job ${toLock.id} for project $projectId. Active jobs before: $newLockedJobIds")
+      newLockedJobIds.add(toLock.id)
+      return newLockedJobIds
+    }
```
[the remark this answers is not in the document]
ajeans: @bdshadow coderabbit is right, initial code was correct, we might have 2 entries with a concurrency of 1: the job id returned by getInitialJobId and the one that we are trying to lock.
I will fix this in the next commit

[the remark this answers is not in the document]
bdshadow: sorry, i still don't understand how `newLockedJobIds ` can have a size other than 1 after
```kt
val newLockedJobIds = mutableSetOf<Long>(initialJobId)
```
when `initialJobId` is a non-null Long value.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
bdshadow: @ajeans i agree with your points.

@JanCizmar please, take a look here too. You're more familar with the code base and you once pointed me to the `CacheCleaner`. Maybe there is smth else we're missing.

Also @JanCizmar , please, also comment about the restart of all instances at once. ArgoCD does it one by one, but it's of course possible to make it stop all pods by manually helping with `replcas: 0`.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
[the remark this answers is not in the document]
bdshadow: @coderabbitai what's the point of `newLockedJobIds.size < batchProperties.projectConcurrency` condition you are offering, when `newLockedJobIds.size` is always of size one,  because on previous line of code it's just `mutableSetOf<Long>(initialJobId)`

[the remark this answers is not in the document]
bdshadow: @coderabbitai humans are so inattentive sometimes. Thanks. 
I pointed too much attention to the usage of `newLockedJobIds.size`, that it always equals to one. And that's right, and now i finally got why it makes sense to write it like this and not as "1".
@ajeans sorry for confusion

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -71,65 +76,84 @@ class BatchJobProjectLockingManager(
...
+      if (newLockedJobIds.size < batchProperties.projectConcurrency) {
+        logger.debug("Locking job ${toLock.id} for project $projectId. Active jobs before: $newLockedJobIds")
+        newLockedJobIds.add(toLock.id)
+      } else {
+        logger.debug(
+          "Cannot lock job ${toLock.id} for project $projectId, limit reached. Active jobs: $newLockedJobIds"
+        )
+      }
+      return newLockedJobIds
+    }
```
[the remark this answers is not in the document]
bdshadow: @ajeans i think here coderabbit is also right. I think it's better to apply the changes, since it will really be correct when BatchProperties.concurrency is bigger than one

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/batch/BatchJobManagementControllerCancellationTest.kt
```diff
@@ -30,72 +33,91 @@ class BatchJobManagementControllerCancellationTest :
...
     simulateLongRunningChunkRun = false
+    batchProperties.projectConcurrency = originalProjectConcurrency
   }
 
-  @Test
+  @ParameterizedTest
+  @ValueSource(ints = [1, 2])
   @ProjectJWTAuthTestMethod
-  fun `cancels a job`() {
-    batchDumper.finallyDump {
```
bdshadow: why is `batchDumper` removed here? It's really convinient for investigation in case of a test failure

bdshadow: just maybe `batchDumper` should be extended with smth like
```kt
  fun <T> finallyDumpAll(fn: () -> T): T {
    return try {
      fn()
    } finally {
      getAllJobs().forEach { this.dump(it.id) }
    }
  }

  fun getAllJobs(): List<BatchJob> = entityManager.createQuery("""from BatchJob""", BatchJob::class.java).resultList
```

bdshadow: while playing with it i actually noticed that setting `batchProperties.projectConcurrency = 2` doesn't work here. It must be passed as application properties in the `@SpringBootTest(`

bdshadow:
@ajeans one more comment for discussion here. Jan has already written these concerns about multiple jobs per projects:

> When implementing this I wanted to prevent some unexpected situations. For example, when user would start different jobs with the same keys and then 2 parallel jobs might modify data of the same keys. E.g., one job would translate the keys while another job would delete them. I also wanted to prevent one project to take over all the workers while other project would have to wait.

How do you think it must be handled? Did you try to run 2 jobs with the same keys? Or to do batch delete while batch translation is running? Actually, some tests for these cases would be good. It would show what is the expected behaviour in these cases.

ajeans:
> @ajeans one more comment for discussion here. Jan has already written these concerns about multiple jobs per projects:
> 
> > When implementing this I wanted to prevent some unexpected situations. For example, when user would start different jobs with the same keys and then 2 parallel jobs might modify data of the same keys. E.g., one job would translate the keys while another job would delete them. I also wanted to prevent one project to take over all the workers while other project would have to wait.
> 
> How do you think it must be handled? Did you try to run 2 jobs with the same keys? Or to do batch delete while batch translation is running? Actually, some tests for these cases would be good. It would show what is the expected behaviour in these cases.

@bdshadow unfortunately, that is beyond my area of expertise.

We've been running with a project concurrency at 10 and have no seen no significant problems in production.
I admit we have only one LLM configured to asynchronously modify translations, not several providers at once.

My assumption was that you would keep the project concurrency at 1. We could even add a warning about this, saying anything >1 is experimental.

Regards

bdshadow:
@ajeans thank you for the clarification.
As far as i remember, the main problem you were trying to solve is that the batch translation of a huge number of keys took you too much time. Did you try to increase the `BatchProperties.concurrency` after the first fix is merged? I saw your email, where you said that you experienced problems with it, but i believe it was before the first pr was merged.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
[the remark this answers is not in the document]
ajeans: @bdshadow you mean I should switch make `getInitialJobId` turn into `getInitialJobIds`?

I'd rather keep the PR minimal and as close to the version I have in production as possible.

Plus this is very theoretical: once, when nothing is in redis, you might not use as many concurrent batches as possible with the new configuration.
What is going to happen is that it will then fill the remaining slots one minute later at most by adding jobs to the locks using the standard case.

If we really wanted to make this more understandable, I'd totally remove `getInitialJobId` and this specific seeding, and always let normal operations take locks through the standard paths. Now that normal operations favor RUNNING jobs (from the previous merged PR), it would be much simpler and would work.

Again, I'd rather make this PR as contained as possible, and as close as my production / battle tested version as possible if that works for you guys.

@coderabbitai is my understanding correct or am I missing something?

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/batch/BatchJobManagementControllerCancellationTest.kt
ajeans: re `batchDumper`, I missed this, will have a look.

re `batchProperties.projectConcurrency = 2` how did you notice it doesn't work? Spring will start it with its standard configuration, and then it is overridden before the test, and then returned to the original configuration after the tests.
As the checks against the properties in `BatchJobProjectLockingManager` are always done dynamically, it should work.
This approach allows "@ParameterizedTest" to be used. With spring test properties, it would require instantiating a different spring boot test context with different methods.

ajeans:
> @ajeans thank you for the clarification. As far as i remember, the main problem you were trying to solve is that the batch translation of a huge number of keys took you too much time. Did you try to increase the `BatchProperties.concurrency` after the first fix is merged? I saw your email, where you said that you experienced problems with it, but i believe it was before the first pr was merged.

@bdshadow 

What we are currently running in production is 
* the branch in the original PR: https://github.com/tolgee/tolgee-platform/pull/3262
* 2 servers that serve HTTP traffic running with `TOLGEE_BATCH_CONCURRENCY=10` and `TOLGEE_BATCH_PROJECT_CONCURRENCY=1`
* 2 servers that *don't* serve HTTP traffic running with `TOLGEE_BATCH_CONCURRENCY=10` and `TOLGEE_BATCH_PROJECT_CONCURRENCY=5`

In terms of code, the running code is very close to https://github.com/tolgee/tolgee-platform/pull/3265 plus https://github.com/tolgee/tolgee-platform/pull/3271. Main differences are tests (much better in the upstream version) and some code style improvements that didn't change the logic.

Now if I explain the history on our production systems.
* version `v3.123.13` failed to process batch jobs, we would handle at most 10 a day
* version `v3.123.13`+ https://github.com/tolgee/tolgee-platform/pull/3265 processed between 300 and 1_000 batch jobs a day
* version `v3.123.13`+ https://github.com/tolgee/tolgee-platform/pull/3265 + https://github.com/tolgee/tolgee-platform/pull/3271 with the setup described above processes 10_000 to 20_000 batch jobs a day.

As we have 1 project in production, obviously the per project lock is a major bottleneck.

We are well on our way to get the batch jobs backlog down to 0 in a few days with this version.

My current plan is that when this PR is merged, and the backlog is gone in production, to move to the latest vanilla version and do the regression testing in QA, and the battle testing in production.

Hope it makes sense. 🙏

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/batch/BatchJobManagementControllerCancellationTest.kt
ajeans: I just pushed the requested `batchDumper` changes.

Unfortunately tests are all failing on my machine today (due to PG setup), but the change makes sense and it all compiles... Let's see if the CI agrees :D 

@bdshadow Would you have any idea why the tests might be failing locally?
The build system seems to have launched the docker containers successfully

```
CONTAINER ID   IMAGE                           COMMAND                  CREATED       STATUS      PORTS                                           NAMES
ed02aeb8b7a6   postgres:16.3                   "docker-entrypoint.s…"   2 weeks ago   Up 3 days   0.0.0.0:55434->5432/tcp, [::]:55434->5432/tcp   tolgee_backend_tests_postgres_ee
bf2b3485c2f4   postgres:16.3                   "docker-entrypoint.s…"   2 weeks ago   Up 3 days   0.0.0.0:55433->5432/tcp, [::]:55433->5432/tcp   tolgee_backend_tests_postgres_main
```

bdshadow: sorry, about `batchProperties.projectConcurrency=2` comment. Seems like i had some other changes locally, which affected what i saw during the debug

about failing tests. Is it still failing for you? What error message to you have?

bdshadow:
@ajeans thank you very much for a detailed explanation.

I understand your motivation behind these changes. However, it will affect lots of other batch processes, which can lead to unpredictable results when launched simultaneously. I also understand that in your specific case it works, but i'm trying to find a solution which is more general. 

Did you try to just increase the TOLGEE_BATCH_CONCURRENCY? I mean run it with just the first fix and for servers that don't serve HTTP traffic just increase the TOLGEE_BATCH_CONCURRENCY to, for example, 20, or even more. Maybe in this case, the `BatchProperties.chunkQueuePopulationSize` can be increased too

ajeans:
> @ajeans thank you very much for a detailed explanation.
> 

Hi @bdshadow 

> I understand your motivation behind these changes. However, it will affect lots of other batch processes, which can lead to unpredictable results when launched simultaneously. I also understand that in your specific case it works, but i'm trying to find a solution which is more general.
> 

Feel free to find other ways to address the scalability issue with big projects with a lot of translations. 
I'm eager to have Tolgee servers that can address my traffic in production, but I have no emotional attachment to this PR.

I only went this route and submitted because @JanCizmar suggested it was a way to address this.
And if you make "1" the default, there is absolutely no downside to it.
I don't understand the reluctance, but that is obviously your call.

Irrelevant of this PR or any other solution, do you think tolgee will support my performance requirements in the near future? I'd rather know now if this will not happen please. 
Production traffic will keep rising in the coming months, and I'd rather plan ahead the next steps.

> Did you try to just increase the TOLGEE_BATCH_CONCURRENCY? I mean run it with just the first fix and for servers that don't serve HTTP traffic just increase the TOLGEE_BATCH_CONCURRENCY to, for example, 20, or even more. Maybe in this case, the `BatchProperties.chunkQueuePopulationSize` can be increased too

AFAIU, current main is identical to the version that managed 300 to 1000 batches a day. (see https://github.com/tolgee/tolgee-platform/pull/3271#issuecomment-3418621310). There were no changes in the batches in the last 3 months, and my patches applied cleanly.

I don't believe that playing with chunks will be a 10x improvement to the workload, especially as we are mostly bound to the active number of threads calling OpenAI translations.

I'm on a vacation next week. If you have more questions I will answer them when I'm back.

@JanCizmar by then, can you weigh in as whether you see tolgee handling our performance requirements?

JanCizmar:
Hello @ajeans !

Thanks for raising this pull request and for the detailed context 😊

@bdshadow is part of the Tolgee team and he’s currently focused on improving batch operations performance across both self‑hosted and Tolgee Cloud instances. We absolutely agree that the current batch throughput is a bottleneck.

At the same time, we’re cautious about making the project lock limit configurable. Allowing multiple concurrent jobs per project could lead to race conditions or data inconsistencies when two jobs operate on the same keys. That’s why we originally kept the per‑project limit at one. If you need more throughput, increasing the global TOLGEE_BATCH_CONCURRENCY should have the same effect. If it doesn’t, it suggests another bottleneck that we need to fix, and we’ll look into that.

Our goal is to find a solution that keeps project locking safe while giving a significant performance boost. Dmitrii is working hard on this now, and we hope to have a robust fix soon. We’re should also improve our internal monitoring so we can see potential bottlenecks in Grafana.

Thanks again for tackling this and providing so much context; it really helps us prioritise and improve Tolgee 🚀

ajeans:
@JanCizmar 

Thank you for the detailed response. I'd like to provide more context on why increasing global TOLGEE_BATCH_CONCURRENCY alone doesn't solve the scalability issue for our production workload.
I also want to share the performance requirements that we have and that would be expected from the upstream version of tolgee

 
**Production Timeline**


 
tolgee-production-timeline


  This graph shows daily successful batch job completions from September through today:

  1. Baseline (Sept 1-22): ~10-100 jobs/day with various TOLGEE_BATCH_CONCURRENCY settings (1-20). The system was severely bottlenecked.

  2. Mass Cancellation (Sept 22): We manually cancelled ~110,000 stuck jobs to try to recover.

  3. PR #3265 Applied (~Sept 29): The phantom lock fixes brought us to ~1,000 jobs/day - a significant improvement, but still insufficient.

  4. PR #3265 + #3271 Applied (~Oct 6): Adding multi-job-per-project capability with projectConcurrency=5 jumped throughput to 15,000-21,000 jobs/day - a 15x improvement over #3265 alone.

 

  **Our Production Architecture**

  - 4 Tolgee servers in a Redis-coordinated cluster

    - 2 servers handle HTTP traffic: TOLGEE_BATCH_CONCURRENCY=10, TOLGEE_BATCH_PROJECT_CONCURRENCY=1

    - 2 dedicated batch workers: TOLGEE_BATCH_CONCURRENCY=10, TOLGEE_BATCH_PROJECT_CONCURRENCY=5

  - 1 large project with >300,000 translation keys

  - Continuous batch translation workload (LLM-based)

  **Why Global Concurrency Alone Doesn't Work**

  The issue is the per-project lock bottleneck:

  4 servers × 10 concurrent workers = 40 threads trying to process jobs
                      ↓
              All jobs belong to 1 project
                      ↓
      Only 1 job can hold the project lock at a time
                      ↓
    Only chunks from that 1 job can be processed across all servers
                      ↓
       Other threads spin waiting for chunks from the locked job


  **With the current single-job-per-project locking:**

  - Only 1 job across all 4 servers can hold the project lock

  - Threads can process multiple chunks from that one job concurrently (good!)

  - BUT when chunks from the locked job aren't readily available in the local queue, threads must wait

  - Meanwhile, chunks from other jobs (which could be processed) are blocked by the project lock

  - The system becomes starved: threads are idle waiting for the "right" chunks while work sits in queue

  - Result: ~1,000 jobs/day maximum throughput despite 40 available threads


  **With projectConcurrency=5:**

  - 5 different jobs can hold the project lock simultaneously

  - This means chunks from 5 different jobs are eligible for processing at any time

  - Threads have a much higher probability of finding processable chunks in their queue

  - Better work distribution across the 40 available threads

  - Result: ~15,000 jobs/day throughput
 

  The bottleneck isn't CPU or I/O - it's work starvation caused by the single-job lock limiting which chunks are eligible for processing.

 

  **Addressing the Safety Concerns**

  I understand the concern about concurrent modifications to the same keys. However: 

  1. Chunk-level isolation already exists: Each BatchJobChunkExecution is an atomic unit. Two chunks never operate on the same keys simultaneously - that's guaranteed by the chunking logic itself.

  2. Job-level conflicts are rare in practice: In our production workload, users don't typically launch conflicting job types (translate + delete) on the same keys simultaneously. When they do create a new batch job, it's usually after the
  previous one completes.

  3. Default value keeps existing behavior: With projectConcurrency defaulting to 1, existing deployments get zero behavior change. The safety model remains identical.

  4. Opt-in for advanced users: Deployments like ours with a single large project and controlled batch workflows can opt into higher concurrency after understanding the trade-offs. 

  **Path Forward**

  I respectfully suggest:

  - Keep projectConcurrency=1 as the default (no behavior change for existing users)

  - Add a configuration warning in the docs: "Values >1 are experimental and should only be used in controlled environments where batch job conflicts are managed externally"

This unblocks high-throughput single-project deployments like ours

Gives the Tolgee team time to explore a more comprehensive solution for multi-project safety, while Rakuten France can safely operate during the shopping season.

  The alternative is that deployments with large projects cannot scale beyond ~1,000 jobs/day on the current architecture, even with unlimited hardware resources (tolgee instances or postgres databases).
 
Thanks for your consideration!

bdshadow:
@ajeans thank you very much for such a detailed report. It is really helpful.

What I don't like is the part "With the current single-job-per-project locking" about the processing of chunks by threads. Especially the following:
> BUT when chunks from the locked job aren't readily available in the local queue, threads must wait

And when i say i don't like it, i mean the way tolgee does it. 

I think i found the root of a problem with populating the in-memory queue. Please check the first pull request: https://github.com/tolgee/tolgee-platform/pull/3287. It has a detailed explanation + it's duplicated in the commit messages.

Shortly: locking the db rows didn't work when selecting chunks for processing. 

But as described in the commit messages, it still doesn't fully fix the problem. Let's imagine we have 2 pods running, each trying to populate the in-memory queue. The first one locks the rows in the chunks table and sends them to processing. Transaction is finished, rows in the chunks table are unlocked. Now the 2nd pod comes, before the  actual processing of the chunks is actually started in the first pod(update of the status from PENDING to RUNNING). And the second pod selects absolutely the same chunks and sends them to its own queue. Now these pods are conflicting for processing the same chunks.

I see that it can be fixed by introducing a new status for chunks: NEW. And when `BatchJobChunkExecutionQueue.populateQueue` runs with locking db rows, it will also update the status from NEW to PENDING. This way, another pod won't be able to take them and chunks will be uniquely distributed among all k8s pods.

I hope with these changes there won't be need in allowing multiple batch jobs per project

bdshadow:
I've made a draft of the refactored batch job processing. Although i like it more :), the processing perfomance was just a little bit better (meanwhile i think it's a too big refactoring for smth that works - read further :) )

I've started testing the main branch locally. I've run 2 instances of the app, with redis, and tried the batch machine translation of 10k keys:
 BatchProperties.concurrency | BatchProperties.chunkQueuePopulationSize | Time
--- | --- | ---
1 | 1000 | 3:30
2 | 1000 | 2:03
3 | 1000 | 1:41
4 | 1000 | 1:42

So bumped into the same problem - increasing the number of threads doesn't improve the performance. The reason is
> BUT when chunks from the locked job aren't readily available in the local queue, threads must wait

BUT! The solution is simple: increase `BatchProperties.chunkQueuePopulationSize`:
|   |   |   |
|---|---|---|
|4| 1200 | 1:20

Now the queue is populated quicker, never empty, threads don't wait, performance improved.

@ajeans could you please try the latest version from main, increase the `BatchProperties.chunkQueuePopulationSize` as well as the `BatchProperties.concurrency`. 

Alternatively we can add an additional trigger to fetch pending chunks from db when queue is empty. For example, somwhere in the `BatchJobConcurrentLauncher.getSleepTime` like this:
```kt
  private fun getSleepTime(
    startTime: Long,
    somethingHandled: Boolean,
  ): Long {
    if (!batchJobChunkExecutionQueue.isEmpty() && jobsToLaunch > 0 && somethingHandled) {
      return 0
    }
    batchJobChunkExecutionQueue.populateQueue();
    return MIN_TIME_BETWEEN_OPERATIONS - (System.currentTimeMillis() - startTime)
  }
```

ajeans:
> I've made a draft of the refactored batch job processing. Although i like it more :), the processing perfomance was just a little bit better (meanwhile i think it's a too big refactoring for smth that works - read further :) )
> 
> I've started testing the main branch locally. I've run 2 instances of the app, with redis, and tried the batch machine translation of 10k keys:
> BatchProperties.concurrency 	BatchProperties.chunkQueuePopulationSize 	Time
> 1 	1000 	3:30
> 2 	1000 	2:03
> 3 	1000 	1:41
> 4 	1000 	1:42
> 
> So bumped into the same problem - increasing the number of threads doesn't improve the performance. The reason is
> 
> > BUT when chunks from the locked job aren't readily available in the local queue, threads must wait
> 
> BUT! The solution is simple: increase `BatchProperties.chunkQueuePopulationSize`:
> 		
> 4 	1200 	1:20
> 
> Now the queue is populated quicker, never empty, threads don't wait, performance improved.
> 
> @ajeans could you please try the latest version from main, increase the `BatchProperties.chunkQueuePopulationSize` as well as the `BatchProperties.concurrency`.
> 


Hello @bdshadow 

Very interesting.

Unfortunately, I cannot test this in production, maybe early next year after the shopping season.
Plus with the current production, my backlog problem has been solved, we translate asynchronously in near real time.

Several questions:
* what happens if you create 100k keys, does it increase time linearly or is the degradation worse? If you have an easy way to test this (maybe even test scripts you can share)
* just to make sure: in your tests, do all keys belong to the same project, and they all threads synchronize on the redis project lock?
* in your tests, I assume you mocked the translations. Do your mocks return instantly or do they actually pause for seconds to simulate an actual API call. I believe the mean translation time of 30 seconds hurts my use case a lot.

> Alternatively we can add an additional trigger to fetch pending chunks from db when queue is empty. For example, somwhere in the `BatchJobConcurrentLauncher.getSleepTime` like this:
> 
> ```kotlin
>   private fun getSleepTime(
>     startTime: Long,
>     somethingHandled: Boolean,
>   ): Long {
>     if (!batchJobChunkExecutionQueue.isEmpty() && jobsToLaunch > 0 && somethingHandled) {
>       return 0
>     }
>     batchJobChunkExecutionQueue.populateQueue();
>     return MIN_TIME_BETWEEN_OPERATIONS - (System.currentTimeMillis() - startTime)
>   }
> ```

Oh you want to populate the queue faster. The problems I saw at the beginning were with too big queues blocked on the lock, but there you would focus on pending chunks so it might be a good idea.

JanCizmar:
Hey @bdshadow,
can you please tell us if this can be closed?

@ajeans do you still encounter this issues after the updates we released?

bdshadow:
@ajeans hello. I would kindly ask you for a review of the pr: https://github.com/tolgee/tolgee-platform/pull/3363, if you have time. I think such a change will solve your problem with filling the job queue with the same execution items

ajeans:
> Hey @bdshadow, can you please tell us if this can be closed?
> 
> @ajeans do you still encounter this issues after the updates we released?

We are still on the forked version from October in production.
Now that the shopping season is over, I can look into moving to the latest released version. 

Preferably once there is a version with PR #3363 as this looks promising.
We are having issues from time to time that we cannot readily explain with many chunk executions in status PENDING. Having a clear NEW vs PENDING status would help us troubleshoot better.

JanCizmar:
I am closing this for now in favor of https://github.com/tolgee/tolgee-platform/pull/3392.

Anyway, thanks a lot for this PR and for the effort you put into this PR. We really appreciate it. ❤️

I might be wrong, but based on the original architecture, increasing the number of jobs locked per project likely improved performance due to a side effect. This should be addressed differently and should respect the original design.

We did deeper analysis with tests on larger jobs and proper measurements. It looks like the optimization in #3392 should resolve the issue.

@ajeans Can you please test it with the latest released version to confirm we are going in the right direction. Thanks.

ajeans:
Hi @JanCizmar 

We plan to test the latest vanilla release in production in February, as we'll need to speed up batch processing by an order of magnitude by March.
I'll keep you posted.

CC @jbbouille

PR #3277, opened 2025-10-15
feat: add subscription and org details to Chatwoot support widget

dzienisz:
## Summary by CodeRabbit

* **New Features**
  * Help chat moved to a centralized hook-driven integration: availability reflects account/org support features and user state.
  * Opening chat now closes the help menu and consistently launches the support widget.

* **Bug Fixes**
  * More reliable script loading, initialization, and user/config propagation to prevent duplicate or inconsistent widget behavior.

### DIFF
--- a/webapp/src/hooks/useChatwoot.ts
+++ b/webapp/src/hooks/useChatwoot.ts
@@ -32,16 +32,14 @@ function loadScript(doc: Document, url: string) {
 }
 
 async function loadChatwoot(websiteToken: string, darkMode: boolean) {
-  // @ts-ignore
-  window.chatwootSettings = {
+  window['chatwootSettings'] = {
     darkMode: darkMode ? 'auto' : 'light',
     hideMessageBubble: true,
   };
 
   await loadScript(document, BASE_URL + '/packs/js/sdk.js');
 
-  // @ts-ignore
-  window.chatwootSDK?.run({
+  window['chatwootSDK']?.run({
     websiteToken,
     baseUrl: BASE_URL,
   });
@@ -56,8 +54,7 @@ async function loadChatwootOnce(websiteToken: string, darkMode: boolean) {
 }
 
 function setChatwootUser(user: User) {
-  // @ts-ignore
-  window.$chatwoot?.setUser(user.id, {
+  window['$chatwoot']?.setUser(user.id, {
     email: user!.username,
     name: user!.name,
     url: window.location,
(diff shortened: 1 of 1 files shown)


webapp/src/component/HelpMenu.tsx
```diff
@@ -124,6 +124,20 @@ export const HelpMenu = () => {
         name: user!.name,
         url: window.location,
       });
+
+      // Set custom attributes for paying customers
+      if (preferredOrganization) {
+        const subscription = preferredOrganization.activeCloudSubscription;
+        // @ts-ignore
+        window.$chatwoot.setCustomAttributes({
```
JanCizmar: Hmm, this looks like potential issue when the `$chatwoot` is not defined. I wouldn't use `@ts-ignore` ignore here.

Also I would move this to separate hook `usePrepareChatwootAttributes`. This component us already long enough.

webapp/src/component/HelpMenu.tsx
```diff
@@ -124,6 +124,20 @@ export const HelpMenu = () => {
         name: user!.name,
         url: window.location,
       });
+
+      // Set custom attributes for paying customers
```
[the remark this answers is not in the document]
JanCizmar: The comment is actually misleading. You are setting the attributes for everyone. Not just for paying customers. This coderabbit comment is correct.

JanCizmar (changes requested):

JanCizmar:
Sorry for all the tasks, but can you also finish this? @Anty0

webapp/src/component/HelpMenu.tsx
Anty0: I ended up moving the entire Chatwoot setup to a separate hook.

The `ts-ignore` must stay because it's needed to access the property, but I wonder how to handle the case where the property is missing — for example, if Chatwoot loading fails. Currently, it just fails silently.

Anty0: Or, I guess, what Coderabbit suggested: "type declaration file for the Window interface".

JanCizmar: I guess you can access it like this. `window["$chatwoot"]?.setC`... Right?

Anty0: I'm feeling like I could have thought of this too 😅
Done ^^

JanCizmar (approved):

PR #3282, opened 2025-10-20
Added PO handling for new lines + tests

pasotee:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of multiline translations in PO files, including correct handling of LF and CRLF line breaks.
  * Enhanced interpretation of escape sequences inside quoted strings so escaped characters are correctly rendered.

* **Tests**
  * Updated test fixtures and expectations to reflect additional parsed translations and multiline handling.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/po/in/PoParser.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/po/in/PoParser.kt
@@ -126,14 +126,18 @@ class PoParser(
 
   private fun Char.handleOther() {
     if (currentEscaped) {
-      val specialEscape: Char? = if (quoted) when (this) {
-        'n'  -> '\n'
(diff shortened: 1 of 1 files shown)


backend/data/src/main/kotlin/io/tolgee/formats/po/in/PoParser.kt
```diff
@@ -126,8 +126,25 @@ class PoParser(
...
+        'n' -> '\n'
+        'r' -> '\r'
+        't' -> '\t'
+        '"' -> '"'
+        '\\' -> '\\'
+        else -> null
+      }
+      } else {
+        null
+      }
```
Anty0, line 140: This appears to be incorrectly formatted, but the linter didn't catch it—we should check why later.

PR #3238, opened 2025-09-08
feat: UserAccount's lastActivity in /v2/administration/users

xsvetli1:
## Summary by CodeRabbit

- New Features
  - Administration > Users now displays each user’s last activity (shows a timestamp or “No activity yet”) with improved readability.
- Documentation
  - Added example Tolgee backend auth configuration.
  - Added instructions to refresh the webapp API schema after backend changes.
- Chores
  - Removed deprecated public LLM prompt endpoint and related types from the public API.
  - Extended user model to include last activity for use in the UI.

PR #3286, opened 2025-10-22
User last activity

### LINKED ISSUE #2973: [API] Get user's last connection/interaction date/time
On self-hosted instances, when working on open-source projects, most of the times anonymous voluteers will register to a server to contribute to the translation, but at some point they will stop contributing (and other people will take their place), leaving their accounts abandoned on the server. This will eventually lead to the reach of the maximum seats limit associated to the current license.

It would be very useful to be able to get the user's last connection date from the API (via `/v2/user` or `/v2/administration/users`), so inactive users can be removed/disabled via API, and new people can replace the old abandoned accounts.

Thanks!

zomp:
last activity screenshot

Requested in https://github.com/tolgee/tolgee-platform/issues/2973.

Based on https://github.com/tolgee/tolgee-platform/pull/3238.

## Summary by CodeRabbit

* **New Features**
  * Users list now shows formatted "last activity" with a compact two-column user item and action area.

* **Documentation**
  * Development guide refined for clarity, formatting, examples and authentication snippets.

* **Chores**
  * Public user schemas and API responses migrated to an administration-focused user model.
  * Administration endpoints now require stricter super-admin access.
  * Public LLM prompt endpoint and related schemas removed.

* **Tests**
  * E2E helpers updated and a test added to verify user last-activity display and date forcing.


JanCizmar (commented):
Thanks for working on this! 🙏

I have checked it fast and I have few points.
- I don't think that the feature request specified the need for frontend. However, it's nice to have.
- We need to add backend test
- we need to add e2e tests

e2e/cypress/e2e/administration/base.cy.ts
```diff
@@ -71,9 +77,25 @@ describe('Administration', () => {
...
+      .contains('No activity yet');
+    forceDate(new Date('2023-08-28T02:30').getTime()); // app displays local time, set this one in local timezone
+    createProject({
+      name: 'just.to.record.activity',
+      languages: [{ name: 'cs', originalName: 'čj', tag: 'cs' }],
+    });
+    gcy('settings-menu-item').contains('Users').click(); // reload
+    getUserListItem('Peter Administrator')
+      .findDcy('administration-user-activity')
+      .contains('Last activity on August 28, 2023 at 2:30 AM');
```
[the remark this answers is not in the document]
zomp: This is a false issue. Frontend displays the date in its locale which is based on the test browser's machine, and the test sets the date in the locale of the test executor's machine, which is the same machine, so the date matches and can be matched fully.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationControllerTest.kt
```diff
@@ -69,6 +72,41 @@ class AdministrationControllerTest : AuthorizedControllerTest() {
...
+    performAuthGet("/v2/administration/users?search=${userAccount!!.username}").andAssertThatJson {
+      node("_embedded.users") {
+        node("[0]") {
+          node("lastActivity").isEqualTo(forcedDate.time)
+        }
+      }
+    }
+
+    currentDateProvider.forcedDate = prevForcedDate
+  }
```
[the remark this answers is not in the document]
zomp: Fixed in https://github.com/tolgee/tolgee-platform/pull/3286/commits/731eb8a94e2d86a8468cf8b60b2245addf29e9d2

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationControllerTest.kt
```diff
@@ -69,6 +72,43 @@ class AdministrationControllerTest : AuthorizedControllerTest() {
...
+
+      performAuthGet("/v2/administration/users?search=${userAccount!!.username}").andAssertThatJson {
+        node("_embedded.users") {
+          node("[0]") {
+            node("lastActivity").isEqualTo(forcedDate.time)
+          }
+        }
+      }
+    } finally {
+      currentDateProvider.forcedDate = prevForcedDate
```
Anty0: I'd just put `currentDateProvider.forcedDate = null` in `@AfterEach` function. (Default is always null for the `forcedDate`.)

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationControllerTest.kt
```diff
@@ -69,6 +72,43 @@ class AdministrationControllerTest : AuthorizedControllerTest() {
...
+
+      performAuthGet("/v2/administration/users?search=${userAccount!!.username}").andAssertThatJson {
+        node("_embedded.users") {
+          node("[0]") {
+            node("lastActivity").isNull()
+          }
+        }
+      }
+
+      val organization = dbPopulator.createOrganization("just.to.record.activity", userAccount!!)
```
Anty0: We want to phase out the `dbPopulator` and replace it with test data everywhere. In this case, I would use the default organization of one of the existing users from the `AdministrationTestData` (you will need to create a property for it in the `AdministrationTestData` to access it from the test).

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
```diff
@@ -260,7 +261,28 @@ interface UserAccountRepository : JpaRepository<UserAccount, Long> {
...
+      userAccount.role,
+      userAccount.isInitialUser,
+      userAccount.totpKey,
+      userAccount.deletedAt,
+      userAccount.disabledAt,
+      la.lastActivity
+    )
     from UserAccount userAccount
+    left join userAccount.emailVerification ev
+    left join lastActivityCTE la on la.authorId = userAccount.id
```
Anty0, line 285: Could this break the paging? I didn't look too closely, but Code Rabbit is complaining. I don't know how Spring handles count queries, but I think specifying one here might be a good idea if Spring can't reduce this SQL to the `select count(id) from UserAccount where...`. I believe Spring can't do that when there is a join from `with`.

webapp/src/views/administration/AdministrationUsers.tsx
```diff
@@ -70,9 +80,29 @@ export const AdministrationUsers = ({
...
+                        ? t('administration_user_no_activity')
+                        : t('administration_user_last_activity', {
+                            date: formatDate(new Date(u.lastActivity), {
+                              dateStyle: 'long',
+                              timeStyle: 'short',
+                            }),
+                          })}
+                    </Typography>
+                  </Box>
+                </Box>
```
Anty0: Maybe we should split this into `AdministrationUserItem`. The tree is starting to get deep.

webapp/src/websocket-client/WebsocketClient.ts
```diff
@@ -192,7 +192,10 @@ export type EntityModification<T> = T extends keyof schemas
 
 export type WebsocketEvent<Data> = {
   activityId: number;
-  actor: { type: 'user'; data: components['schemas']['UserAccountModel'] };
+  actor: {
+    type: 'user';
+    data: components['schemas']['UserAccountAdministrationModel'];
```
Anty0: This is (and was) wrong. I looked at the code, and on the backend side, we send `SimpleUserAccountModel`, so there was a typo. (It really doesn't make sense for the websocket to receive the`UserAccountAdministrationModel`.)

webapp/src/component/security/UserMenu/MenuHeader.tsx
```diff
@@ -3,7 +3,7 @@ import { AvatarImg } from 'tg.component/common/avatar/AvatarImg';
 import { components } from 'tg.service/apiSchema.generated';
 
 type UserAccountModel =
-  | components['schemas']['UserAccountModel']
+  | components['schemas']['UserAccountAdministrationModel']
```
Anty0: Let's use `SimpleUserAccountModel`; it should have all the props we need here. Looking at the code, we only pass either the organization or private user account model here — alternatively, we can use the private model.

e2e/cypress/e2e/administration/base.cy.ts
```diff
@@ -12,18 +18,26 @@ import { administrationTestData } from '../../common/apiCalls/testData/testData'
...
+    }
     administrationTestData.clean();
     administrationTestData.generate().then((res) => {});
     login('admin@admin.com');
-    visitAdministration();
   });
 
   afterEach(() => {
+    if (Cypress.currentTest.title === activityTest) {
+      releaseForcedDate();
```
Anty0: It can be called every time, no matter which test it was.

e2e/cypress/e2e/administration/base.cy.ts
```diff
@@ -12,18 +18,26 @@ import { administrationTestData } from '../../common/apiCalls/testData/testData'
...
+  visitAdministrationOrganizations,
+  visitAdministrationUsers,
 } from '../../common/administration';
 
+const activityTest = "can display user's last activity";
+
 describe('Administration', () => {
   beforeEach(() => {
+    if (Cypress.currentTest.title === activityTest) {
+      forceDate(new Date('2023-08-28T02:30').getTime()); // app displays local time, set this one in local timezone
```
Anty0: I'd call this inside the test. Since we don't need to create test data at that time, we can set the time just before creating the project.

Anty0 (changes requested):
Good job, looking good. I have some notes, but they should all be easy to fix ^^

webapp/src/websocket-client/WebsocketClient.ts
zomp: Fixed in 055045a3d

webapp/src/component/security/UserMenu/MenuHeader.tsx
zomp: Fixed in 055045a3d

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationControllerTest.kt
zomp: Implemented in cab9feb

Originally I wanted to avoid resetting it always because `currentDateProvider.forcedDate` has a custom setter with more logic than just assigning the variable, so I wanted to use it only when date forcing is done. I agree simplicity is important and now I even see the proposed `@AfterEach` is used on various places across the app, so the convention is also important.

webapp/src/views/administration/AdministrationUsers.tsx
zomp: Let me open the topic. I do not see a clear split line – why should we move these things to a separate component, but not move e.g. the MFA badge there? To me, those user attributes belong together. But if we did so, the deep tree would be kept, just in the new component. ... To me the current state looks better than the proposed one, but I am opened to ideas.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/administration/AdministrationControllerTest.kt
zomp: Resolved in [8cd1a88](https://github.com/tolgee/tolgee-platform/pull/3286/commits/8cd1a88b68fcb75f0f10c25d425ec7c6d7f523af)

(It appeared this approach is already used in the file.)

e2e/cypress/e2e/administration/base.cy.ts
zomp: Would it be OK with you if I moved the test and the test specific setup and teardown to a nested `describe`, respectively its `beforeEach` and `afterEach`? I would like to avoid invoking date releasing everywhere, because it is quite time-wise expensive.

e2e/cypress/e2e/administration/base.cy.ts
zomp: ~~Please let us move the topic to https://github.com/tolgee/tolgee-platform/pull/3286#discussion_r2485889587~~

... If you are OK with the nested `describe` because of forced date releasing in a one-time `afterEach`, I will move `forceDate` to the test just before creating the project.

e2e/cypress/e2e/administration/base.cy.ts
Anty0: Sure, it should be okay here. ^^

webapp/src/views/administration/AdministrationUsers.tsx
Anty0: I think there was a misunderstanding. I see that GitHub only selected the new code, even though I tried to reference the entire lambda function.

The entire `renderItem` function is essentially a separate component that takes the user as a prop and renders the item for that user, so I would split it off. That would mean moving all the user attributes to that new component. ^^

e2e/cypress/e2e/administration/base.cy.ts
zomp: Implemented in baa5b9e

e2e/cypress/e2e/administration/base.cy.ts
zomp: Implemented in baa5b9e

webapp/src/views/administration/AdministrationUsers.tsx
zomp: Implemented in bcc8011

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
zomp: The code seemed to work. There is already some support implemented (e.g. since Hibernate 6.2).

To make sure there is no hidden issue, `countQuery` was added in https://github.com/tolgee/tolgee-platform/pull/3286/commits/56ee88f7bdac79ce8d1d0c3f02ad9fb59350aea5

backend/data/src/main/kotlin/io/tolgee/dtos/queryResults/UserAccountAdministrationView.kt
```diff
@@ -0,0 +1,22 @@
...
+
+data class UserAccountAdministrationView(
+  val id: Long,
+  override val username: String,
+  val name: String,
+  val emailAwaitingVerification: String?,
+  val avatarHash: String?,
+  @Deprecated("See `UserAccount` for details.")
+  /** @see [io.tolgee.model.UserAccount.thirdPartyAuthType] */
+  override val accountType: UserAccount.AccountType?,
```
Anty0, line 15: @JanCizmar What do you think about this? I wouldn't mark it as deprecated. If anything, the `accountType` is duplicated data in the database, but from the API perspective, I wouldn't stop providing the `accountType` – I wouldn't deprecate it.

Anty0 (approved):
LGTM (except for the note on the deprecated property). The PR itself needs to wait until we extend the activity; otherwise, the `lastActivity` would lag behind when the user is not performing any of the tracked activities.

backend/data/src/main/kotlin/io/tolgee/dtos/queryResults/UserAccountAdministrationView.kt
zomp: It comes from my possible misunderstanding of:
```kotlin
  /**
   * This property is redundant - it's value can be derived from other existing properties.
   * Kept for legacy reasons.
```
in ` backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt`.

Anty0: Yup, from my perspective, it isn't really deprecated. It can be calculated from other columns rather than stored in its own database column—that's what the comment is trying to say.

Anty0:
The pitch for the missing feature was created: #3304

Once implemented, this PR can be merged.

PR #3303, opened 2025-11-10
chore: reenable ktlint rules + reformat files

Anty0:
Should be merged at the same time as https://github.com/tolgee/billing/pull/207

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,1 @@
+9e29d3af2c20aabf2ab04ed66b8fb6c196d333e3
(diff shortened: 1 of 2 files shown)


cyyynthia:
Might be a good time to introduce a `.git-blame-ignore-revs` file ;)

Anty0:
@cyyynthia Yup, we should add the final squashed commit to that file for sure :3
(Might be even worth splitting the actual editor config update and upcoming ktlint update into a separate commit so only formatting changes are ignored.)

bdshadow:
@cyyynthia  nice thing, didn't know about it

@Anty0 agree with both points. Update of the .editorconfig must be done in a separate commit (i see you already did this). I will add the update of the ktlint version in a separate commit too

Also we should add to the developer notes info that `.git-blame-ignore-revs` file must be added to config: `git config blame.ignoreRevsFile .git-blame-ignore-revs`

PR #3328, opened 2025-11-29
feat: Store data for LLM Analytics, Diagnostics, Fixed LLM errors

JanCizmar:
## Summary by CodeRabbit

* **New Features**
  * Activity tracking for AI prompt operations (create/update/delete)
  * Configurable max token limits per LLM provider
  * Translations now record which prompt produced them and include prompt usage details in responses

* **Chores**
  * Refactored prompt response models and API responses
  * Improved diagnostics (Sentry context) for LLM integrations
  * Database schema updated to support prompt → translation association

### DIFF
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/hateoas/model/prompt/PromptResponseUsageModel.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/api/v2/hateoas/model/prompt/PromptResponseUsageModel.kt
@@ -6,4 +6,4 @@ data class PromptResponseUsageModel(
   val inputTokens: Long? = null,
   val outputTokens: Long? = null,
   val cachedTokens: Long? = null,
-) : RepresentationModel<PromptResponseModel>()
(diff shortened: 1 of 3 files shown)


backend/data/src/main/kotlin/io/tolgee/service/translation/AutoTranslationService.kt
```diff
@@ -113,16 +113,16 @@ class AutoTranslationService(
...
 
-    translation.setValueAndState(project, text, usedService)
+    translation.setValueAndState(project, text, usedService, promptId)
   }
 
   private fun getAutoTranslatedValue(
     config: AutoTranslationConfig,
     translation: Translation,
-  ): Pair<String?, MtServiceType?>? {
+  ): Triple<String?, MtServiceType?, Long?>? {
```
Anty0, line 125: Nitpick: I'm not sure Triple is great for readability. Deciphering it feels harder than it should. It isn't too bad, but I would need to look at the source code of `getAutoTranslatedValue` to understand that the last value is `promptId`. But it's a small private function, so probably not worth changing.

PR #3334, opened 2025-12-02
feat: add `pnpm` support

zomp:
List of changes to be done to add pnpm support.

Reasons to use pnpm:
- Should be faster on CI.
- Packages are installed for the whole repo via a single command to a +- single location.

Reasons to stay with npm:
- Standard & better tested.
- Easier to use, easier to resolve issues (packages are installed next to each subproject independently).

The PR is not complete. If we decide to go the pnpm way, we should:
- Remove `package-lock.json` files in favor of single `pnpm-lock.yaml`.
- Use pnpm for installing and managing dependencies.
- Update Gradle tasks (`installWebappDeps`...).
- Update CI.

Other notes:
- If nothing else, pnpm identified some bugs and their fixes are worthy to be applied even if npm is kept.
- Changes would deserve more testing.
- Once marked stable, Storybook addon will be moved to own repository, so to init the project, only `npm install --prefix library && npm install --prefix webapp` will be necessary. IMO pnpm will not make sense then. ... Unless we decide to move a monorepo way.
- pnpm cannot be used the standard way – instead of linking dependencies, we hoist them to the repo root (`nodeLinker: hoisted`). We have to, because we also have to preserve symlinks (`preserveSymlinks: true`) in Vite because of local dependency modules loading (e.g. dependencies of /library have to be relative to library project when library is directly included from /webapp, not relative to webapp itself –  e.g. some dependencies would not be present or on different versions there).
- E2E tests are not migrated to pnpm in this PR.

## Summary by CodeRabbit

* **Chores**
  * Established pnpm workspace configuration with hoisted node linker
  * Updated core dependencies: Husky, Vite, React, and related tooling
  * Streamlined build scripts and initialization processes
  * Updated ESLint configurations across packages for improved TypeScript project resolution
  * Expanded Vite dev server file system access for better development experience


zomp:
@coderabbitai review

JanCizmar (dismissed):
Looks like you would like to get higher in line change statistics 👍

PR #3338, opened 2025-12-07
chore: existing symlink not removed properly

aditya-mitra:
## problem

When doing `npm install` in the webapp folder, the following error pops up as the symlink is not deleted.

image

## solution

`existsSync()` returns false for broken/existing symlinks. `lstatSync` throws error if the path does not exist.

## Summary by CodeRabbit

* **Chores**
  * Improved robustness of the module preparation script: safer handling of file operations to avoid race conditions and transient errors during initialization, reducing spurious failures in setup and the build process.

### DIFF
--- a/webapp/scripts/prepareEe.js
+++ b/webapp/scripts/prepareEe.js
@@ -17,10 +17,10 @@ const eeModuleLocal = path.join(
   '../src/eeSetup/eeModule.current.tsx'
 );
 
-try {
-  lstatSync(eeModuleLocal);
+// remove the symlink if it exists
+if (!!lstatSync(eeModuleLocal, { throwIfNoEntry: false })) {
   unlinkSync(eeModuleLocal);
-} catch {}
+}
 
 if (existsSync(srcDir)) {
   symlinkSync(eeModuleEe, eeModuleLocal);


JanCizmar:
Cool! Thx.

webapp/scripts/prepareEe.js
```diff
@@ -17,9 +17,10 @@ const eeModuleLocal = path.join(
   '../src/eeSetup/eeModule.current.tsx'
 );
 
-if (existsSync(eeModuleLocal)) {
+try {
+  lstatSync(eeModuleLocal);
   unlinkSync(eeModuleLocal);
-}
+} catch {}
```
zomp: Hi Aditya, this is a great catch. Thank you. 🙇

It took me some time to understand the bug and I would not get it without your PR description. Would it be OK with you to make the fix easier to understand even to the slower or us? 🦥

I think either

```suggestion
try {
  unlinkSync(eeModuleLocal);
} catch {}
```

or

```suggestion
if (!!lstatSync(eeModuleLocal, { throwIfNoEntry: false })) {
  unlinkSync(eeModuleLocal);
}
```

could be easier to understand.

What is your opinion please?

aditya-mitra: Makes sense.
Thanks for point that out Jan!

zomp: Thank you.

To speed it up, I pushed one of the proposals. Feel free to shout if you are unhappy about something, otherwise I will merge the PR.

PR #3341, opened 2025-12-09
feat: expandable sidebar

aditya-mitra:
It is hard to recognize the items by their icons - I often hover over them for a second to know where they will redirect.

Added an expandable sidebar to quickly know where it would be navigating.

https://github.com/user-attachments/assets/be92043b-eb83-49b4-b679-09c39d25c2e1

## Summary by CodeRabbit

* **New Features**
  * Added collapsible project menu with toggle button
  * Menu items display labels when expanded, icons only when collapsed
  * Smooth transitions between expanded and collapsed states

### DIFF
--- a/webapp/src/views/projects/projectMenu/ProjectMenu.tsx
+++ b/webapp/src/views/projects/projectMenu/ProjectMenu.tsx
@@ -9,6 +9,8 @@ import {
   Translate01,
   UploadCloud02,
   User01,
+  ChevronLeft,
+  ChevronRight,
 } from '@untitled-ui/icons-react';
 import { Link, LINKS, PARAMS } from 'tg.constants/links';
 import { useConfig } from 'tg.globalContext/helpers';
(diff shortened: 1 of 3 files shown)


JanCizmar:
Hey Aditya! That's slay! ❤️

@ZuzanaOdstrcilova, what do you think?

ZuzanaOdstrcilova:
Hey Aditya! Thank you for making nice improvements. We did talk about it in the past, but then we postponed it. I had a similar idea, based on the fact that users don't need to see the menu all the time. We will have to check its responsiveness (for other views). We discussed your solution with Jan, and we might move it to later to gather more info. We will probably need to reorganize the menu so we can do it together. Thank you a lot:).

A quick AI preview of our different idea:

https://github.com/user-attachments/assets/afe83f3c-ae0e-43f2-b3af-62994b133a9d

@JanCizmar will plan a call with you.

Issue #3336, opened 2025-12-04
Allow deselecting all selected languages in translation view


bratzie:
**Is your feature request related to a problem? Please describe.**
I very often switch between languages since. It's annoying having to un-click everything when it could quicker.

**Describe the solution you'd like**
I think a "Deselect all" toggle or something would make a lot of sense here.

**Describe alternatives you've considered**
I have un-clicked everything one by one a few times now and I feel like my time could be spent doing other things. (Like writing feature requests)

**Additional context**
This is the dropdown I am referring to:

<img width="276" height="542" alt="Image" src="https://github.com/user-attachments/assets/d3e2072a-5b40-4c47-8574-eb226fd61bbd" />

aditya-mitra:
yes, this feature seems handy

PR #3345, opened 2025-12-10
feat: select/deselect all languages in translations language selector

aditya-mitra:
Allow to select all and deselect all when in the translations multi-selector.

https://github.com/user-attachments/assets/f70451a8-ea96-432e-b061-302e612e1d99

closes https://github.com/tolgee/tolgee-platform/issues/3336

## Release Notes

**New Features**
- Users can now select an empty language option in the language selector.
- Batch operation controls ("All" and "None") are now available in translation contexts.

### DIFF
--- a/webapp/src/component/common/form/LanguagesSelect/getLanguagesContent.tsx
+++ b/webapp/src/component/common/form/LanguagesSelect/getLanguagesContent.tsx
@@ -64,7 +64,8 @@ export const getLanguagesContent = ({
   );
 
   // Показываем кнопки "All" и "None" только для batch операций
-  const isBatchOperation = context === 'batch-operations';
+  const isBatchOperation =
+    context === 'batch-operations' || context === 'translations';
 
   const languageItems = languages.map((lang) => (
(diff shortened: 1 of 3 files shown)


JanCizmar:
Thanks! Looks cool.

@ZuzanaOdstrcilova, can you check it?

JanCizmar:
One thing comes into my mind and that's we have limit that only 10 languages can be selected in the same time. Doesn't it break when I select all?

aditya-mitra:
> One thing comes into my mind and that's we have limit that only 10 languages can be selected in the same time. Doesn't it break when I select all?

Let me give it a try today and find out

ZuzanaOdstrcilova:
Hey Aditya, I like your ideas, but unfortunately, there's a limit of 10 languages. Someone already asked us about this. We will discuss it internally and let you know. Thank you:)

JanCizmar:
Maybe we can only keep the unselect all option, or add some clear button which would deselect all except the base language?

PR #3363, opened 2026-01-05
perf: refactor batch job local queue population

bdshadow:
perf: refactor batch job local queue population

The refactoring is actually bigger than adding a new status. Here is a brief before-vs-after comparison:

BEFORE:
1) The PENDING status of the execution item wasn't changed, when the item was already in the queue. So in case of long-running jobs, such executions items are read from db again
2) All executions items were sent to all pods through redis. This didn't make sense, because
    a) they were competing for the lock in db;
    b) items were unnecessarily occupying memory - they are processed by a single pod anyway
    c) since the queue was big, it took time to filter out the items which are already in process by the other instance
3) non-reliable adding to queue. Because it could happen that when item was saved to db and sent to redis - other pod can concurrently receive this message from redis as well as read is from db by a scheduled job. There is a check `if (queue.contains(item)) {add(it)}`, but it's not atomic and could easily lead to duplicities.

AFTER
1) The execution items are saved to db with status=NEW
2) Only these new NEW items are read from db. NOTE: in rare cases, for example during deployment, it can happen that pod is killed and items are stuck in PENDING state - that's why there is an additional condition to check for such cases; and even if we take an item which is still in queue, there won't be concurrency problem described in BEFORE-3 - this item is in this queue for a long time already)
3) When an item is taken to local queue, its status is updated to PENDING
4) An item is not sent through redis to other pods
5) Other pods are notified about a new job and can concurrently read items from db themselves.
6) Now a local queue in a single pod has almost unique (among all pods) list of items - meaning less memory taken, quicker to process, less concurrency conflicts both in the app and db

## Release Notes

* **New Features**
  * Introduced new status state for batch job executions, enabling improved tracking and visibility into job progress.

* **Improvements**
  * Enhanced batch job queue reliability with improved detection and handling of stuck or stale executions.
  * Strengthened multi-instance batch processing coordination through improved event-driven messaging infrastructure.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
@@ -6,6 +6,10 @@ import io.tolgee.batch.data.BatchJobChunkExecutionDto
 import io.tolgee.batch.data.ExecutionQueueItem
 import io.tolgee.batch.data.QueueEventType
 import io.tolgee.batch.events.JobQueueItemsEvent
+import io.tolgee.batch.events.JobQueueNewChunkExecutionsEvent
+import io.tolgee.batch.events.NewJobEvent
+import io.tolgee.batch.events.OnBatchJobCreated
+import io.tolgee.component.CurrentDateProvider
 import io.tolgee.component.UsingRedisProvider
 import io.tolgee.configuration.tolgee.BatchProperties
 import io.tolgee.model.batch.BatchJobChunkExecution
@@ -13,6 +17,8 @@ import io.tolgee.model.batch.BatchJobChunkExecutionStatus
 import io.tolgee.model.batch.BatchJobStatus
 import io.tolgee.pubSub.RedisPubSubReceiverConfiguration
 import io.tolgee.util.Logging
+import io.tolgee.util.addMinutes
+import io.tolgee.util.executeInNewTransaction
 import io.tolgee.util.logger
 import io.tolgee.util.trace
 import jakarta.persistence.EntityManager
@@ -20,12 +26,15 @@ import org.hibernate.LockMode
 import org.hibernate.LockOptions
 import org.hibernate.Session
 import org.springframework.beans.factory.InitializingBean
+import org.springframework.context.ApplicationContext
 import org.springframework.context.annotation.Lazy
 import org.springframework.context.event.EventListener
 import org.springframework.data.redis.core.StringRedisTemplate
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
+import org.springframework.transaction.PlatformTransactionManager
 import org.springframework.transaction.annotation.Transactional
+import org.springframework.transaction.event.TransactionalEventListener
 import java.util.concurrent.ConcurrentLinkedQueue
 
 @Component
@@ -36,6 +45,9 @@ class BatchJobChunkExecutionQueue(
   @Lazy
   private val redisTemplate: StringRedisTemplate,
   private val metrics: Metrics,
+  private val applicationContext: ApplicationContext,
+  private val platformTransactionManager: PlatformTransactionManager,
+  private val currentDateProvider: CurrentDateProvider,
 ) : Logging,
   InitializingBean {
   companion object {
@@ -48,9 +60,23 @@ class BatchJobChunkExecutionQueue(
   @EventListener
   fun onJobItemEvent(event: JobQueueItemsEvent) {
     when (event.type) {
-      QueueEventType.ADD -> this.addItemsToLocalQueue(event.items)
+      // will actually not be used at all
+      QueueEventType.ADD -> this.addItemsToQueue(event.items)
+
+      // still makes sense, because in very rare cases (for example on app restart)
+      // we can have executionItem in multiple app instances
       QueueEventType.REMOVE -> queue.removeAll(event.items.toSet())
     }
+  }
+
+  @EventListener
+  fun onNewJobEvent(event: NewJobEvent) {
+    fetchAndQueueNewJobExecutions(event.jobId)
+  }
+
+  @TransactionalEventListener
+  fun onNewChunkFromDb(event: JobQueueNewChunkExecutionsEvent) {
+    addExecutionsToQueue(event.items)
   }
 
   @Scheduled(fixedDelay = 60000)
@@ -62,20 +88,24 @@ class BatchJobChunkExecutionQueue(
       // which is not available in the jpa Query class.
       entityManager
         .unwrap(Session::class.java)
+        // select only NEW execution items
+        // select pending only which are in pending state for too long (stuck probably due to pod restart on deployment)
         .createQuery(
           """
-          select new io.tolgee.batch.data.BatchJobChunkExecutionDto(bjce.id, bk.id, bjce.executeAfter, bk.jobCharacter)
           from BatchJobChunkExecution bjce
           join bjce.batchJob bk
-          where bjce.status = :executionStatus
+          where bjce.status = :newExecutionStatus
+              or (bjce.status = :pendingExecutionStatus and bjce.createdAt < :isStuckBefore)
           order by 
             case when bk.status = :runningStatus then 0 else 1 end,
             bjce.createdAt asc, 
             bjce.executeAfter asc, 
             bjce.id asc
           """.trimIndent(),
-          BatchJobChunkExecutionDto::class.java,
-        ).setParameter("executionStatus", BatchJobChunkExecutionStatus.PENDING)
+          BatchJobChunkExecution::class.java,
+        ).setParameter("newExecutionStatus", BatchJobChunkExecutionStatus.NEW)
+        .setParameter("pendingExecutionStatus", BatchJobChunkExecutionStatus.PENDING)
+        .setParameter("isStuckBefore", currentDateProvider.date.addMinutes(-2))
         .setParameter("runningStatus", BatchJobStatus.RUNNING)
         // setLockMode here is per alias of the tables from the queryString.
         // will generate: "select ... for no key update of bjce skip locked"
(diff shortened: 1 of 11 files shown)


backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
```diff
@@ -48,11 +60,25 @@ class BatchJobChunkExecutionQueue(
...
+  fun onNewJobEvent(event: NewJobEvent) {
+    fetchAndQueueNewJobExecutions(event.jobId)
+  }
+
+  @TransactionalEventListener
+  fun onNewChunkFromDb(event: JobQueueNewChunkExecutionsEvent) {
+    addExecutionsToQueue(event.items)
+  }
+
   @Scheduled(fixedDelay = 60000)
```
ajeans, line 82: This has been this way before, but I'm thinking this should be made configurable
(`fixedDelayString` as in ScheduledJobCleaner.kt)

And I'm also thinking that adding jitter would be better (randomized `initialDelayString`). On several machines, because the output of the SQL script is ordered, we'll have several nodes competing to do the SELECT FOR UPDATE

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
```diff
@@ -62,20 +88,24 @@ class BatchJobChunkExecutionQueue(
...
             bjce.createdAt asc, 
             bjce.executeAfter asc, 
             bjce.id asc
           """.trimIndent(),
-          BatchJobChunkExecutionDto::class.java,
-        ).setParameter("executionStatus", BatchJobChunkExecutionStatus.PENDING)
+          BatchJobChunkExecution::class.java,
+        ).setParameter("newExecutionStatus", BatchJobChunkExecutionStatus.NEW)
+        .setParameter("pendingExecutionStatus", BatchJobChunkExecutionStatus.PENDING)
+        .setParameter("isStuckBefore", currentDateProvider.date.addMinutes(-2))
```
ajeans, line 108: Two minutes delay seems aggressive, if you consider a use case where a batch job has 10 to 20 chunks, with each chunk being a call to OpenAI or similar with 20s / 30s latency.

It would be nice if this could be made configurable, or increased (5 minutes?)

ajeans (approved):
This seems like a very nice step.

I'm still a bit baffled by the database vs redis acting as the lock holder, but this seems to make things simpler and the rationale sounds good.

I added two comments that I think would be improvements but are not necessary.

Once this PR lands and I get a released version, I'll try this version in production at Rakuten France.

JanCizmar:
I am closing this for now in favor of https://github.com/tolgee/tolgee-platform/pull/3392.

We did deeper analysis with tests on larger jobs and proper measurements. It looks like the optimization in #3392 should resolve the issue.

@ajeans Can you please test it with the latest released version to confirm we are going in the right direction. Thanks.

ajeans:
> I am closing this for now in favor of #3392.
> 
> We did deeper analysis with tests on larger jobs and proper measurements. It looks like the optimization in #3392 should resolve the issue.
> 
> @ajeans Can you please test it with the latest released version to confirm we are going in the right direction. Thanks.

@JanCizmar reviving this thread
We upgraded to version `3.158.0` today, and performance is back to horrible levels: in the last hour, not one batch was processed.
Where is the best place to discuss? Should I open a new issue?

JanCizmar:
Oh no... Sorry for the inconvenience. Can you please provide me with your batch job server configuration properties and how big job are we speaking about? 😢

ajeans:
@JanCizmar 

Here is the config, note that we have 
* concurrency of 1 for all servers but one, who has 20
* project concurrency is immaterial, this only means something when we have our fork (who has multiple jobs under the redis lock)

```
    JAVA_TOOL_OPTIONS: '"-Xmx2g"'
    SPRING_DATASOURCE_HIKARI_KEEPALIVE_TIME: "45000"
    SPRING_DATASOURCE_HIKARI_MAXIMUM_POOL_SIZE: {{ .Values | get "tolgeeDatasourceHikariMaximumPoolSize" "25" }}
    SPRING_DATASOURCE_HIKARI_MAX_LIFETIME: "180000"
    SPRING_DATASOURCE_URL: {{ .Values | get "postgresTolgeeUrl" "" }}
    SPRING_DATASOURCE_USERNAME: "postgres"
    SPRING_DATA_REDIS_HOST: {{ .Values | get "redisTolgeeHost" "" }}
    SPRING_DATA_REDIS_PORT: {{ .Values | get "redisTolgeePort" "" }}
    SPRING_PROFILES_ACTIVE: json-logging
    TOLGEE_AUTHENTICATION_ENABLED: "true"
    TOLGEE_AUTHENTICATION_INITIAL_USERNAME: admin
    TOLGEE_BATCH_CONCURRENCY: {{ .Values | get "tolgeeBatchConcurrency" "1" }}
    TOLGEE_BATCH_JOB_CHUNK_QUEUE_POPULATION_SIZE: "250"
    TOLGEE_BATCH_PROJECT_CONCURRENCY: {{ .Values | get "tolgeeBatchProjectConcurrency" "1" }}
    TOLGEE_BATCH_SCHEDULED_HANDLE_STUCK_JOB_DELAY: "1800000"
    TOLGEE_BATCH_SCHEDULED_UNLOCK_JOB_DELAY: "60000"
    TOLGEE_CACHE_ENABLED: "true"
    TOLGEE_CACHE_USE_REDIS: "true"
    TOLGEE_MAX_TRANSLATION_TEXT_LENGTH: "100000"
    TOLGEE_POSTGRES-AUTOSTART_ENABLED: "false"
    TOLGEE_RATE_LIMITS_USER_REQUEST_LIMIT: "2000"
```

Can I share some specific details with you?
We currently have 56k batches in status PENDING, and metric `tolgee_batch_job_execution_queue_size` reported 114k

We have already pushed to production a fork which is `v3.158.0` + https://github.com/tolgee/tolgee-platform/commit/ffd658b45e1fc0ddce526839dfac7ae38332ec1f

See branch at https://github.com/RakutenFrance/tolgee-platform/tree/rakuten/v3.158.0

At least this brings back some batch processing, we are now at 1 to 2 batches processed per minute, and metric `tolgee_batch_job_execution_queue_size` reports around 1k elements in the in-memory queue.

JanCizmar:
Hey!                                                                                                         
                                                                                                                                                                                                                                                                                                                      
Sorry again for the inconvenience. We've tested batch jobs on 100k+ items and the performance was >100 items/s, so there's likely something specific to your instance / setup causing the bottleneck.

To help us get to the bottom of this, we recently added OpenTelemetry tracing to the batch job pipeline. Would you be open to setting up a local observability stack? We have docs for that here: https://github.com/tolgee/tolgee-platform/tree/main/docs/local-observability-stack (introduced in
  https://github.com/tolgee/tolgee-platform/pull/3363).

If that works for you, once you have tracing running, please reproduce the slow batch job and share traces for the following spans:

  1. populateQueue - If this is slow or infrequent, jobs may be sitting idle before processing even starts.
  2. handleItem - The main span for processing each chunk. Look at the overall duration and the attributes batch.job.id, batch.job.type, and batch.chunk.execution.id.
  3. getExecutionIfCanAcquireLockInDb - this acquires a DB lock per chunk. Long durations here would indicate lock contention.
  4. processChunk - this is where the actual work happens. This is the most important one — if this is slow, the bottleneck is in the processing itself (e.g., slow DB queries, external API calls).
  5. handleProgress / handleChunkCompletedCommitted - these handle post-processing bookkeeping. Usually fast, but worth checking.
 
  Ideally, we'd need:
  - A screenshot or export of a few representative traces showing the span hierarchy and durations
  - The batch.job.type you're running (e.g., machine translation, auto translation, etc.)
  - The total number of keys/items in the batch

  This should help us pinpoint exactly where the time is being spent.

ajeans:
Hey @JanCizmar 

As as short stop gap measure, we have re-implemented the two patches we currently have. See https://github.com/RakutenFrance/tolgee-platform/issues/16 for details.

> Sorry again for the inconvenience. We've tested batch jobs on 100k+ items and the performance was >100 items/s, so there's likely something specific to your instance / setup causing the bottleneck.
> 

We suspect a combination of two things
* All our keys are in the same project.
  * Was this tested in your performance setup? I expect that this is where the Redis project lock will hurt the most.
* We see weird timings for MACHINE_TRANSLATE tasks
  * Metrics `http_client_requests_seconds_sum` divided by `http_client_requests_seconds_count` show that HTTP calls to OpenAI are fast (between 2 to 5 seconds)
  * New metrics `tolgee_batch_chunk_execution_seconds_sum` divided by `tolgee_batch_chunk_execution_seconds_count` show that `MACHINE_TRANSLATE` batches take much longer and we have weird quantiles. Note that `TAG_KEYS`batches look fine.

This is a bit confusing.


> To help us get to the bottom of this, we recently added OpenTelemetry tracing to the batch job pipeline. Would you be open to setting up a local observability stack? We have docs for that here: https://github.com/tolgee/tolgee-platform/tree/main/docs/local-observability-stack (introduced in #3363).
> 
> If that works for you, once you have tracing running, please reproduce the slow batch job and share traces for the following spans:
> 
>     1. populateQueue - If this is slow or infrequent, jobs may be sitting idle before processing even starts.
> 
>     2. handleItem - The main span for processing each chunk. Look at the overall duration and the attributes batch.job.id, batch.job.type, and batch.chunk.execution.id.
> 
>     3. getExecutionIfCanAcquireLockInDb - this acquires a DB lock per chunk. Long durations here would indicate lock contention.
> 
>     4. processChunk - this is where the actual work happens. This is the most important one — if this is slow, the bottleneck is in the processing itself (e.g., slow DB queries, external API calls).
> 
>     5. handleProgress / handleChunkCompletedCommitted - these handle post-processing bookkeeping. Usually fast, but worth checking.
> 
> 
> Ideally, we'd need:
> 
>     * A screenshot or export of a few representative traces showing the span hierarchy and durations
> 
>     * The batch.job.type you're running (e.g., machine translation, auto translation, etc.)
> 
>     * The total number of keys/items in the batch
> 
> 
> This should help us pinpoint exactly where the time is being spent.

Thanks for the documentation. 
Yes we'll do this. 
We'll schedule this analysis very soon, and try to pinpoint where the performance loss is.

Regards

PR #3381, opened 2026-01-13
feat: glossary improvements

### LINKED ISSUE #3320: Glossary improvements
### 1. Problem
The glossary feature is a strong request from users. We built an initial version, but some usability issues remain and parts of the workflow are still inefficient.

Problems:

1. Glossary table is hard to use:
E.g. Users struggle with translating terms into languages when the glossary has too many rows. It’s not always clear which column belongs to which language.
2. It’s unclear how languages are assigned to the glossary. 
3. Because the glossary doesn’t explicitly define its languages, empty languages are not exported, which can feel confusing or broken.
4. Navigation between glossary and translation view is complicated.
5. The functionality is missing easy editing options (e.g. quickly translating a term into a specific language).
6. Some interactions and empty states could be clearer and easier to understand.

---

### 2. Appetite
2 weeks? _To discuss with Jan and Jirka._

---

### 3. Solution (basic drawings)
We defined the list of issues that we want to address.

**1. Glossary Table**
- Sticky header for better readability (the row with languages will always stay visible).
- Non-translatable cells – hover highlight + tooltip explaining why the cell is not editable.
- Remember the selected languages for the user (session storage).

Image

**2. Languages in glossary**
- Multiselect when creating/editing a glossary
   - by default, all project languages are selected
- It will be clearer which languages the glossary uses
- It will fix export behavior (exporting languages defined in the glossary)
- (i) we can hide info in the end of the form

Image

**3. Export all languages**
- The glossary will explicitly define its languages, so the entire table can be exported, including previously empty languages.

**4. Quick navigation in glossary**
- Switching between glossaries in org
- Infobar with links to projects which are using current glossary

UI variant
Image

**How to format CSV link**
- change link to docs - new title and link

**5. Quick edits in translation view, better empty states and links**
- Empty states can be more specific, clearer, and include better links (open in a new tab)

Image

- Quick edits can be inline, inside the tooltip component
- If we show as tooltip a whole card actions as "click on info icon" will not be confusing
- Opening the glossary should go to a new tab, useful when doing more edits in the glossary

UI Variant
Image

### 4. UI
Variants to discuss
https://www.figma.com/design/UYsyo57SCcWFpyMirhGmv4/Tasks-2025?node-id=1630-9577&t=kpYOVviZweGK1Tnj-1

---

### 5. Rabbit Holes
- Big complexity for appetite.
- Languages on glossary.

---

### 6. No-Goes
- Not full term edit in translation view.
- Language settings on glossaries should stay easy.

---

### 7. Success Criteria

- Fewer switches between glossary and translation view.
- More quick edits done directly in the translation view.
- Correct export of all glossary languages (no workaround needed).

**Mesurement:**
- PostHog events: opening glossary in a new tab, quick-edit usage.
- Feedback and satisfaction from users who previously reported issues.

Anty0:
- Glossary Table – made header sticky
- Glossary Export – exports all associated languages
- Navigation improvements - switch between glossaries from the navigation bar, navigate to projects from the glossary and back to the glossary from a project
- Inline editing for glossaries on the translations page
- Persist selected languages when browsing glossary terms
- Show full preview of glossary on hover in Glossary Panel
- Fixed empty glossaries page spacing
- Fixed links to docs

Fixes #3320

## Summary by CodeRabbit

* **New Features**
  * Inline glossary term editing in the glossary panel with keyboard support.
  * Glossary switcher, popover selector, and Project link component.
  * Project-aware glossary listing endpoint and organization-scoped tag retrieval.
  * Persistent per-glossary language preferences and export of organization-level tags.

* **Improvements**
  * Centralized permission hook for glossary actions.
  * Updated glossary empty-state and navigation UX; doc links refreshed.

* **Bug Fixes**
  * Updated test selectors and UI positioning/styling; stricter glossary term validation.

* **Tests**
  * New end-to-end tests for inline glossary editing and panel flows.

### DIFF
--- a/webapp/src/ee/glossary/components/GlossaryLinksList.tsx
+++ b/webapp/src/ee/glossary/components/GlossaryLinksList.tsx
@@ -45,7 +45,7 @@ export const GlossaryLinksList = ({
     <Tooltip
       placement="bottom-start"
       title={
-        <Typography variant="body2" component="span">
+        <Typography variant="body2" component="div">
           {remainingGlossaries.map((glossary) => (
             <div key={glossary.id}>
               <GlossaryLink


webapp/src/ee/glossary/hooks/useSelectedGlossaryLanguages.ts
```diff
@@ -1,20 +1,36 @@
...
+    if (selectedFor !== glossary.id) {
+      return defaultValue;
+    }
     if (selected === undefined) {
-      return undefined;
+      return defaultValue;
     }
     return [glossary?.baseLanguageTag || '', ...(selected ?? [])];
-  }, [selected, glossary]);
+  }, [selectedFor, selected, glossary]);
```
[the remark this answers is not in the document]
Anty0: Is this one fixed with the latest changes?

webapp/src/ee/glossary/components/GlossariesPanel.tsx
```diff
@@ -38,32 +41,55 @@ const useGlossaryTermsHighlightsForPanel = ({
 
 export const GlossariesPanel: React.VFC<PanelContentProps> = (data) => {
   const { language, baseLanguage, project, appendValue } = data;
+  const { preferredOrganization } = usePreferredOrganization();
+
   const terms = useGlossaryTermsHighlightsForPanel(data);
+  const editEnabled = ['OWNER', 'MAINTAINER'].includes(
```
gabrielshanahan: This should live in a different place. Why not put this in `globalContext/helpers.tsx`, where we already have `useIsAdmin` etc., and implement it in a similar way?

webapp/src/ee/glossary/components/GlossaryTermPreview.tsx
```diff
@@ -105,7 +131,69 @@ export const GlossaryTermPreview: React.VFC<GlossaryTermPreviewProps> = ({
...
+  const handleCancel = (e: React.MouseEvent) => {
+    e.stopPropagation();
+    setIsEditing(false);
+  };
+
+  const handleSave = (e?: React.MouseEvent) => {
+    e?.stopPropagation();
+    if (!preferredOrganization) return;
+
+    saveMutation.mutate(
```
gabrielshanahan, line 174: Shouldn't we display an error (or do something) if this fails?

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
```diff
@@ -72,6 +72,29 @@ interface LanguageRepository : JpaRepository<Language, Long> {
...
+      join organization o on o.id = p.organization_owner_id
+      where $ORGANIZATION_FILTER
+    """,
+    nativeQuery = true,
+  )
+  fun findAllTagsByOrganizationId(
+    organizationId: Long?,
+    projectIds: List<Long>,
+    anyProject: Boolean,
+  ): Set<String>
```
gabrielshanahan, line 96: Nit: It would be much better if this (and the other methods) were strongly typed, i.e. returned some sort of `Tag` value/data class (not the entity, just a normal object). I think it would work if we returned an interface, although I dislike that as that's even more magic in the background.

You can try it and see if it works for data/value classes OOTB, however I suspect it's not doable without configuration changes that go beyond the scope of this PR.

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
```diff
@@ -355,6 +355,17 @@ class LanguageService(
     cacheManager.getCache(Caches.LANGUAGES)?.evict(language.project.id)
   }
 
+  fun getTagsByOrganization(
```
gabrielshanahan: Just for educational purposes (don't fix) - having a method that just delegates to another, almost-identically-named method, is an architecture smell - it hints that something is not right and needlessly complicated. This is not currently fixable without significant changes that would make the code inconsistent and probably do more damage than good, but I just wanted to point it out as something to think about.

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
```diff
@@ -355,6 +355,17 @@ class LanguageService(
...
+    organizationId: Long,
+    projectIds: List<Long>?,
```
gabrielshanahan: Nullable method parameters are booleans in disguise and as such are [usually a code smell](https://alexkondov.com/should-you-pass-boolean-to-functions/). Whenever possible, the responsibility for handling null should be pushed up the call stack, where there is more information, and therefore more context on how to handle the null case (is it a programmer mistake? a data error? invalid data from the client? etc.).

Furthermore, looking at how this is actually used, I don't really understand the need for the nullability in the first place.

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
```diff
@@ -72,6 +72,29 @@ interface LanguageRepository : JpaRepository<Language, Long> {
...
+      join project p on p.id = l.project_id
+      join organization o on o.id = p.organization_owner_id
+      where $ORGANIZATION_FILTER
+    """,
+    nativeQuery = true,
+  )
+  fun findAllTagsByOrganizationId(
+    organizationId: Long?,
+    projectIds: List<Long>,
+    anyProject: Boolean,
```
gabrielshanahan, line 95: Nit: (don't fix, just for educational purposes) boolean method parameters are [usually a code smell](https://alexkondov.com/should-you-pass-boolean-to-functions/). I didn't spend too much time looking at this, but on first glance, my hunch is it should probably be solved by having two different methods - one which filter by project ID, and one which doesn't.

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
gabrielshanahan: This is also true of the other places that return `Long` etc. I'll just keep this one comment, and leave it up to you if you want to spend time on it or not.

gabrielshanahan, line 96: If I'm not mistaken, using a `Set` will cascade into columns in the export appearing in arbitrary order, no? Don't have enough info to know if that's a going to be a problem UX-wise or not, but unless there's a good reason to use `Set`, I would just switch to `List`.

gabrielshanahan (changes requested):
I unfortunately didn't have time to review further than the 3rd commit. My feedback would predominantly be focused on code style, however since we're this close to release, we have very little space to address those issues anyway.

Therefore, I would recommend:
* if this is okayed from the UX/functional POV, and we have reasonable test coverage, let's ship it as is
* in the future, let's please split this into many smaller changes and create separate PRs

webapp/src/ee/glossary/components/GlossaryTermPreview.tsx
Anty0: This shouldn't happen - many things will fail first. If the preferred organization is undefined, the FE won't really show anything. If it is unset on the backend side, the backend will try to set it. There are many places in FE that ignore it. I'd say it could be better to remove the if and let the FE die if it is not set - less code.

Do you have any ideas on how we can handle this better? ^^

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
Anty0: Ye, this is a little bit of an unusual case. I remember having a hard time checking whether `projectIds` are null in the query, so I added a separate boolean for that. Also, there are other queries that optionally filter by some things, so many parameters are optional in repositories in general.

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
Anty0: Since we are collecting language tags for export from multiple queries, I've delegated sorting to the backend. I don't need them sorted here. It would be possible to have a single special query for export that would retrieve all tags in a single request, but it felt like overoptimizing for something that probably won't take much runtime ^^

Honestly, the only reason I went with `Set` is that later in the code, I can `set1 + set2` to get a union, and since the query doesn't sort them, I don't see why not use `Set` for all such queries. On the other hand, we may decide that all queries should sort by something reasonable, even if the sorting is unused, and always use `List`. Might make sense ^^

Anty0: I'll leave it as is, but we can decide later what that would look like in general.

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
Anty0: Ye, most of our services do that - they delegate to the repository if no additional logic is needed to obtain the data.

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
Anty0: I'll split this one into two. It makes more sense in this case.

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
bdshadow: @gabrielshanahan i disagree with you. In Spring it's a common approach. You have three layers - controllers, services and repositories. And that's ok to just user a service as a "proxy" in simple cases. Calling a repository from a controller would be artchtecturally wrong

gabrielshanahan: > Ye, most of our services do that
🎯  Exactly my point

> In Spring it's a common approach
Yes, and one I've come to strongly disagree with. It brings no benefit that I can see, and quite a bit of cost - pointless boilerplate that does nothing and serves no use apart from growing the codebase and increasing complexity and cognitive load. The only reason I could ever see that this is actually used by everyone is "because everyone says so" - that's not merit, that's going with the herd.

Now of course, I could be wrong! It's perfectly possible that there's some huge benefit I'm missing that offsets all this cost, and I'm happy to learn about it if that's the case. But just mandating "thou shalt organize your code around controllers services and repositories because Spring/DDD/whoever says so", and if you don't, "it would be architecturally wrong" is not how we should approach things. In general, mandating "thou shalt X because Y says so" is not how we should approach things.

I'm happy to hear your POV of why this is a good idea, and they'll be plenty of discussions about this soon, but I do want the discussions to be based on merit, not herd behavior.

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
gabrielshanahan: My POV is that branching logic shouldn't be pushed to SQL (and in general, unless absolutely necessary, logic shouldn't be in SQL, period), pretty much for exactly the reasons that the article talks about - a query is equivalent to a method, in that it accepts some parameters and returns a result.

All this is basically an analogy of the unix principle - it's better to have `n` queries doing `1` thing then `1` query doing `n` things. Imho applies to pretty much all "developer constructs", not just queries.

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
bdshadow: there are some things that do not need to be explained. There are standards, rules, the ways frameworks are used. If we want to argue about everything, try to prove everything to everybody, it's a wrong way and it will take too much time. It's like asking why you should stop on red light while driving...  There is a rule to do it. These rules are based on peoples experience...
Of course we can discuss it, but then let's discuss every tiny thing in kotlin and spring (+postgres, redis...). We will burry oursleves in these discussions

backend/data/src/main/kotlin/io/tolgee/repository/LanguageRepository.kt
gabrielshanahan: Ah, I wasn't aware that it was sorted down the road - in that case, that's fine.

webapp/src/ee/glossary/components/GlossaryTermPreview.tsx
gabrielshanahan: No, it just caught my eye. If it's effectively unreachable, then I'm fine with leaving it as-is 👍

Issue #3379, opened 2026-01-11
Batch jobs: Increase chunk size for MachineTranslationChunkProcessor


bdshadow:
Currently `MachineTranslationChunkProcessor.getChunkSize` gives 1, because there was a problem with a bigger number, that when AI gave error on one of the items, the whole chunk failed. We should collect such errors, without failing the whole chunk, and then give it back to user (only for the failed items).

This will alow to increase the chunk size and significantly imrove the performace of the MT translation batch jobs

bdshadow:
closing, implemented

PR #3382, opened 2026-01-13
feat: gh-3379 introduce aggregate exception handling for items in chunk

bdshadow:
## Summary by CodeRabbit

* **New Features**
  * Jobs now aggregate multiple-item failures into a single failure with combined error details and preserved partial successes for coordinated retries.
  * Processors use simplified, fixed chunk sizes for translation batches.

* **Bug Fixes**
  * Retry counts and delays are computed from aggregated error info to avoid premature/over-retries.
  * Concurrency launch tightened to prevent over-starting executions and race conditions.

* **Tests**
  * Expanded test utilities and new tests covering multi-item failures, retry behavior, and concurrency.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobConcurrentLauncher.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobConcurrentLauncher.kt
@@ -6,7 +6,6 @@ import io.tolgee.batch.data.ExecutionQueueItem
 import io.tolgee.component.CurrentDateProvider
 import io.tolgee.configuration.tolgee.BatchProperties
 import io.tolgee.fixtures.waitFor
-import io.tolgee.model.batch.BatchJobChunkExecutionStatus
 import io.tolgee.util.Logging
 import io.tolgee.util.logger
 import io.tolgee.util.trace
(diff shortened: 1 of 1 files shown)


backend/data/src/main/kotlin/io/tolgee/batch/processors/MachineTranslationChunkProcessor.kt
```diff
@@ -60,12 +60,6 @@ class MachineTranslationChunkProcessor(
     request: MachineTranslationRequest,
     projectId: Long?,
   ): Int {
-    projectId ?: throw IllegalArgumentException("Project id is required")
```
JanCizmar, line 63: This should be fixed also in AutoTranslationChunkProcessor. There maybe we might go to lower number to maybe paralelize it better 🤔

bdshadow: made it 3 there

bdshadow: i don't have a strong opinion here, hope 3 is fine

bdshadow:
added here one more commit which fixes the failing test checking the max number of coroutines launched for batch jobs

bdshadow:
so this pr can be merged now

backend/data/src/main/kotlin/io/tolgee/batch/processors/MachineTranslationChunkProcessor.kt
JanCizmar: Hmm, I was thinking about that and I realized that there also is the import, which triggers auto-translation operation. We can consider this later. Let's keep 3 now.

PR #3384, opened 2026-01-16
feat: support branching in SingleKeyView and cleanup

zomp:


JanCizmar:
Hey! Thx for the PR ❤️.

Just my 2 cents: this is definitely not a chore. it's a feat.

bdshadow (approved):

webapp/src/ee/task/components/taskCreate/TaskCreateForm.tsx
```diff
@@ -104,8 +104,8 @@ export const TaskCreateForm = ({
         branch,
       };
       return {
-        url: '/v2/projects/{projectId}/tasks/calculate-scope',
-        method: 'post',
+        url: '/v2/projects/{projectId}/tasks/calculate-scope' as const,
+        method: 'post' as const,
```
dkrizan, line 108: is this necessary ?

zomp: `tsc` raised an issue here because of incompatible types – it took this as regular `string` and could not match it to generated schema endpoints. Adding `const` assured `tsc`, so it is able to use endpoint types. ... Strange thing is there is an issue on `useApiQueries`, but all `useApiQuery` work fine.

PR #3385, opened 2026-01-16
fix: add translation suggestions scope to Edit role

dkrizan:
#3386

## Summary by CodeRabbit

* **New Features**
  * Users with EDIT permissions now have access to translation suggestions functionality, enabling collaborative translation workflows.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/model/enums/ProjectPermissionType.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/enums/ProjectPermissionType.kt
@@ -47,6 +47,7 @@ enum class ProjectPermissionType(
       Scope.KEYS_VIEW,
       Scope.TRANSLATIONS_VIEW,
       Scope.TRANSLATIONS_EDIT,
+      Scope.TRANSLATIONS_SUGGEST,
       Scope.KEYS_EDIT,
       Scope.KEYS_DELETE,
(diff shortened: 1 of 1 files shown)


Anty0:
I think @JanCizmar should look at this. I don't have a great insight into translation suggestions - no idea how it should behave in this situation. I know there was a whole thing about some projects wanting users to be able to suggest changes only, without the ability to directly edit. I'd have to check whether and how this affects functionality.

dkrizan:
> I think @JanCizmar should look at this. I don't have a great insight into translation suggestions - no idea how it should behave in this situation. I know there was a whole thing about some projects wanting users to be able to suggest changes only, without the ability to directly edit. I'd have to check whether and how this affects functionality.

I understand @Anty0 👍

I was looking at it from a simpler perspective: the Edit role is higher than the previous roles, so it should implicitly include all their scopes, including working with suggestions. From that point of view, blocking editing because suggestions exist feels inconsistent.

If we decide that the Edit role should not include this scope, then the current behavior still needs to be fixed somehow, because right now it results in a bug where users cannot edit translations at all when suggestions are present.

But I agree, it makes sense to double-check this with @JanCizmar to ensure it aligns with the intended permissions model.

Anty0:
@dkrizan Definitely +1 on "the current behavior needs to be fixed", just wasn't sure if it won't break anything. ^^

PR #3396, opened 2026-01-19
chore: cleanup and perf improvements of  branching

bdshadow:


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/branching/DefaultBranchCreator.kt
```diff
@@ -21,9 +21,11 @@ class DefaultBranchCreator(
   @Transactional
   fun create(project: Project): Branch {
     val branch = Branch.createMainBranch(project)
-    keyService.getAll(project.id).forEach { it ->
+    // TODO: ideally, must be a simple update by projectId, but this will ignore activity creation
+    keyRepository.getAllByProjectIdLazily(project.id).forEach {
```
dkrizan: actually, skipping activity creation is makes sense, as this is not real update, it justs "migrate" data (I don't want to show it in activity log) .. therefore doing single update is good idea

bdshadow: Aha... Will do

bdshadow: done

dkrizan: thanks a lot !

dkrizan (approved):

PR #3397, opened 2026-01-20
fix: branches backend cosmetic fixes

bdshadow:


backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -4720,10 +4720,7 @@
     </changeSet>
     <changeSet author="danielkrizan (generated)" id="1758202102054-2" runInTransaction="false">
         <sql>
-            create unique index concurrently on public.key (project_id, "name", coalesce(branch_id, -1)) where namespace_id is null;
-        </sql>
-        <sql>
-            create unique index concurrently on public.key (project_id, "name", coalesce(branch_id, -1), namespace_id) where namespace_id is not null;
+            create unique index concurrently on public.key (project_id, branch_id, "name", namespace_id);
```
dkrizan: I’m not sure this change is correct. Because branch_id and namespace_id are nullable, UNIQUE (project_id, branch_id, name, namespace_id) allows duplicates when either is NULL The previous partial indexes prevented this.

bdshadow: @dkrizan thank you
youre right. I got it back. `copying a lot data is not slow` is still passing

PR #3401, opened 2026-01-22
fix: add scheduled job for cleaning old batch jobs

JanCizmar:
## Summary
- Adds `OldBatchJobCleaner` scheduled job that automatically removes old batch jobs to prevent database growth
- SUCCESS/CANCELLED jobs are deleted after 3 days (configurable via `tolgee.batch.completed-job-retention-days`)
- FAILED jobs are deleted after 30 days (configurable via `tolgee.batch.failed-job-retention-days`)
- Uses native SQL with batch deletion (1000 jobs per batch) to avoid database locks on large tables
- Implements distributed locking via `withLockingIfFree` for multi-node safety
- Exposes Prometheus metrics for monitoring (`tolgee_batch_job_cleanup_*`)
- Removes unused `DEBOUNCED` status from `BatchJobStatus` enum

## Test plan
- [x] Integration tests added in `OldBatchJobCleanerTest` covering:
  - Deletion of old SUCCESS/CANCELLED jobs
  - Deletion of old FAILED jobs  
  - Retention of recent jobs within period
  - Protection of RUNNING/PENDING jobs
  - Cleanup of associated chunk executions
  - Multiple batch handling

## Summary by CodeRabbit

* **New Features**
  * Automatic cleanup of old batch jobs with configurable retention, schedule, batch sizing and metrics.
  * Non-blocking lock option to skip work when an operation is already running.

* **Tests**
  * Added comprehensive tests validating cleanup behavior, retention rules, safe deletion of related records, and various job statuses.

* **Chores**
  * Added configuration flags to control job cleanup.
  * Removed an obsolete batch job status value.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/BatchProperties.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/BatchProperties.kt
@@ -43,7 +43,9 @@ class BatchProperties {
   var jobCleanupBatchSize: Int = 1000
 
   @DocProperty(
-    description = "Lock lease time in milliseconds for old job cleanup (to prevent lock expiration during long cleanups)",
+    description =
+      "Lock lease time in milliseconds for old job cleanup " +
+        "(to prevent lock expiration during long cleanups)",
     defaultExplanation = "1 day",
   )
   var jobCleanupLockLeaseTimeMs: Long = 86400000
--- a/backend/data/src/test/kotlin/io/tolgee/security/ratelimit/RateLimitServiceTest.kt
+++ b/backend/data/src/test/kotlin/io/tolgee/security/ratelimit/RateLimitServiceTest.kt
@@ -251,6 +251,7 @@ class RateLimitServiceTest {
 
     override fun <T> withLockingIfFree(
       name: String,
+      leaseTime: java.time.Duration,
       fn: () -> T,
     ): T? {
       TODO("Not yet implemented")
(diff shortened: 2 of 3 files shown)


backend/data/src/main/kotlin/io/tolgee/batch/cleaning/OldBatchJobCleaner.kt
```diff
@@ -0,0 +1,246 @@
...
+      ).setParameter("jobIds", jobIds)
+      .executeUpdate()
+  }
+
+  private fun deleteBatchJobs(jobIds: List<Long>) {
+    entityManager
+      .createNativeQuery(
+        """
+        DELETE FROM tolgee_batch_job
+        WHERE id IN :jobIds
```
bdshadow, line 212: This one will fail on first run with
`Caused by: org.postgresql.util.PSQLException: PreparedStatement can have at most 65,535 parameters.`

Because there are definitely more than this number of jobs to delete

bdshadow:
As i mentioned before, ideally, the table must be partitoned and old partitions are dropped. Otherwise, the delete query can last for too long

backend/data/src/main/kotlin/io/tolgee/batch/cleaning/OldBatchJobCleaner.kt
JanCizmar: There is a configurable limit per each iteration it defaults to 1000.

JanCizmar: Oh, looks like you were almost right. Just on wrong place... :D

```
        UPDATE activity_revision
        SET batch_job_id = NULL, batch_job_chunk_execution_id = NULL
        WHERE batch_job_id IN :jobIds
        OR batch_job_chunk_execution_id IN :executionIds
```

PR #3402, opened 2026-01-22
fix: e2e & co. fixes

zomp:


webapp/src/fixtures/getPermissionTools.ts
```diff
@@ -23,7 +24,11 @@ export const getPermissionTools = (permissions: PermissionModel) => {
       return satisfiesLanguageAccess(permissions, scope, languageId);
     },
     satisfiesPermissionWithBranching(scope: Scope) {
-      return satisfiesPermissionWithBranching(scopes, scope);
+      return satisfiesPermissionWithBranching(
+        scopes,
+        scope,
+        useBranchEditAccess()
```
dkrizan, line 30: What is the reason to move it here ?

zomp: The reason is, we need to remove `useBranchEditAccess` from the other file –  `permissions.ts`. Fixtures dir is meant for utils which can be used from elsewhere. Because of that, there should not be imports from elsewhere to prevent cyclic dependencies and to prevent other issues. There was such other issue in `permissions.ts` – it is imported from `e2e` (Cypress) project. As such, it cannot handle typescript aliases from the `webapp` project (`e2e` project does not know what `tg.views` in `import {useBranchEditAccess} from 'tg.views/projects/translations/context/services/useBranchEditAccess'` is, the same applies on its imports). The simplest solution is removing `tg.views...` from  `permissions.ts` and using it in `getPermissionTools.ts` which is the only place that invokes it.

zomp: TLDR `tg.views` (or something which imports such aliases) cannot be imported in `permissions.ts`, so I moved it here.

dkrizan: Ok, makes sense, thanks for a exhausting explanation 👍

PR #3421, opened 2026-01-27
feat: count unique keys by name across branches

bdshadow:
@marketachalupnikova @ZuzanaOdstrcilova adding you here to to be on the same page regarding billing, when branching is finished. Please check the comments in the `OrganizationStatsServiceTest`, which describes it well, when a key is considered unique or not
```kt
    // - key1 (main only) = 1 unique key
    // - key2 (main and feature) = 1 unique key
    // - key3 (feature only) = 1 unique key
    // - key4 with namespace1 (main and feature) = 1 unique key
    // - key4 with namespace2 (main only) = 1 unique key
    // - key4 without namespace (main only) = 1 unique key
```

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationStatsService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationStatsService.kt
@@ -23,7 +23,7 @@ class OrganizationStatsService(
     return entityManager
       .createQuery(
         """
-        select count(k) from Key k
+        select count(distinct k.name, k.namespace) from Key k
         where k.project.id = :projectId and k.project.deletedAt is null
         """.trimIndent(),
       ).setParameter("projectId", projectId)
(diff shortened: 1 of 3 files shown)


backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationStatsService.kt
```diff
@@ -23,7 +23,7 @@ class OrganizationStatsService(
     return entityManager
       .createQuery(
         """
-        select count(k) from Key k
+        select count(distinct concat(k.name, ':', coalesce(cast(k.namespace.id as string), ''))) from Key k
```
dkrizan: I am bit afraid this may (unlikely) cause a collision if key.name contains ":".
Wouldn’t it be better to switch to a native query instead and use a multi-column DISTINCT (name, namespace_id) ? If I am not mistaken, JPQL does not support that.

JanCizmar:
Have you tested this behavior also with billing enabled in billing repo?

backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationStatsService.kt
bdshadow: you're right. Updated

bdshadow:
@JanCizmar thanks for requesting it. Indeed, there are problems with it

PR #3424, opened 2026-01-28
fix: rework project properties access

Anty0:
This changeset refactors the Tolgee backend to centralize property access through a single `TolgeeProperties` holder instead of injecting individual `*Properties` classes via Spring's `@ConfigurationProperties` scanning.

## Summary

- Switch from `@ConfigurationPropertiesScan` to `@EnableConfigurationProperties(TolgeeProperties::class)`; remove `@ConfigurationProperties` from the sub-property classes
- Replace direct sub-property injections with `TolgeeProperties` injection + delegated getters (holder pattern), including the webhook consumers
- Add `@NestedConfigurationProperty` to nested config so Spring binding/metadata keeps working
- Restore config-properties validation (`SelfValidatingProperties` / `validateNestedProperties`) and content-delivery cache-purging wiring that the bean refactor had dropped
- Keep method-param injection in `initInitialUser`
- Rename `rateLimit` → `rateLimits` and `llmProperties` → `llm`; `MtServiceType` now uses lambda property getters instead of class references
- Behavior change: machine-translation services' default enablement now follows configured values instead of a previous unconditional default
- Tests: replace the manual per-test `TolgeeProperties` snapshot/restore boilerplate with a `ConfigurationPropertiesResetListener` that auto-restores every `io.tolgee` `@ConfigurationProperties` root before/after each test (Jackson in-place merge preserving nested-instance identity; covers Tolgee/Ee/Billing roots)

## Summary by CodeRabbit

* **Refactor**
  * Configuration access unified into a single central properties holder; behavior preserved but internal wiring simplified.
* **Behavior change**
  * Machine translation services' default enablement now follows configured values instead of a previous unconditional default.
* **Tests**
  * Test suites improved with clearer setup/teardown and fewer scattered property spies/mocks to reduce cross-test leakage and increase reliability.


bdshadow:
please, don't merge it before branching

bdshadow:
and what's the benefit of it?
For example, now instead of just injecting the `aws`, you need to have 
```kotlin
private val awsMachineTranslationProperties get() = tolgeeProperties.machineTranslation.aws
```

Before this refactoring it was also possible to inject consolidated `tolgeeProperties` and access all nested layers like this. So this pr just is just prohibits it, leading to more code and injecting whole bunch of properties when you need only a single one.
Maybe i'm wrong, but i don't see it to be beneficial.

Anty0:
@bdshadow The issue is that we currently have (at least) two instances of each `*Properties`. I've been burned by this before - you override something in tests through `TolgeeProperties`, and it doesn't work. You don't know why until you realize the component is accessing the (let's say) AWS properties directly, not through `TolgeeProperties`, so you need to modify it there, or even worse, both, if different code paths use the property from `TolgeeProperties`. I wanted to solve this. (This doesn't really affect unit tests, fortunately, as we create the properties there directly, so we create only a single instance of everything. On the contrary, e2e tests can only set properties on the `TolgeeProperties`, because that's what `PropertiesController` does. I feel like this is going to bite us.)

Other than this, we currently have two things broken, which I want to fix this way:
- Rate limits: You may notice that their `prefix` is set to `rate-limits` while the property is named `rateLimit`. On Tuesday, I realized that this means `TolgeeProperties` are looking for `rate-limit`, while directly accessing `RateLimitProperties` will look for `rate-limits`. The fact that this is even possible is, in my opinion, completely insane. Contrary to most other properties, we access some rate limits through `TolgeeProperties`, so they don't get set at all (they stay the default).
- LLM Properties: Same issue. Fortunately, we always access them directly, so there are no configuration issues, but if anyone were to use `TolgeeProperties` to access them, they would just get default values.

I'm not saying we have to fix it like this, but it felt like the best way to make sure we never run into the same issue again.

I can also just fix the names of the two broken properties and leave it be for now. ^^
(I'll go and do that anyway, as I want it fixed sooner rather than later; this rework can wait after branching.)

JanCizmar (dismissed):
Thanks. 👍 I would merge it after branching.

bdshadow:
@Anty0 
>The fact that this is even possible is, in my opinion, completely insane.

if jakarta validation like `@NotNull`, `@NotEmpty` is put on the properties, then it will fail on the start of the app, when thre is such a mistype. So it's easy to fix. And must be fixed like this.

Another thing is about accessing the properties. This seems weird, need to check. But probably this is because fields inside `TolgeeProperties` are not injected ,they are created by a default constructor. I might be wrong here, but just annotating it with `NestedConfigurationProperty` and removing initizalization must be enough to solve this problem. (another option is just to declare a constructor of the `TolgeeProperties`, where all fields will be injected by spring)

bdshadow (changes requested):
I'll dare to request changes :) Sorry :)
Feel free to ping me

Anty0:
@bdshadow

> if jakarta validation like @NotNull, @NotEmpty is put on the properties

You mean they won't have default values anymore? I'm not sure I understand this one. ^^

> Another thing is about accessing the properties

Yes, please check and let me know (after branching). If there is a different way we can make sure the `*Properties` instances won't be created multiple times, I'm all ears ^^

bdshadow:
@Anty0 my comment was based on my previos experience in java and nested static classes. You can how it works here: https://docs.spring.io/spring-boot/3.5/reference/features/external-config.html#features.external-config.typesafe-configuration-properties.validation

In Tolgee there are too many properties. Moving them to nested classes of `TolgeeProperties` would just explode it and it won't be useful. And it will even bigger change.

Also the default values are set right in the koltin code, and before i had it all in properties files. In that case validation perfectly worked. Here it is different.

So sorry for delaying this pr, but i'm glad i found it all

bdshadow (dismissed):

Anty0:
The failing test needs to be fixed outside of this PR. (We probably tipped over a limit of how many translations we can have statically typed by TypeScript or something like that.)

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/batch/BatchJobTestBase.kt
```diff
@@ -52,21 +53,33 @@ class BatchJobTestBase {
...
-    whenever(awsMock.accessKey).thenReturn("mock")
-    whenever(awsMock.secretKey).thenReturn("mock")
+    machineTranslationProperties.aws.defaultEnabled = false
+    machineTranslationProperties.aws.accessKey = "mock"
+    machineTranslationProperties.aws.secretKey = "mock"
+  }
 
-    whenever(machineTranslationProperties.aws).thenReturn(awsMock)
+  fun tearDown() {
+    tolgeeProperties.internal.fakeMtProviders = fakeMtProvidersBefore
```
dkrizan: I understand the advantages of this PR, but it also has a downside: the developer has to take care of resetting the properties’ values in the tearDown method.

Could we introduce a generic reset mechanism instead of per-test teardown? E.g., snapshot the whole `TolgeeProperties` before each test and restore it after ?

Anty0: Implemented auto-restore of all configuration properties between each test.

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/machineTranslation/MachineTranslationProperties.kt
```diff
@@ -1,18 +1,23 @@
...
 )
 open class MachineTranslationProperties(
+  @NestedConfigurationProperty
   var google: GoogleMachineTranslationProperties = GoogleMachineTranslationProperties(),
+  @NestedConfigurationProperty
   var aws: AwsMachineTranslationProperties = AwsMachineTranslationProperties(),
+  @NestedConfigurationProperty
   var deepl: DeeplMachineTranslationProperties = DeeplMachineTranslationProperties(),
+  @NestedConfigurationProperty
   var azure: AzureCognitiveTranslationProperties = AzureCognitiveTranslationProperties(),
```
dkrizan, line 19: As far as I can tell, on main the key is `tolgee.machine-translation.azurecognitive.*` (from the injected AzureCognitiveTranslationProperties prefix), but after the refactor it binds from the field name `machineTranslation.azure` → `tolgee.machine-translation.azure.*.` So existing self-hosted configs using `azurecognitive` would just silently stop working? And `azurecognitive` only seems to live on as a `@DocProperty` string now.

If that's right, maybe rename the field back to `azurecognitive` to keep it backwards compatible ?

PR #3427, opened 2026-01-29
chore: remove redundant liquibase migrations for branching feature

bdshadow:
Remove migrations that were added and later reverted during development:
- archived_at column on branch table
- unique constraints on origin_branch_id and (project_id, name)
- branch_key_meta_snapshot_tags table and its foreign keys

Also fix unique index to use deleted_at instead of removed archived_at.


dkrizan (approved):
Thanks 🙌

PR #3428, opened 2026-01-29
fix: implement round-robin scheduling for batch job chunks

JanCizmar:
## Summary

- Large batch jobs were monopolizing all worker coroutines, causing smaller jobs to wait hours
- Implements fair round-robin scheduling that rotates between jobs
- Each job gets one chunk processed before any job gets a second chunk

**Before (FIFO):**
```
Queue: [A1, A2, A3, B1, B2, C1]
Poll order: A1, A2, A3, B1, B2, C1  # Job A monopolizes workers
```

**After (Round-Robin):**
```
Queue: [A1, A2, A3, B1, B2, C1]
Poll order: A1, B1, A2, B2, A3, C1  # All jobs progress fairly
```

## Changes

- Add `lastServedJobId` field to `BatchJobChunkExecutionQueue` to track rotation state
- Add `pollRoundRobin()` method that alternates between jobs
- Update `BatchJobConcurrentLauncher` to use `pollRoundRobin()` instead of `poll()`
- Add test verifying round-robin behavior (9 job switches in 10 polls = perfect alternation)

## Test plan

- [x] New test `round-robin scheduling prioritizes second job over more chunks from first job` passes
- [x] All 14 `BatchJobConcurrentIntegrationTest` tests pass
- [x] All batch job tests pass (39 tests total)

## Summary by CodeRabbit

* **New Features**
  * Queue polling now uses round‑robin distribution for fairer chunk scheduling across concurrent jobs.

* **Tests**
  * Added a comprehensive concurrent batch-job test suite covering concurrency, fairness, recovery, locking, progress, cancellation, parallelism and stress scenarios.
  * Removed a legacy concurrent integration test.

* **Refactor**
  * Small change to enable subclassing of a batch job state storage component.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
@@ -227,6 +227,10 @@ class BatchJobChunkExecutionQueue(
    * Polls using round-robin across jobs for fair distribution.
    * Each job gets one chunk processed before any job gets a second chunk.
    * This prevents large jobs from monopolizing all worker coroutines.
+   *
+   * Thread-safety: This method handles concurrent modifications gracefully.
+   * If another thread removes an item between finding and removing it,
+   * we verify the removal succeeded and retry if needed.
    */
   fun pollRoundRobin(): ExecutionQueueItem? {
     if (queue.isEmpty()) {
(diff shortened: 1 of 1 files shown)


backend/data/src/main/kotlin/io/tolgee/batch/BatchJobChunkExecutionQueue.kt
```diff
@@ -216,6 +223,54 @@ class BatchJobChunkExecutionQueue(
...
+    for (i in jobIds.indices) {
+      val jobIndex = (startIndex + i) % jobIds.size
+      val targetJobId = jobIds[jobIndex]
+
+      // Find and remove first item for this job
+      val iterator = queue.iterator()
+      while (iterator.hasNext()) {
+        val item = iterator.next()
+        if (item.jobId == targetJobId) {
+          iterator.remove()
```
bdshadow: if there are 2 concurrent threads, the other might have already called `remove()` here, or `poll`, then the behavior of the `remove` here would be unexpected. 

Or am i missing smth and there is no concurrency here and it's just called in one thread? Seems like that, it's called only from the `BatchJobConcurrentLauncher`,  but then why making `lastServedJobId` volatile? The only case is multthreading...

Please, explain, what am i missing?

bdshadow (approved):

PR #3430, opened 2026-01-29
feat: main branch for the existing keys, tasks, imports, language stats

bdshadow:
## Summary by CodeRabbit

* **New Features**
  * New projects now automatically get a default "main" branch at creation.

* **Bug Fixes**
  * Keys, tasks and imports are now required to be associated with a branch to avoid missing references.
  * Default-branch resolution now throws an error when no default exists to prevent ambiguous behavior.

* **Chores**
  * Database migration added to backfill and enforce non-null branch references for existing data.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/component/demoProject/DemoProjectCreator.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/demoProject/DemoProjectCreator.kt
@@ -9,6 +9,7 @@ import io.tolgee.model.Language
 import io.tolgee.model.Organization
 import io.tolgee.model.Project
 import io.tolgee.model.Screenshot
+import io.tolgee.model.branching.Branch
 import io.tolgee.model.enums.TranslationCommentState
 import io.tolgee.model.enums.TranslationState
 import io.tolgee.model.key.Key
@@ -73,6 +74,7 @@ class DemoProjectCreator(
         this@apply.organizationOwner = organization
         this.description = "This is a demo project of a packing list app"
       }
+    Branch.createMainBranch(project)
     projectService.save(project)
     setAvatar(project)
     project
--- a/backend/data/src/main/kotlin/io/tolgee/component/demoProject/DemoProjectCreator.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/demoProject/DemoProjectCreator.kt
@@ -192,6 +192,7 @@ class DemoProjectCreator(
             this.pluralArgName = pluralArgName
           }
           this@apply.project = this@DemoProjectCreator.project
+          this@apply.branch = this@DemoProjectCreator.project.getDefaultBranch()
         }
       keyService.save(key)
       key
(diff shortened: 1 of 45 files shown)


backend/data/src/main/kotlin/io/tolgee/component/demoProject/DemoProjectCreator.kt
```diff
@@ -73,6 +74,7 @@ class DemoProjectCreator(
         this@apply.organizationOwner = organization
         this.description = "This is a demo project of a packing list app"
       }
+    Branch.createMainBranch(project)
```
dkrizan, line 77: I am missing `persist` on branch. You can use `DemoProjectCreator` which already calls `persist`.

backend/data/src/main/kotlin/io/tolgee/development/DbPopulatorReal.kt
```diff
@@ -123,6 +124,7 @@ class DbPopulatorReal(
     project.name = projectName
     project.organizationOwner = organization
     project.slug = slugGenerator.generate(projectName, 3, 60) { true }
+    Branch.createMainBranch(project)
```
dkrizan, line 127: same as comment above

backend/data/src/main/kotlin/io/tolgee/development/DbPopulatorReal.kt
```diff
@@ -150,6 +152,7 @@ class DbPopulatorReal(
     val project = Project()
     project.name = projectName
     project.organizationOwner = organization
+    Branch.createMainBranch(project)
```
dkrizan, line 155: same as comment above

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5035,4 +5035,70 @@
...
+            )
+            WHERE ls.branch_id IS NULL
+             AND NOT EXISTS (
+              SELECT 1 FROM language_stats ls2
+                JOIN language l ON l.id = ls.language_id
+                JOIN branch b ON b.project_id = l.project_id AND b.name = 'main'
+              WHERE ls2.language_id = ls.language_id AND ls2.branch_id = b.id
+             );
+        </sql>
+    </changeSet>
```
dkrizan, line 5128: In this case we can make `Key.branch` not-null, right ?

dkrizan (changes requested):

backend/data/src/main/kotlin/io/tolgee/component/demoProject/DemoProjectCreator.kt
bdshadow: but this is the `DemoProjectCreator`. And i don't see `persist` here

backend/data/src/main/resources/db/changelog/schema.xml
bdshadow: yes

bdshadow: i guess the same is for tasks and imports. However language_stats seems differnt

bdshadow: added those constraints

PR #3431, opened 2026-01-29
feat: add indexes on activity_revision and batch_job timestamps

gabrielshanahan:
## Summary
- Add B-tree index on `activity_revision.timestamp` for efficient time-range queries on the activity audit trail
- Add B-tree index on `tolgee_batch_job.created_at` for efficient time-range queries on batch job history

These are standard indexes on timestamp columns of append-mostly tables and carry negligible write overhead.

## Test plan
- [ ] Verify Liquibase migration applies cleanly on a fresh database
- [ ] Verify migration applies cleanly on existing production database
- [ ] Confirm indexes exist after migration: `\di IDXcg7y08d599p7jxdebbxwavak0` and `\di IDX71c7yr8a185u5jsdyndom5fa7`

### DIFF
--- a/backend/data/src/main/resources/db/changelog/schema.xml
+++ b/backend/data/src/main/resources/db/changelog/schema.xml
@@ -4653,4 +4653,14 @@
             <column name="storage_json" type="JSONB"/>
         </addColumn>
     </changeSet>
+    <changeSet author="gabriel.shanahan (generated)" id="1769694787155-164">
+        <createIndex indexName="IDXcg7y08d599p7jxdebbxwavak0" tableName="activity_revision">
+            <column name="timestamp"/>
+        </createIndex>
+    </changeSet>
+    <changeSet author="gabriel.shanahan (generated)" id="1769694787155-144">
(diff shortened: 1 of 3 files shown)


backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -4653,4 +4653,14 @@
...
+    <changeSet author="gabrielshanahan" id="1738142400000-1">
+        <createIndex indexName="idx_activity_revision_timestamp" tableName="activity_revision">
+            <column name="timestamp"/>
+        </createIndex>
+    </changeSet>
+    <changeSet author="gabrielshanahan" id="1738142400000-2">
+        <createIndex indexName="idx_batch_job_created_at" tableName="tolgee_batch_job">
+            <column name="created_at"/>
+        </createIndex>
+    </changeSet>
```
Anty0: I think it would be better to add them to the `BatchJob` and `ActivityRevision` classes (you can see other indexes are defined there) and generate the changelog entries using `./gradlew diffChangeLog`.

Btw, I don't know why, but Claude Code always ends the `id` with a bunch of zeros, which at least makes it easy to recognize that the changelog wasn't generated by the `diffChangeLog`:3

PR #3445, opened 2026-02-02
feat: Add OpenTelemetry tracing spans for batch job operations

JanCizmar:
## Summary
- Added OpenTelemetry instrumentation to improve observability of batch job operations
- Added `@WithSpan` annotations to key methods handling database operations on activity tables
- Added span attributes to `handleItem` for better trace filtering

### Dependencies
- Added `io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.10.0`

### Instrumented Components
**OldBatchJobCleaner** - Batch job cleanup operations:
- All cleanup methods including DELETE/UPDATE operations on activity tables

**BatchJobActivityFinalizer** - Activity revision merging:
- Methods handling UPDATE/DELETE operations on activity tables during job finalization

**ChunkProcessingUtil** - Chunk processing:
- Main chunk processing logic

**BatchJobActionService** - Chunk execution:
- Database locking and chunk execution handling
- Added span attributes: `batch.job.id`, `batch.chunk.execution.id`, `project.id`

## Benefits
- Properly named spans for all batch job operations visible in tracing tools
- Easy filtering and identification of specific jobs, chunks, and projects
- Better visibility into DELETE/UPDATE operations on activity tables (activity_revision, activity_modified_entity, activity_describing_entity)

## Testing
- Code compiles successfully
- ktlint formatting applied

## Summary by CodeRabbit

* **Chores**
  * Enhanced system observability and monitoring infrastructure: Added comprehensive tracing and instrumentation support across batch processing operations, enabling improved visibility into system behavior, performance diagnostics, and troubleshooting capabilities. Updated build dependencies to support modern observability standards throughout the batch job lifecycle.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobActionService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/BatchJobActionService.kt
@@ -1,7 +1,7 @@
 package io.tolgee.batch
 
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
-import io.opentelemetry.api.GlobalOpenTelemetry
+import io.opentelemetry.api.trace.Span
 import io.opentelemetry.api.trace.StatusCode
 import io.opentelemetry.instrumentation.annotations.WithSpan
 import io.sentry.Sentry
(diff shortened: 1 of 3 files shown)


backend/data/src/main/kotlin/io/tolgee/batch/BatchJobActionService.kt
```diff
@@ -53,76 +56,91 @@ class BatchJobActionService(
...
+    val span = tracer.spanBuilder("BatchJobActionService.handleItem").startSpan()
+    val scope = span.makeCurrent()
+
     try {
-      val execution =
-        catchingExceptions(executionItem) {
-          executeInNewTransaction(transactionManager) { transactionStatus ->
+      span.setAttribute("batch.job.id", batchJobDto.id.toString())
+      span.setAttribute("batch.chunk.execution.id", executionItem.chunkExecutionId.toString())
+      batchJobDto.projectId?.let { span.setAttribute("project.id", it.toString()) }
```
gabrielshanahan: Pro tip: tell claude to instead use [otel baggage](https://opentelemetry.io/docs/concepts/signals/baggage/) and it's spring support to set up these global things like project ID and organisation ID globally. It should be a fairly easy change, and we'll get these attributes everywhere.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobActionService.kt
```diff
@@ -53,76 +56,91 @@ class BatchJobActionService(
     executionItem: ExecutionQueueItem,
     batchJobDto: BatchJobDto,
   ) {
-    val coroutineContext = coroutineContext
-    var retryExecution: BatchJobChunkExecution? = null
+    val tracer = GlobalOpenTelemetry.getTracer("tolgee")
```
gabrielshanahan: Why are you doing it like this? I would expect just `@WithSpan` (`class.method` name is the default) and then `val span = Span.current();` 🤔

gabrielshanahan (changes requested):

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
```diff
@@ -1,5 +1,6 @@
 package io.tolgee.batch
 
+import io.opentelemetry.instrumentation.annotations.WithSpan
```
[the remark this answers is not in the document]
bdshadow: i think it's also worth looking at `opentelemetry-spring-boot-starter` , so that we can include all needed library at once.  And maybe even manage its versions by bom: `opentelemetry-instrumentation-bom`

https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/getting-started/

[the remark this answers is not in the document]
JanCizmar: Good point.

backend/data/src/main/kotlin/io/tolgee/batch/cleaning/OldBatchJobCleaner.kt
```diff
@@ -56,16 +58,19 @@ class OldBatchJobCleaner(
...
     deleteJobsOlderThan(listOf("SUCCESS", "CANCELLED"), cutoffDate, "completed")
   }
 
+  @WithSpan
   fun cleanupFailedJobs() {
     val cutoffDate = currentDateProvider.date.addDays(-batchProperties.failedJobRetentionDays)
     deleteJobsOlderThan(listOf("FAILED"), cutoffDate, "failed")
   }
 
+  @WithSpan
```
bdshadow, line 73: it won't work on private methods, because it is based on spring aop proxies: https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/annotations/

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobProjectLockingManager.kt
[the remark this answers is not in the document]
gabrielshanahan: Please definitelly no. Instrumentation via java agent is much superior, I specifically elected to do it like this. There is no reason to use the speing boot starter that I can see.

backend/data/src/main/kotlin/io/tolgee/batch/cleaning/OldBatchJobCleaner.kt
gabrielshanahan: We are not instrumenting using the spring boot starter, we are using the otel javasgent. I'm pretty sure this should work just fine

JanCizmar: Yep. According to the docs, Gabe is right.

backend/data/src/main/kotlin/io/tolgee/batch/BatchJobActionService.kt
JanCizmar: In this specific case, after discussions with Claude, I would rather keep it using span, not baggage. Looks like there is no more global context for chunk exectuion.

PR #3448, opened 2026-02-03
chore: add a way for running local observability stack with Tolgee running outside of docker

Anty0:
## Summary by CodeRabbit

* **New Features**
  * Opt-in run task to launch the app with a local observability/tracing stack.
  * Helper to prepare the OpenTelemetry agent for local/IDE use.
  * Dedicated local observability Docker Compose for development (tracing, metrics, logging, DB, Redis).

* **Documentation**
  * Updated setup docs with two clear options: full Docker stack or local app with Docker-based observability.
  * Clarified commands, verification steps, and troubleshooting for traces and dashboards.

### DIFF
--- a/backend/app/build.gradle
+++ b/backend/app/build.gradle
@@ -310,44 +310,39 @@ tasks.register('copyOtelAgent', Copy) {
 }
 
 // Run with local observability stack (docker-compose.local-observability-stack-dev.yaml)
+// Drop-in replacement for bootRun: supports --args and all other BootRun options.
 // See: docs/observability/
-tasks.register('bootRunWithObservability') {
+tasks.register('bootRunWithObservability', org.springframework.boot.gradle.tasks.run.BootRun) {
     group = 'application'
     description = 'Runs Tolgee with OpenTelemetry tracing, using the local Docker observability stack'
 
+    // Inherit classpath and main class from bootRun (configured by Spring Boot plugin)
+    classpath = bootRun.classpath
+    mainClass.set(bootRun.mainClass)
+
+    // OTEL configuration
+    environment 'OTEL_JAVAAGENT_ENABLED', 'true'
+    environment 'OTEL_SERVICE_NAME', 'tolgee-platform'
+    environment 'OTEL_EXPORTER_OTLP_ENDPOINT', 'http://localhost:4318'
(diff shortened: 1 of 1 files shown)


backend/app/build.gradle
```diff
@@ -298,6 +298,49 @@ bootRun {
     }
 }
 
+// Copy OTEL agent to a predictable location for IntelliJ usage
+// See: docs/observability/
+tasks.register('copyOtelAgent', Copy) {
```
gabrielshanahan, line 306: I don't see this actually used anywhere - it should imho either be called as part of something, mentioned in the docs, or removed

Anty0: It's used by the IntelliJ IDEA runner in the other repo, though I'm not sure myself if it's required.

docs/local-observability-stack/1_Setting_Up_the_Tracing_Stack.md
```diff
@@ -30,15 +46,52 @@ This starts:
...
+- **Loki** - stores and queries logs
+- **Promtail** - ships logs from containers to Loki
+- **Prometheus** - stores metrics generated from traces
+- **OpenTelemetry Collector** - receives traces and forwards them
+- **PostgreSQL** and **Redis** - Tolgee dependencies
+
+**Run Tolgee locally with tracing enabled:**
+
+```bash
+./gradlew server-app:bootRunWithObservability --args='--spring.profiles.active=dev'
```
dkrizan, line 81: `--args='--spring.profiles.active=dev'` is probably not needed since this is already set in task `bootRunWithObservability` 🤔

Anty0: Nice catch! ^^
Though I'd prefer removing it from the `bootRunWithObservability` so it works the same as `bootRun`. Does that make sense? Otherwise, I'd change `bootRun` directly. I think both should use the same profile by default.

dkrizan: Agree, lets remove it from `bootRunWithObservability` 🤝

Anty0: I had a feeling the current setup would not pass the args, so I modified it a bit. Hope it still works :3

PR #3455, opened 2026-02-05
fix: Resilient Redis cache reads for rate limiting

gabrielshanahan:
## Summary

- **Exclude actuator endpoints from rate limit filters** — health probes (`/actuator/health`) no longer go through `GlobalIpRateLimitFilter`/`GlobalUserRateLimitFilter`, preventing crash loops when Redis cache entries are corrupted
- **Add `ResilientCacheAccessor` component** — catches `RedisException` on direct `cache.get()` calls (not covered by `TolgeeCacheErrorHandler` which only handles `@Cacheable`), evicts bad entries, returns `null` (cache miss)
- **Integrate into `RateLimitService`** — corrupted Bucket entries are treated as cache misses instead of 500 errors

### Root cause (v3.154.0 crash loop)

`Bucket` class gained `strikeCount`/`lastStrikeAt` fields → old Kryo-serialized entries in Redis couldn't deserialize → `KryoBufferUnderflowException` → health checks hit rate limit filter → HTTP 500 → K8s restarts pod → crash loop.

### Defense in depth

| Layer | What it does | Protects against |
|-------|-------------|-----------------|
| Actuator bypass | Skips rate limiting for `/actuator/**` | Health probe failures from any cache issue |
| ResilientCacheAccessor | Catches `RedisException`, evicts, returns null | 500 errors on API requests from corrupted cache |

## Test plan

- [x] `ResilientCacheAccessorTest` — 5 tests (normal get, null return, RedisException caught + eviction, non-Redis propagation)
- [x] `RateLimitServiceTest` — 15 tests including new `corrupted cache entry is treated as cache miss`
- [ ] Full backend test suite (`server-app:runStandardTests`)
- [ ] Manual: start with Redis, corrupt a `rateLimits` entry, verify graceful handling

## Summary by CodeRabbit

* **New Features**
  * Improved cache resilience: corrupted cache entries are evicted and treated as misses to prevent runtime errors.
  * Rate limiting now skips actuator endpoints so health and metrics remain accessible (including when an application context path is set).

* **Tests**
  * Added tests validating cache resilience, eviction on corruption, null-handling, and exception propagation.
  * Expanded rate limiting tests to verify actuator endpoint exemptions and context-path handling.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/component/ResilientCacheAccessor.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/ResilientCacheAccessor.kt
@@ -1,0 +1,58 @@
+/**
+ * Copyright (C) 2023 Tolgee s.r.o. and contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.tolgee.component
+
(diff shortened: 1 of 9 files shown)


JanCizmar:
Can we remove the `CacheSchemaCleaner` now?

PR #3457, opened 2026-02-06
fix: stop escaping percent signs in Compose Multiplatform XML export

### LINKED ISSUE #3368: Compose Multiplatform export adds escaping characters when they are not supported by compose multiplatform resource handliing
**Describe the bug**
Hey we are running into an issue where text such as "25% Commission Active" gets exported as "25\% Commission Active" for Compose Multiplatform xml, this is not correct as compose multiplatform is able to deal with percentage signs. If we exported this out as json it instead becomes "25%% Commission Active" which also seems wrong? At any case this is 100% wrong for compose multiplatform and it is quite annoying to every time we export have to manually clean this up

**To Reproduce**
1. Add a key that contains value "25% Commission Active"
2. Export the key as compose multiplatform multiplatform .xml
3. The xml now has the key value as "25\% Commission Active"

**Expected behavior**
Compose Multiplatform does not use character escaping and ends up being rendered. This behaviour is documented here https://kotlinlang.org/docs/multiplatform/compose-multiplatform-resources-usage.html#strings "You don't need to escape special XML characters like "@" or "?" as you do for Android "

**Screenshots**

**Versions and environment**
- Tolgee Platform version: latest
- Environment: Through tolgee web interface

Anty0:
## Summary
- Fixes #3368 — Compose Multiplatform XML export was incorrectly escaping `%` as `\%`
- Compose Multiplatform does not use backslash-escaping for special characters (unlike Android XML)
- Added `escapePercentWithBackslash` parameter to `MobileStringEscaper`, gated on `isAndroid` in `TextToXmlResourcesConvertor` — following the same pattern used for 4 other escaping flags

## Summary by CodeRabbit

* **New Features**
  * Added configurable percent sign escaping behavior for Android and Compose XML resource exports.

* **Bug Fixes**
  * Improved handling of percent signs in exported text content to better align with XML standards.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/MobileStringEscaper.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/MobileStringEscaper.kt
@@ -14,6 +14,7 @@ class MobileStringEscaper(
   private val escapeQuotes: Boolean,
   // Android need 'u' and iOS 'U'
   private val utfSymbolCharacter: Char,
+  private val escapePercentWithBackslash: Boolean = true,
 ) {
   private enum class State {
     DEFAULT,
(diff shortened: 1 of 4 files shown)


amusejfo:
@Anty0 This now adds double %% for Compose Multiplatform instead of `\%`, so while the escape is gone, it is adding extra % for no reason instead

Anty0:
@amusejfo Weird 🤔
I think there is even a specific test for that.
I'll check more closely to see if I've missed anything. Sorry about that!

PR #3459, opened 2026-02-06
feat: store msgctxt from Gettext (.po) files

### LINKED ISSUE #3053: Store `msgctxt` from Gettext (.po) files
**Problem**

Currently, we ignore the `msgctxt` from .po files. However, users who use this field have no workaround to get their project working with Tolgee.

Also for Unreal integration, this would be super useful, since it's also used there.

**Solution**

- [ ] Import the msgctxt as a park for key name `<msgctxt><separator><msgid>`. 
- [ ] Highlight it some simple way in the UI. 
- [ ] Make it editable. Maybe we can put it to advanced tab of key edition in the UI, because usually this will be imported from outside?

**Appetite**
- 1 week

**No goes**
- We don't have to make the UI super fancy. It would be much beneficial if we support it anyhow. 

*Notes*

More context about the separator: https://chatgpt.com/share/6818b7a9-9218-800b-9993-362d3eb38bc3

Anty0:
## Summary
- Parse and store `msgctxt` from PO files instead of ignoring it with a warning
- Combine `msgctxt` + `\u0004` (EOT) separator + `msgid` as the key name, following the GNU gettext standard
- Store raw `msgctxt` in key custom metadata (`_poFileMsgCtxt`)
- Export `msgctxt` line before `msgid` when the custom value is present
- Display msgctxt visually in the translations UI with a dimmed prefix and `|` separator

Closes #3053

## New Features

* Added full support for PO file message contexts (msgctxt). Translations with context information can now be properly imported, stored, and exported while maintaining all context metadata
* Translation keys with message context are now visually distinguished in the interface, making it easier to identify and manage contextual translations

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/po/contsants.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/po/contsants.kt
@@ -1,3 +1,5 @@
 package io.tolgee.formats.po
 
 const val PO_FILE_MSG_ID_PLURAL_CUSTOM_KEY = "_poFileMsgIdPlural"
+const val PO_FILE_MSG_CTXT_CUSTOM_KEY = "_poFileMsgCtxt"
+const val PO_MSGCTXT_SEPARATOR = "\u0004"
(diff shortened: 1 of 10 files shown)


JanCizmar:
Q: Why do you store it also in the meta field?

sybbear:
Are we able to progress with this? Very important feature and actually a blocker for those who use contexts

Anty0:
Hi, @sybbear! ^^
Sorry, I didn't have enough time to finish this during the last cooldown period. It's one of those smaller tasks we handle between main development cycles. I'll try to look at it during the next cooldown period—either this week or in a little over a month. I had to spend most of this cooldown on work that overflowed from the development cycle, so I won't have much time for small tasks this cycle. Sorry about that!

sybbear:
@Anty0 Ok, hopefully it will be implemented soon :)
Meanwhile namespaces (domain+context), seem to do the work

Anty0:
Superseeded by #3694

PR #3461, opened 2026-02-06
fix: improve handling of disabled accounts

### LINKED ISSUE #3276: Improve handling of disabled accounts
**Describe the bug** Currently, disabled accounts remain in the database, but the system doesn't properly check for them during login or sign-up. Because the UserAccount model enforces a unique constraint on username for non-deleted users, re-registering a previously disabled user causes a constraint violation that isn't handled gracefully.

**Expected behavior** Users should be directed to the organization owner or server administrator to have their disabled account restored.

Anty0:
## Summary
- Disabled accounts are now properly detected during login, sign-up, and third-party auth flows
- Login returns a specific `user_account_disabled` error instead of generic "bad credentials"
- Sign-up detects disabled accounts gracefully instead of causing a database constraint violation
- Added `findActiveOrDisabled()` repository method that finds non-deleted users regardless of disabled status
- Frontend maps the new error code to a user-friendly message directing users to contact their org owner or server admin

Closes #3276

## Summary by CodeRabbit

* **New Features**
  * Disabled user accounts now produce a specific "user_account_disabled" error across sign-up, login, JWT, OAuth2 and personal access token flows; UI shows a corresponding localized message.

* **Tests**
  * Added tests validating authentication is rejected for disabled accounts across multiple authentication methods.

* **API**
  * Error schema updated to include the new "user_account_disabled" error code.


Anty0:
Needs to be ironed out with @dkrizan in relation to case-insensitive usernames work in progress.

backend/data/src/main/kotlin/io/tolgee/service/security/UserCredentialsService.kt
```diff
@@ -20,6 +20,12 @@ class UserCredentialsService(
   ): UserAccount {
     val userAccount = userAccountService.findActive(username)
     if (userAccount == null) {
+      // Reporting the disabled state before the password is verified lets an unauthenticated caller
+      // tell a disabled account apart from a non-existent one (user enumeration). Accepted tradeoff:
+      // sign-up already reveals whether an email is taken, so this exposes no extra capability.
+      userAccountService.findActiveOrDisabled(username)?.let {
```
dkrizan: I think we should find the user in the first call (even the disabled one), and check if it is disabled after the credentials verification - this way pre-auth visitors cannot abuse this endpoint to look for a disabled accounts. Workflow:

1. find user by email (even disabled)
2. if user is null -> bad credentials
3. if found, continue, check credentials
4. after the credentials are verified, check if user is disabled, if so, throw the `Message.USER_ACCOUNT_DISABLED`

Anty0: I was thinking about that, but I was worried it would add complexity to the flow - every login path would need to end with a check "is user disabled," and if we miss that, the user would be able to log in even when disabled.

I believe there is a filter that won't let such a user send any authenticated requests anyway, so the impact would be minimal, still worth considering...

Since both Jan and you flagged this option, I'll try switching the implementation to it, and we'll see how it looks once implemented before deciding. ^^

dkrizan: it's a general concern, but for this specific change it's contained to the native email and password path

backend/app/src/test/kotlin/io/tolgee/controllers/PublicControllerTest.kt
```diff
@@ -128,6 +131,50 @@ class PublicControllerTest : AbstractControllerTest() {
...
+    val dto2 =
+      SignUpDto(
+        name = "Another User",
+        password = "bbbbbbbbb",
+        email = "disabled@test.com",
+      )
+    performPost("/api/public/sign_up", dto2)
+      .andIsBadRequest
+      .andAssertError
+      .hasCode("user_account_disabled")
```
dkrizan: how about `Message.USER_ACCOUNT_DISABLED.code`

backend/data/src/main/kotlin/io/tolgee/service/security/UserCredentialsService.kt
Anty0: True, it is a very simple change if I do it just for this path :3

dkrizan (approved):

PR #3466, opened 2026-02-10
feat: add branch support to CDN content delivery configs

dkrizan:
## Summary

- Links CDN configs to branches via a foreign key (`branch_id`) instead of the plain `filterBranch` string column
- Adds branch selector to the CDN config dialog when branching is enabled
- Shows branch name chip on CDN list items
- Cleans up CDN configs when a branch is deleted
- Migrates existing configs to the default branch during branching enablement
- Feature guard rejects CDN create/update with `filterBranch` when BRANCHING feature is disabled

## Test plan

- [x] All 12 `ContentDeliveryConfigBranchingTest` tests pass, including:
  - Create/update CDN config with specific branch
  - Create without branch defaults to default branch
  - Non-existent branch returns 404
  - Delete branch cascades to CDN configs
  - Feature guard rejects create when branching not enabled
  - Feature guard rejects update when branching not enabled

### DIFF
--- a/ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/EeSubscriptionProviderImplTest.kt
+++ b/ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/EeSubscriptionProviderImplTest.kt
@@ -55,6 +55,8 @@ class EeSubscriptionProviderImplTest : AbstractSpringTest() {
 
   @AfterEach
   fun cleanup() {
+    schedulingManager.cancelAll()
+    eeSubscriptionServiceImpl.delete()
     eeProperties.checkPeriodInMs = oldCheckPeriodProperty
   }
 


backend/data/src/main/kotlin/io/tolgee/service/contentDelivery/ContentDeliveryConfigService.kt
```diff
@@ -214,4 +223,17 @@ class ContentDeliveryConfigService(
   fun save(config: ContentDeliveryConfig): ContentDeliveryConfig {
     return contentDeliveryConfigRepository.save(config)
   }
+
+  fun deleteAllByBranchId(
```
bdshadow, line 228: there are multiple `delete` operations, it must be `@Transactional`

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5057,4 +5057,20 @@
...
+    <changeSet author="danielkrizan (generated)" id="1770751028716-2">
+        <createIndex indexName="IDXmpg39yyj1lqn91onyyb64471v" tableName="content_delivery_config">
+            <column name="branch_id"/>
+        </createIndex>
+    </changeSet>
+    <changeSet author="danielkrizan (generated)" id="1770751028716-3">
+        <addForeignKeyConstraint baseColumnNames="branch_id" baseTableName="content_delivery_config" constraintName="FK5jtc226unswiodyqp8r4isds6" deferrable="false" initiallyDeferred="false" referencedColumnNames="id" referencedTableName="branch" validate="true"/>
+    </changeSet>
+    <changeSet author="danielkrizan (generated)" id="1770751028716-4">
+        <dropColumn columnName="filter_branch" tableName="content_delivery_config"/>
```
bdshadow, line 5074: please add a proper `<rollback></rollback>`

bdshadow (changes requested):

backend/data/src/main/resources/db/changelog/schema.xml
dkrizan: the column is actually never used, but I added rollback, just in case

PR #3246, opened 2025-09-17
feat: branching

dkrizan:

PR #3470, opened 2026-02-14
fix: sort export translations by key name (Sentry TOLGEE-BACKEND-3ED)

bdshadow:
The StructureModelBuilder requires translations sorted alphabetically by key name path, but the database collation may return them in a different order. Sort in GenericStructuredFileExporter.prepare() to ensure correct ordering regardless of database collation.

Likely triggered by the branching feature (#3246, merged Feb 6) which added a LEFT JOIN on Branch to the export query. The additional join changed the query execution plan, surfacing a collation mismatch between PostgreSQL's locale-aware sorting and Java's byte-by-byte string comparison for keys with punctuation (e.g. "profile.my-goals" vs "profile.my-goals.plan_other").

## Summary by CodeRabbit

* **Bug Fixes**
  * Export validation now detects and prevents collisions between plural key suffixes and existing keys during file export operations. Users receive a clear error message when conflicts are detected, specifying the involved keys and suffix to enable quick resolution.


webapp/src/i18n/en.json
```diff
@@ -739,6 +738,7 @@
   "cannot_change_your_own_access_tooltip": "Cannot revoke access to yourself",
   "cannot_delete_base_language_message": "Cannot delete base language",
   "cannot_delete_branch_with_children": "Cannot delete this branch because other branches were created from it.",
+  "export_key_plural_suffix_collision": "Export failed: plural key \"{pluralKey}\" with suffix \"_{suffix}\" collides with existing key \"{collidingKey}\". Rename one of the keys to resolve the conflict.",
```
Anty0, line 741: Shouldn't be part of the changeset.

PR #3486, opened 2026-02-23
feat: key archiving

bdshadow:
## Summary by CodeRabbit

* **New Features**
  * Trash for keys: searchable, branch-aware Trash view with per-item restore and permanent delete; trash banner, counter, batch bar, and navigation.
  * Batch jobs for restoring and hard-deleting keys via new API endpoints and UI progress; automatic purge of trashed keys after 7 days.
  * UI components/hooks: TrashPage, TrashRow, TrashBatchBar, TrashBanner, useTrashCount, UserSearchSelect.

* **Behavior**
  * Keys now support soft-delete (deletedAt/deletedBy) with restore and hard-delete workflows; namespaces preserved while referenced by trashed keys.

* **Tests**
  * Broad unit/integration/e2e coverage for soft-delete, restore, hard-delete, purge, permissions, import, branching and UI flows.


bdshadow:
@coderabbitai  review, please

webapp/src/i18n/en.json
dkrizan, line 1: this file should not be changed, I guess claude did that

backend/data/src/main/kotlin/io/tolgee/service/key/KeyTrashPurgeScheduler.kt
```diff
@@ -0,0 +1,71 @@
...
+    } while (batch.hasNext())
+
+    if (totalPurged > 0) {
+      logger.info("Total purged {} expired trashed keys older than {}", totalPurged, cutoffDate)
+    }
+  }
+
+  companion object {
+    private const val PURGE_LOCK_NAME = "key_trash_purge_lock"
+    private const val RETENTION_DAYS = 7
```
dkrizan: `RETENTION_DAYS = 7` is times in changes

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/keys/KeyTrashController.kt
```diff
@@ -0,0 +1,140 @@
...
+  @RequiresProjectPermissions([Scope.KEYS_VIEW])
+  @AllowApiAccess
+  fun list(
+    @ParameterObject
+    @SortDefault("deletedAt")
+    pageable: Pageable,
+    @ParameterObject
+    @ModelAttribute
+    params: TranslationFilters,
+  ): PagedModel<*> {
```
dkrizan: probably `PagedModel<TrashedKeyWithTranslationsModel>` ?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/keys/KeyTrashController.kt
```diff
@@ -0,0 +1,140 @@
...
+        projectId = projectHolder.project.id,
+        languages = languages.toSet(),
+        pageable = pageable,
+        params = params,
+      )
+    val keyIds = data.content.map { it.keyId }
+    val screenshotsByKeyId =
+      if (keyIds.isNotEmpty()) screenshotService.getScreenshotsForKeys(keyIds) else emptyMap()
+    val assembler =
+      TrashedKeyWithTranslationsModelAssembler(
```
dkrizan: ` screenshotModelAssembler`, `languageService`, `avatarService` could be injected in `TrashedKeyWithTranslationsModelAssembler` (which should be `@Component`),

webapp/src/views/projects/translations/trash/TrashRow.tsx
```diff
@@ -0,0 +1,453 @@
...
+        <StyledCheckbox
+          size="small"
+          checked={selected}
+          onChange={onToggle}
+          data-cy="trash-row-checkbox"
+        />
+        <KeyCellContent keyName={data.name} description={data.description} />
+        {data.screenshots?.length > 0 && (
+          <StyledScreenshots>
+            {data.screenshots.map((sc: any) => {
```
dkrizan: we should avoid using `any` if possible .. since those types should be well-defined in apiSchema

webapp/src/views/projects/translations/trash/TrashRow.tsx
dkrizan, line 1: this file is pretty long, would be nice to split it to more components

webapp/src/views/projects/translations/trash/TrashPage.tsx
```diff
@@ -0,0 +1,527 @@
...
+
+  const handleSelectAll = useCallback(() => {
+    if (allPageSelected) {
+      setSelectedKeys([]);
+    } else {
+      setSelectedKeys(pageKeyIds);
+    }
+  }, [allPageSelected, pageKeyIds]);
+
+  const handleRestore = useCallback(() => {
```
dkrizan, line 211: all three useCallback wrappings here create a new function on every render causing `TrashRow` to re-render unnecessarily on every data refresh

this might be a better choice, but need to be tested
```
const queryClient = useQueryClient();

// useCallback is meaningful here — TrashRow is React.memo'd, passed to every row
const handleRestore = useCallback(() => {
  setSelectedKeys([]);
  invalidateUrlPrefix(queryClient, '/v2/projects/{projectId}/keys/trash');
}, [queryClient]);

const handleDelete = useCallback(() => {
  setSelectedKeys([]);
  invalidateUrlPrefix(queryClient, '/v2/projects/{projectId}/keys/trash');
}, [queryClient]);

// useCallback not needed — TrashBatchBar is not memoized
const handleBatchFinished = () => {
  setSelectedKeys([]);
  invalidateUrlPrefix(queryClient, '/v2/projects/{projectId}/keys/trash');
};
```

webapp/src/views/projects/translations/trash/TrashPage.tsx
dkrizan, line 1: again big component, should be splitted

webapp/src/component/UserSearchSelect/UserSearchSelect.tsx
dkrizan, line 1: this seems to be a copy of `AssigneeSearchSelect`, this should be unified

webapp/src/component/UserSearchSelect/UserSearchSelectPopover.tsx
dkrizan, line 1: same, duplicate of `AssigneeSearchSelectPopover`, it should be easy to unify

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/keys/KeyTrashController.kt
```diff
@@ -0,0 +1,140 @@
...
+      translationViewDataProvider.getData(
+        projectId = projectHolder.project.id,
+        languages = languages.toSet(),
+        pageable = pageable,
+        params = params,
+      )
+    val keyIds = data.content.map { it.keyId }
+    val screenshotsByKeyId =
+      if (keyIds.isNotEmpty()) screenshotService.getScreenshotsForKeys(keyIds) else emptyMap()
+    val assembler =
```
dkrizan: too much logic which in my opinion should be grouped in service

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/keys/KeyTrashController.kt
```diff
@@ -0,0 +1,140 @@
...
+  @Operation(summary = "Restore a trashed key")
+  @RequestActivity(ActivityType.KEY_RESTORE)
+  @RequiresProjectPermissions([Scope.KEYS_CREATE])
+  @AllowApiAccess
+  fun restore(
+    @PathVariable keyId: Long,
+  ): KeyModel {
+    val key = keyService.restoreKey(projectHolder.project.id, keyId)
+    val view =
+      KeyView(
```
dkrizan, line 105: assembler for this as well ?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/keys/KeyTrashController.kt
```diff
@@ -0,0 +1,140 @@
...
+  @Operation(summary = "Permanently delete a trashed key")
+  @RequestActivity(ActivityType.KEY_HARD_DELETE)
+  @RequiresProjectPermissions([Scope.KEYS_DELETE])
+  @AllowApiAccess
+  fun permanentlyDelete(
+    @PathVariable keyId: Long,
+  ) {
+    val projectIds = keyService.getSoftDeletedProjectIdsForKeyIds(listOf(keyId))
+    if (projectIds.isEmpty() || projectIds.single() != projectHolder.project.id) {
+      throw NotFoundException(Message.KEY_NOT_FOUND)
```
dkrizan: logic can be in service too + error message should be `Message.KEY_NOT_FROM_PROJECT`

e2e/cypress/e2e/translations/softDeleteKeys.cy.ts
```diff
@@ -0,0 +1,224 @@
...
+    );
+
+    // Translations view should only show key2
+    visitTranslations(project.id);
+    waitForGlobalLoading();
+    cy.contains('key1').should('not.exist');
+    cy.contains('key2').should('be.visible');
+
+    // Trash should contain both trashed key1 entries
+    cy.visit(`${HOST}/projects/${project.id}/translations/trash`);
```
dkrizan: you should create E2 compound for Translations Trash section and use it (see e.x.: Branching tests for reference) - it also prevents calling duplicatiated funs like 
`cy.visit(`${HOST}/projects/${project.id}/translations/trash`);`

e2e/cypress/e2e/translations/softDeleteKeys.cy.ts
```diff
@@ -0,0 +1,224 @@
...
+    // key1 should not be in translations either
+    visitTranslations(project.id);
+    waitForGlobalLoading();
+    cy.contains('key1').should('not.exist');
+    cy.contains('key2').should('be.visible');
+  });
+
+  it('soft-deleted key is excluded from task and restored key reappears', () => {
+    // 1. Create a translate task with both keys
+    v2apiFetch(`projects/${project.id}/tasks`, {
```
dkrizan: It think, in general, E2E tests should replicate real user behavior, meaning the requests should be triggered by interacting with the UI (clicking, typing, navigating), not by calling the APIs directly from the test.

Using API calls for core behavior under test, it would be better to drive the scenario through the UI and assert on the visible outcome. Otherwise, this becomes closer to an integration/API test rather than a true e2e test and may reduce confidence that the actual user flow works correctly.

e2e/cypress/e2e/translations/softDeleteKeys.cy.ts
```diff
@@ -0,0 +1,224 @@
...
+
+describe('Soft delete keys', () => {
+  let project: ProjectDTO;
+  let key1Id: number;
+  let key2Id: number;
+  let englishLanguageId: number;
+
+  beforeEach(() => {
+    login().then(() => {
+      createProject({
```
dkrizan: initial data should be creating using `TestData`

backend/data/src/main/kotlin/io/tolgee/service/key/KeyService.kt
```diff
@@ -332,36 +339,116 @@ class KeyService(
...
+
+  @Transactional
+  fun restoreKeys(
+    ids: Collection<Long>,
+    projectId: Long,
+  ) {
+    val keys = keyRepository.findSoftDeletedByIdsAndProjectId(ids, projectId)
+    keys.forEach { key ->
+      val branch = key.branch?.name
+      val existing = find(projectId, key.name, key.namespace?.name, branch)
```
dkrizan: N+1 problem

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/QueryGlobalFiltering.kt
```diff
@@ -288,4 +290,13 @@ class QueryGlobalFiltering(
       )
     }
   }
+
+  private fun filterDeletedByUserId() {
+    if (!params.filterDeletedByUserId.isNullOrEmpty()) {
+      val deletedByJoin = queryBase.root.join(Key_.deletedBy, JoinType.LEFT)
```
dkrizan: When `trashed = true` and `filterDeletedByUserId` is set, two separate LEFT JOIN user_account clauses are generated - one from `addDeletedBySelection()` in QueryBase, one from `filterDeletedByUserId()` in QueryGlobalFiltering. They should share the same join.

I am afraid this might be issue in other JOINs too, probably branches too :/ I am not sure about the performance impact this can cause, but is definitely worth checking.

webapp/src/views/projects/translations/trash/TrashPage.tsx
bdshadow: React Query memoizes refetch internally, so the useCallback dependency [trashLoadable.refetch] doesn't change, and TrashRow (which is React.memo'd) won't re-render unnecessarily. However, the suggested invalidateUrlPrefix approach has a separate, real benefit: it invalidates all queries matching the /keys/trash prefix, including the useTrashCount query on the translations page.

I will refactor it

webapp/src/component/UserSearchSelect/UserSearchSelectPopover.tsx
bdshadow: it's leftover, i have unified it, but then rolled it back serveral times. Thanks. Will solve

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/keys/KeyTrashController.kt
bdshadow: `KeyController.edit` uses the same approach. I don't think it's needed here

PR #3488, opened 2026-02-24
refactor: make limit checkers lazy and add unit tests

dkrizan:
## Summary

- **Bug fix**: unlimited detection used `< 0`; changed to `<= 0` so `limit == 0` is also treated as unlimited
- **Optimization**: `GenericLimitChecker` now accepts the required count as a lambda, so the DB fetch is skipped entirely when the limit is unlimited or the event is non-additive (early exit before invoking the provider)
- **Tests**: add pure-Kotlin unit tests (no Spring context) for `GenericLimitChecker`, `StringsLimitChecker`, `KeysLimitChecker`, and `SeatsLimitChecker` — 19 tests total covering all branching paths including the unlimited-detection bug fix

## Test plan

- [ ] `./gradlew :ee-test:test --tests "io.tolgee.ee.unit.limitsAndReporting.*"` — all 19 pass

## Summary by CodeRabbit

* **Refactor**
  * Unified and simplified limit-checking flow for keys, seats and strings into a consistent provider-based validation.

* **Bug Fixes**
  * Corrected detection/validation of unlimited and zero limits.
  * Fixed usage increase calculation on entity updates so text additions/removals are counted correctly.

* **Tests**
  * Added comprehensive unit tests covering generic, keys, seats and strings scenarios (fixed, PAYG, null/zero).

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/util/entityPreCommitEventUsageUtil.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/util/entityPreCommitEventUsageUtil.kt
@@ -25,6 +25,8 @@ fun EntityPreCommitEvent.getUsageIncreaseAmount(): Long {
           val currentText = entity.text
           if (previousText.isNullOrEmpty() && !currentText.isNullOrEmpty()) {
             return 1
+          } else if (!previousText.isNullOrEmpty() && currentText.isNullOrEmpty()) {
+            return -1
           }
         }
       }


backend/data/src/main/kotlin/io/tolgee/util/entityPreCommitEventUsageUtil.kt
```diff
@@ -3,11 +3,36 @@ package io.tolgee.util
...
+    }
+
+    is OnEntityPreUpdate -> {
+      val entity = this.entity
+      if (entity is Translation) {
+        val textIndex = this.propertyNames?.indexOf("text") ?: -1
+        if (textIndex != -1) {
+          val previousText = this.previousState?.get(textIndex) as? String
+          val currentText = entity.text
+          if (previousText.isNullOrEmpty() && !currentText.isNullOrEmpty()) {
```
bdshadow, line 26: shouldn't iе be `-1` for `!previousText.isNullOrEmpty() && currentText.isNullOrEmpty()`?

dkrizan: probably yes, good catch !

PR #3492, opened 2026-02-25
feat: map-based LLM provider config with pricing dialog

JanCizmar:
## Summary
- Add map-based LLM provider configuration support
- Add pricing info dialog for server-configured LLM providers showing token costs (MT credits) and estimated translation costs for 1,000 strings
- Add descriptive text labels above pricing tables

## Test plan
- [ ] Verify pricing dialog opens from info icon on server providers
- [ ] Verify token prices and estimates display correctly
- [ ] Verify EUR column shows when billing is active
- [ ] Run billing e2e tests (see billing PR)

## Summary by CodeRabbit

* **New Features**
  * Provider pricing UI: info button, modal, compact token-price and cost-estimate tables with EUR conversion.
  * Pricing displayed only when billing is enabled.
  * Provider configuration supports per-provider defaults for easier setup.
  * Provider schema includes optional input/output token price fields.

* **Tests**
  * Unit tests covering provider-defaults merging.
  * E2E test ensuring pricing is hidden when billing is disabled.

* **Documentation**
  * Updated API workflow and examples.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/LlmPropertiesService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/LlmPropertiesService.kt
@@ -89,8 +89,8 @@ class LlmPropertiesService(
     listEntry.attempts?.let { base.attempts = it }
     listEntry.tokenPriceInCreditsInput?.let { base.tokenPriceInCreditsInput = it }
     listEntry.tokenPriceInCreditsOutput?.let { base.tokenPriceInCreditsOutput = it }
-    // Non-nullable fields: override only if different from default
-    if (listEntry.type != LlmProviderType.OPENAI) {
+    // Non-nullable fields: override only if different from Spring default
+    if (listEntry.type != LlmProvider.TYPE_DEFAULT) {
       base.type = listEntry.type
     }
     if (listEntry.maxTokens != LlmProvider.MAX_TOKENS_DEFAULT) {
--- a/webapp/src/ee/llm/OrganizationLLMProviders/LlmProviderPricingInfo.tsx
+++ b/webapp/src/ee/llm/OrganizationLLMProviders/LlmProviderPricingInfo.tsx
@@ -2,7 +2,10 @@ import { Box, IconButton, styled, Tooltip } from '@mui/material';
 import { useState } from 'react';
 import { InfoCircle } from '@untitled-ui/icons-react';
 import { useTranslate } from '@tolgee/react';
-import { LlmProviderPricingDialog } from './LlmProviderPricingDialog';
+import {
+  LlmProviderPricingDialog,
+  ProviderWithPricing,
+} from './LlmProviderPricingDialog';
 
 const StyledAction = styled(Box)`
   display: flex;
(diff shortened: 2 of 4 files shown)


webapp/src/ee/llm/OrganizationLLMProviders/LlmProviderPricingInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+const StyledAction = styled(Box)`
+  display: flex;
+  align-items: center;
+  align-self: stretch;
+  justify-self: stretch;
+  color: ${({ theme }) => theme.palette.text.secondary};
+  padding-right: 8px;
+`;
+
+type ProviderWithPricing = {
```
dkrizan: duplicated type, same is in `webapp/src/ee/llm/OrganizationLLMProviders/LlmProviderPricingDialog.tsx`

backend/data/src/main/kotlin/io/tolgee/service/LlmPropertiesService.kt
```diff
@@ -40,4 +41,61 @@ class LlmPropertiesService(
...
+    listEntry.model?.let { base.model = it }
+    listEntry.deployment?.let { base.deployment = it }
+    listEntry.reasoningEffort?.let { base.reasoningEffort = it }
+    listEntry.format?.let { base.format = it }
+    listEntry.priority?.let { base.priority = it }
+    listEntry.attempts?.let { base.attempts = it }
+    listEntry.tokenPriceInCreditsInput?.let { base.tokenPriceInCreditsInput = it }
+    listEntry.tokenPriceInCreditsOutput?.let { base.tokenPriceInCreditsOutput = it }
+    // Non-nullable fields: override only if different from default
+    if (listEntry.type != LlmProviderType.OPENAI) {
```
dkrizan: are you sure about this two conditions ? if listEntry has type = ANTHROPIC, defaults type = OPENAI, result still stays ANTHROPIC, same for maxTokens

webapp/src/ee/llm/OrganizationLLMProviders/LlmProviderPricingInfo.tsx
JanCizmar: Fixed — exported `ProviderWithPricing` from `LlmProviderPricingDialog.tsx` and importing it in `LlmProviderPricingInfo.tsx`.

backend/data/src/main/kotlin/io/tolgee/service/LlmPropertiesService.kt
JanCizmar: The logic is actually correct for the case you describe: if `listEntry.type = ANTHROPIC` and defaults `type = OPENAI`, then `ANTHROPIC != TYPE_DEFAULT` is true, so `base.type = ANTHROPIC` — the list entry wins as expected.

The condition exists because `type` and `maxTokens` are non-nullable with Spring Boot defaults, so we can't distinguish "user explicitly set OPENAI" from "Spring filled in the default". The heuristic: if the value equals the Spring default, assume it wasn't explicitly set and keep the defaults-map value.

Known limitation: you can't override a non-OPENAI default back to OPENAI via the providers list (use `provider-defaults` for that). Extracted the defaults into constants (`TYPE_DEFAULT`, `MAX_TOKENS_DEFAULT`) so they stay in sync with the field defaults.

PR #3503, opened 2026-03-05
feat: limit concurrent non-exclusive batch jobs per project

bdshadow:
Prevents a single project with many non-exclusive jobs (e.g. AUTO_TRANSLATE) from monopolizing all worker coroutines and starving other projects.

Adds configurable tolgee.batch.max-non-exclusive-jobs-per-project (default: 2) that limits how many distinct non-exclusive jobs from one project can run concurrently on each instance.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for batch job concurrency behavior, including per-project job limits and cross-project execution scenarios.

* **Chores**
  * Introduced a new configuration property to control the maximum number of concurrent non-exclusive batch jobs per project instance (default: 2).

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/BatchProperties.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/BatchProperties.kt
@@ -54,4 +54,14 @@ class BatchProperties {
     defaultExplanation = "30 seconds",
   )
   var cancellationTimeoutMs: Long = 30000
+
+  @DocProperty(
+    description =
+      "Maximum number of concurrently running non-exclusive jobs per project per instance. " +
+        "This prevents a single project with many non-exclusive jobs (e.g. AUTO_TRANSLATE) " +
+        "from monopolizing all worker coroutines and starving other projects. " +
+        "The limit is enforced locally on each Tolgee instance, not globally across the cluster.",
+    defaultExplanation = "2",
+  )
+  var maxNonExclusiveJobsPerProject: Int = 2
 }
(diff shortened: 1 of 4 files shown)


bdshadow:
@coderabbitai review

JanCizmar (approved):

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/BatchProperties.kt
```diff
@@ -54,4 +54,14 @@ class BatchProperties {
...
+
+  @DocProperty(
+    description =
+      "Maximum number of concurrently running non-exclusive jobs per project per instance. " +
+        "This prevents a single project with many non-exclusive jobs (e.g. AUTO_TRANSLATE) " +
+        "from monopolizing all worker coroutines and starving other projects. " +
+        "The limit is enforced locally on each Tolgee instance, not globally across the cluster.",
+    defaultExplanation = "2",
+  )
+  var maxNonExclusiveJobsPerProject: Int = 2
```
JanCizmar, line 66: I would set it to 10 by default. We have 20 concurrent threads per instance on production, so 10 should be OK.

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/BatchProperties.kt
```diff
@@ -54,4 +54,14 @@ class BatchProperties {
...
   )
   var cancellationTimeoutMs: Long = 30000
+
+  @DocProperty(
+    description =
+      "Maximum number of concurrently running non-exclusive jobs per project per instance. " +
+        "This prevents a single project with many non-exclusive jobs (e.g. AUTO_TRANSLATE) " +
+        "from monopolizing all worker coroutines and starving other projects. " +
+        "The limit is enforced locally on each Tolgee instance, not globally across the cluster.",
+    defaultExplanation = "2",
```
JanCizmar, line 64: `defaultExplanation` is for explaining the default. It makes no sense to explain it with the same value as the default is.

JanCizmar (changes requested):

Issue #3170, opened 2025-07-09
Add configurable metadata to S3 content delivery


chcatf:
**Is your feature request related to a problem? Please describe.**
We use a self-hosted S3 bucket (non-AWS) that syncs with a CDN and the TTL of any stored object is controlled by the Cache-Control field of each object's metadata.

**Describe the solution you'd like**
An input added to the content delivery configuration form to set custom metadata on the uploaded translation files such as from here https://github.com/tolgee/tolgee-platform/blob/main/backend/data/src/main/kotlin/io/tolgee/component/fileStorage/S3FileStorage.kt#L44

**Describe alternatives you've considered**
Alternatives are all manual/cron solutions of overwriting existing objects with the proper metadata.

Anty0:
Hi! Thanks for the report! Sorry for the delay.
It seems like a pretty simple addition. If anyone wants to help out, feel free to create a PR!

Issue #2504, opened 2024-10-02
Publishing to Azure storage doesn't set correct MIME type


cfmano:
**Describe the bug**
When publishing exported translation files to Azure BLOB storage, they are not exported with correct MIME type, therefore they automatically fall back to application/octet-stream type.

**To Reproduce**
1. Setup use of Azure BLOBs
2. To go project -> Developer settings -> Content delivery
3. Publish translations
4. Log into azure portal, find blob file and check overview parameters

**Expected behavior**
Files should have correct MIME types included:
- .json - application/json
- .yaml - application/yaml
- ...

**Screenshots**
![image](https://github.com/user-attachments/assets/a6548d21-9727-468d-8477-e155cee9dba1)

**Versions and environment**
- Tolgee Platform version: 3.75.1
- Environment: Docker version

JanCizmar:
Hey! Thanks for reporting. Can you please explain why this is a problem? Thanks!

cfmano:
There are basically two main reasons:
1. Azure CDN/FrontDoor's default caching and comprimation bahaviour includes only known types of files, which means that you have explicitly setup caching of these files (using different rules, becuase adding generic application/octet-stream type is discouraged).
2. Browsers usually have undefined behaviour when handling octet-stream files - most of the time it's handled as generic file meaning it downloads and saves it, instead of preperly displaying it. Side part of undefined handling behaviour is potential problem with encoding, BOM and malformed texts.

Side (or software specific) problems can include:
- some BE frameworks don't handle octet-stream as JSON (e.g. fastify)
- MSSQL.net client handles only well formed JSON files with proper MIME-type
- Azure Logic Apps handle octet-stream as binary format instead of text

Some other examples could be googled out.

Anty0:
Reopening because of a related support request on Slack: https://tolgeecommunity.slack.com/archives/C03A0RQR62H/p1737961060464379

madhunandan-gs:
Same issue on Amazon S3 as well.

![Image](https://github.com/user-attachments/assets/935c68f4-298a-40a6-a9b6-1be7d4ddf493)

dievri:
Same issue on Google Cloud Storage as well.  

<img width="579" height="107" alt="Image" src="https://github.com/user-attachments/assets/dd671263-9994-42d6-9fdc-5bb79a7737bf" />

No caching, no compression by default with CloudCDN

Diggsey:
Same issue here 😄

PR #3507, opened 2026-03-08
feat: add custom HTTP metadata support for content delivery configs

dievri:
Related: #2504, #3170

## What this does

Adds support for configuring custom HTTP metadata (e.g. `Cache-Control`, `Content-Type`) per content delivery config. The metadata is set on every S3/GCS `PutObject` call when publishing translation files.

## Motivation

Two practical problems this solves:

1. **CDN caching** — there is currently no way to set `Cache-Control` on published objects, so CDNs cannot be instructed how to cache them per delivery config
2. **CDN compression** — translation files are uploaded as `application/octet-stream` by default, so CDNs (e.g. Google Cloud CDN) skip gzip/brotli compression. Setting `Content-Type: application/json` via this feature fixes that

## Changes

**Backend**
- New `customMetadata` jsonb column on `content_delivery_config` (Liquibase migration included)
- `FileStorage.storeFile()` interface extended with optional `metadata` parameter
- `S3FileStorage`: well-known HTTP headers (`Cache-Control`, `Content-Type`, `Content-Disposition`, `Content-Encoding`, `Content-Language`) are routed to their proper `PutObjectRequest` builder fields — so they appear as actual HTTP response headers, not as `x-amz-meta-*`

**Frontend**
- Key-value metadata editor in the Edit Content Delivery dialog

## Test plan

- [ ] Add `Cache-Control: public, max-age=300` to a content delivery config, publish, verify the header appears on the object (not as `x-amz-meta-cache-control`)
- [ ] Add `Content-Type: application/json`, verify CDN serves it with correct content type and compresses it
- [ ] Verify Liquibase migration runs cleanly from a previous version

## Summary by CodeRabbit

* **New Features**
  * Added custom metadata configuration for content delivery files, allowing users to set object metadata (e.g., Cache-Control headers) on uploaded storage objects for S3-compatible storages.
  * Added UI controls for managing custom metadata key-value pairs in the content delivery configuration dialog.


backend/data/src/main/kotlin/io/tolgee/component/fileStorage/S3FileStorage.kt
```diff
@@ -51,6 +57,26 @@ open class S3FileStorage(
...
   }
 
+  private fun applyMetadata(
+    b: software.amazon.awssdk.services.s3.model.PutObjectRequest.Builder,
+    metadata: Map<String, String>,
+  ) {
+    val customMetadata = mutableMapOf<String, String>()
+    for ((key, value) in metadata) {
+      when (key.lowercase()) {
+        "cache-control" -> b.cacheControl(value)
```
JanCizmar, line 67: I wonder, why we cannot treat all the metadata the same. Why do we need this special cases?

JanCizmar (changes requested):
Hello! Thanks a lot for the PR ❤️

Would you be open to add some tests (backend integration + E2E)?

Issue #3480, opened 2026-02-19
Performance issue: Batch job processing stalled with 115K pending chunks filling the queue


Jbbouille:
**Describe the bug**
When we deployed Tolgee v3.158.0 in production: 

- No batch jobs were being processed despite chunks being queued
- The internal queue accumulated 115,000 pending elements
- CPU usage maxed out, suggesting all CPU time was spent iterating through the massive queue rather than processing actual work

The batch job chunk execution queue was loading ALL pending chunks without any limit. When thousands of chunks accumulated, the system spent all its resources just traversing the queue instead of executing jobs.

**To Reproduce**
Have a tolgee instance with more than 150 000 pending elements to translate (with LLM) on a single project.

**Screenshots**
<img width="888" height="763" alt="Image" src="https://github.com/user-attachments/assets/cd65a093-6633-41e7-be72-8cd218ebd66a" />
<img width="897" height="386" alt="Image" src="https://github.com/user-attachments/assets/f1e8fdd1-cdbe-49c8-9bfa-5cd015f3c228" />

**Versions and environment**
- Tolgee Platform version: v3.158.0

**Notes**
We did a [fix](https://github.com/RakutenFrance/tolgee-platform/commit/65d63a29317722b920ba8cfddce387e07b00b972) that reintroduces the `chunkQueuePopulationSize` configuration (originally from PR https://github.com/tolgee/tolgee-platform/pull/3265, commit https://github.com/RakutenFrance/tolgee-platform/commit/90904f02f57ea98ba5c9b0a8ded8a5d1e840d608) which:

- Limits query results using `setMaxResults()
`- Adds a configurable cap (default: 1000) on chunks fetched per scheduled run
- Prevents queue bloat and restores processing capability

JanCizmar:
Hi @Jbbouille, thanks for reporting this. We're actively working on batch job throughput improvements and trying to reproduce the stall you experienced.

Could you clarify a few things about your setup when this happened?

1. **Single job or multiple jobs?** Were the 115K queued chunks from one large batch job, or from many smaller jobs on the same project?
2. **How many nodes / what concurrency setting** were you running with?
3. **Were you using Redis** for the multi-instance setup?

This would help us reproduce the exact conditions that caused the stall. We've tested with 100K+ chunks in the queue but haven't been able to trigger the complete processing halt you describe, so there's likely a specific combination of factors at play.

JanCizmar:
[...]

PR #3511, opened 2026-03-09
Limit batch queue size to prevent CPU waste during MACHINE_TRANSLATE processing

Jbbouille:
When the batch queue grows very large with many pending items, iterating through
the entire queue to fetch pending batches consumes significant CPU and memory resources
without providing any performance benefit. The queue traversal becomes a bottleneck
that slows down MACHINE_TRANSLATE batch processing.

Set max result limit to 500 items to fetch pending batches efficiently and prevent
the queue from degrading system performance. This ensures batch processing remains
responsive even with high workloads.

Closes: https://github.com/tolgee/tolgee-platform/issues/3480

## Summary by CodeRabbit

* **Performance**
  * Pending batch selection now respects a configurable maximum (default 500), reducing load and improving queue predictability and latency.

* **Configuration**
  * New validated setting to adjust the batch queue populate limit per run (range constrained, default 500).


Jbbouille:
As suggested I have added a validation (min and max) on `queuePopulateLimit`.

Issue #3482, opened 2026-02-19
Performance improvement: Project-level concurrency dramatically increases batch throughput for high-volume, small-batch workloads


Jbbouille:
**Describe the bug**
Without project-level concurrency (`projectConcurrency` setting), batch throughput is severely limited when dealing with high-volume, small-batch workloads. The system can only process ONE batch per project at a time, creating a major bottleneck that results in unacceptably low throughput (batches completed per hour).

**To Reproduce**
1. Configure a single project with on-demand translation API integration
2. Generate many small translation requests (5-20 keys each) via API calls
3. Monitor the batch queue - it fills with 100+ small batches (1-3 chunks each)
4. Observe that batches are processed sequentially (one completes before the next starts)
5. Measure throughput: only 2-4 concurrent LLM requests at any time despite having available resources

**Expected behavior**
With multiple small batches queued and sufficient system resources available, we expect:
- Multiple batches to process concurrently for the same project
- High utilization of CPU and network resources
- Many concurrent LLM API requests (respecting rate limits)
- Acceptable throughput measured in batches completed per hour

**Root Cause**
The `BatchJobProjectLockingManager` enforces strict project-level serialization by storing a single batch ID per project:

```kotlin
// OLD implementation
private val localProjectLocks = ConcurrentHashMap<Long, Long?>()
```

This means only ONE batch can run per project at a time, regardless of available resources.

**Solution (commit 61c7563ed39d1fa8f9ea40bf73ed1f87ae618998)**
The commit changes the lock storage to support multiple concurrent batches per project:

```kotlin
// NEW implementation
private val localProjectLocks = ConcurrentHashMap<Long, Set<Long>>()
```

And adds the `projectConcurrency` configuration:
```kotlin
@DocProperty(description = "How many parallel jobs can be run at once per project across all Tolgee instances")
var projectConcurrency: Int = 1
```

**Performance Impact**

Before (projectConcurrency = 1):
```
Queue: [Batch1: 3 chunks] [Batch2: 2 chunks] ... [Batch100: 3 chunks]

Processing (sequential):
Batch1: ████ (3 LLM requests) → completes
Batch2: ██ (2 LLM requests) → completes
...

→ Only 2-4 concurrent LLM requests
→ Poor throughput
→ Underutilized resources
```

After (projectConcurrency = 10):
```
Queue: [Batch1: 3 chunks] [Batch2: 2 chunks] ... [Batch100: 3 chunks]

Processing (parallel):
[...]

PR #3512, opened 2026-03-09
perf: allow concurrent MACHINE_TRANSLATE batch jobs across projects

Jbbouille:
MACHINE_TRANSLATE was exclusive by default, meaning only one job could run at a time per project. Setting `exclusive = false` allows multiple jobs to run concurrently, improving throughput when several projects trigger machine translation at the same time.

Closes: https://github.com/tolgee/tolgee-platform/issues/3482

## Summary by CodeRabbit

* **Bug Fixes**
  * Machine translation jobs can now execute concurrently, allowing multiple translation operations to run simultaneously without blocking other processes.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/data/BatchJobType.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/data/BatchJobType.kt
@@ -47,6 +47,7 @@ enum class BatchJobType(
     activityType = ActivityType.BATCH_MACHINE_TRANSLATE,
     maxRetries = 3,
     processor = MachineTranslationChunkProcessor::class,
+    exclusive = false
   ),
   AUTO_TRANSLATE(
     activityType = ActivityType.AUTO_TRANSLATE,


JanCizmar:
Hello! Thanks for the PR and the improvement. 🙏

Could you please share a bit more detail about the real usage? For example how many projects you usually run in parallel and how big the performance improvement is. It would help us try to replicate it locally.

 I am also a bit concerned that multiple jobs could run over the same key language pairs at the same time. That might cause some unexpected issues. 🤔

Thanks!

Jbbouille:
> Hello! Thanks for the PR and the improvement. 🙏
> 
> Could you please share a bit more detail about the real usage? For example how many projects you usually run in parallel and how big the performance improvement is. It would help us try to replicate it locally.
> 
> I am also a bit concerned that multiple jobs could run over the same key language pairs at the same time. That might cause some unexpected issues. 🤔
> 
> Thanks!

Hi @JanCizmar! Happy to share more context.

In production, we have a single project with 1.3 million keys. At the time we applied this change, we had a backlog of ~150,000 pending `MACHINE_TRANSLATE` batch jobs, and new batches are continuously being added — roughly 100 new translation batches per hour on average.

With` exclusive = true`, we were processing around 300–500 batches/hour, which meant we were barely keeping up and couldn't catch up with the backlog at all. After switching to `exclusive = false` (alongside the other small fixes), throughput jumped to roughly ~9,000 batches/hour, which allowed us to clear the backlog quickly.

Regarding your concern about multiple jobs running over the same key-language pairs simultaneously: in our case this isn't an issue because our jobs are generated from distinct import operations, so overlapping key-language pairs across concurrent jobs are unlikely (but not impossible... for now we preferred to have duplications than thousand of pending batches). That said, we understand this could be a legitimate concern for other use cases.

One possible middle ground would be to make exclusive configurable per job type ?

Thanks again for reviewing!

Issue #3513, opened 2026-03-09
Performance: Translation memory suggestions are fetched twice


Jbbouille:
## Problem
When `getPrompt()` is called, it creates two `PromptLazyMap` instances from the same `Variable` objects (one for fragment rendering, one for the final template rendering).

Without caching, any `lazyValue` (e.g. `translationMemory.json`) is evaluated twice, triggering duplicate SQL queries to `TranslationMemoryService#getSuggestionsList` for each translated item.

<img width="1792" height="882" alt="Image" src="https://github.com/user-attachments/assets/dafb4193-51cf-4504-8941-4236d06f850a" />

## Performance Trace
Full trace available in: https://gist.github.com/Jbbouille/420020861b069057c45f68e5f10f04ff

## Solution
Memoize lazy variable evaluation in `PromptLazyMap`:
- After the first evaluation, the result is stored in `value`
- `lazyValue` is set to null
- Subsequent accesses return the cached result directly

This prevents duplicate SQL queries and improves performance.

## Related Code
`backend/ee/app/src/main/kotlin/io/tolgee/ee/component/PromptLazyMap.kt`

PR #3514, opened 2026-03-09
perf: memoize lazy variable evaluation in PromptLazyMap

Jbbouille:
\`getPrompt()\` creates two \`PromptLazyMap\` instances from the same \`Variable\` objects (one for fragment rendering, one for the final template rendering).
Without caching, any \`lazyValue\` (e.g. \`translationMemory.json\`) was evaluated twice, triggering duplicate SQL queries to \`TranslationMemoryService#getSuggestionsList\` for each translated item.

After the first evaluation, the result is stored in \`value\` and \`lazyValue\` is set to null, so subsequent accesses return the cached result directly.

Closes: https://github.com/tolgee/tolgee-platform/issues/3513

## Summary by CodeRabbit

* **Performance Improvements**
  * Lazily-provided prompt values are now evaluated once, cached, and reused to avoid repeated computation.
  * String results are produced consistently to prevent redundant conversions and ensure stable output.

* **Tests**
  * Added unit tests verifying direct value retrieval, lazy evaluation behavior, and that lazy values are evaluated at most once.

### DIFF
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/component/PromptLazyMap.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/component/PromptLazyMap.kt
@@ -24,7 +24,7 @@ class PromptLazyMap : AbstractMap<String, Any?>() {
       return lazyMap
     }
 
-    val stringValue = promptValue?.lazyValue?.let { it() } ?: promptValue?.value
+    val stringValue = promptValue?.lazyValue?.invoke() ?: promptValue?.value
     return stringValue?.let { if (it is String) Handlebars.SafeString(it) else it }
   }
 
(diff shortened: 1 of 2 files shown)


Jbbouille:
> This looks like it mutates the shared Variable object as a side effect of `PromptLazyMap.get()`, nulling out lazyValue. Any other code that later accesses variable.lazyValue directly (not through the map) would find null instead of the lambda.
> 
> Maybe the Variable class can done somehow like:
> 
> ```kotlin
> class Variable(                                                                                                                                                                                                  
>     val name: String,                                                                                                                                                                                              
>     var value: Any? = null,                                                                                                                                                                                        
>     lazyValue: (() -> Any?)? = null,                                                                                                                                                                               
>     val description: String? = null,                                                                                                                                                                               
>     val props: MutableList<Variable> = mutableListOf(),                                                                                                                                                            
>     val type: PromptVariableType? = null,                                                                                                                                                                          
>     val option: BasicPromptOption? = null,                                                                                                                                                                         
>   ) {
>     val lazyValue: (() -> Any?)? = lazyValue?.let {                                                                                                                                                                
>       val cached by lazy { it() }
>       return@let { cached }                                                                                                                                                                                        
>     }             
>   }  
> ```
> 
> wdyt?

I think it's better than what I did before.
I made some changes to `Variables` and added unit tests. Could you take another look?

bdshadow (approved):

bdshadow:
@Jbbouille thank you! Please just fix the formatting. `./gradlew ktlintFormat`

Jbbouille:
@bdshadow done. Sorry for the inconvenience.

Issue #3515, opened 2026-03-09
Performance: LanguageStatsProvider#getResultForSingleProject generates a very slow SQL query (+20s in production)


Jbbouille:
## Problem

The method `io.tolgee.service.queryBuilders.LanguageStatsProvider#getResultForSingleProject` is extremely slow in production environments.

The current implementation uses the JPA Criteria API, which generates a query with multiple correlated subqueries — one per `TranslationState` (TRANSLATED, REVIEWED) for both key counts and word counts. This results in 4 correlated subqueries executed for each language in a project, making the query scale poorly as the number of keys and languages grows.

In our production environment, we observed execution times exceeding **20 seconds** for this query.

## Root Cause

The generated SQL issues several correlated subqueries over the `key`, `translation`, and `project` tables, causing repeated full scans. With a large number of keys and translations, the query planner cannot optimize this efficiently.

## Proposed Solution

Replace the Criteria API query with a single native SQL query using CTEs (`WITH ... AS MATERIALIZED`):

1. **`key_base_words`** — fetches all keys for the project with their base language word count, applying branch filtering once
2. **`lang_stats`** — aggregates translated/reviewed key counts and word counts per language in a single pass

This reduces the number of scans significantly and allows the database to optimize the execution plan.

In our production environment, the native query runs in approximately **2 seconds** — a **10x improvement**.

## Related Code

`backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/LanguageStatsProvider.kt`
Method: `getResultForSingleProject()`

PR #3516, opened 2026-03-09
perf: replace Criteria API with native CTE query in LanguageStatsProvider

Jbbouille:
The previous implementation used JPA Criteria API which generated
multiple correlated subqueries (one per TranslationState for both key
counts and word counts), causing repeated full scans on key/translation
tables. This resulted in execution times exceeding 20s in production.

Replace with a single native SQL query using two MATERIALIZED CTEs:
- key_base_words: fetches all project keys with base language word count,
  applying branch filtering once
- lang_stats: aggregates translated/reviewed counts per language in a
  single pass

Execution time in production: ~20s → ~2s

Closes: https://github.com/tolgee/tolgee-platform/issues/3515

## Summary by CodeRabbit

* **Refactor**
  * Language statistics are now produced by a single, native query for simpler and faster processing.
  * Optional branch-based filtering added to produce targeted stats when requested.
  * Result mapping streamlined for consistent, null-safe outputs.
  * Reduced internal complexity and improved performance of language-stats retrieval.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/LanguageStatsProvider.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/LanguageStatsProvider.kt
@@ -1,151 +1,81 @@
 package io.tolgee.service.queryBuilders
 
-import io.tolgee.model.Language_
-import io.tolgee.model.Project
-import io.tolgee.model.Project_
-import io.tolgee.model.branching.Branch
-import io.tolgee.model.branching.Branch_
-import io.tolgee.model.enums.TranslationState
-import io.tolgee.model.key.Key
-import io.tolgee.model.key.Key_
-import io.tolgee.model.translation.Translation
-import io.tolgee.model.translation.Translation_
 import io.tolgee.model.views.projectStats.ProjectLanguageStatsResultView
 import jakarta.persistence.EntityManager
-import jakarta.persistence.criteria.CriteriaBuilder
-import jakarta.persistence.criteria.CriteriaQuery
-import jakarta.persistence.criteria.JoinType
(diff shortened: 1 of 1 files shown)


Jbbouille:
I realized there was a formatting error, which is why I pushed a new version. Sorry for the inconvenience !

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/LanguageStatsProvider.kt
```diff
@@ -1,151 +1,80 @@
...
-
-  private fun joinBranch(
-    subquery: Subquery<*>,
-    join: ListJoin<Project, Key>,
-  ) {
-    val branchJoin = join.join(Key_.branch, JoinType.LEFT)
-    subquery.where(
-      cb.and(
-        branchPredicate(branchJoin, join.get(Key_.branch)),
-        cb.isNull(join.get(Key_.deletedAt)),
```
bdshadow, line 113: i don't see `deletedAt` filter in the rewritten query

Jbbouille: You're absolutely right, thanks for catching this! We'll add the `deleted_at IS NULL` filter to the `key_base_words CTE`. We missed it because we barely ever put keys in the trash in our usage, so it didn't show up in our testing.

Jbbouille: Also it seems that this feature was only introduced a week ago ([#3486](https://github.com/RakutenFrance/tolgee-platform/issues/3486)), so it's very likely no one has actually tried putting keys in the trash yet...

PR #3520, opened 2026-03-10
fix: prevent NonUniqueObjectException on duplicate screenshot references

Anty0:
## Summary

- Fix `NonUniqueObjectException` in `ResolvingKeyImporter.importScreenshots()` when the same `(key, screenshot)` pair is persisted twice in the same Hibernate session
- This happens when the API request contains duplicate `uploadedImageId` entries for the same key (either repeated in the screenshots array, or across multiple items resolving to the same key)
- Added deduplication tracking via a `Set<Pair<Long, Long>>` to skip already-processed references

## Summary by CodeRabbit

* **Bug Fixes**
  * Import operations now prevent duplicate screenshot references from being created during file processing, ensuring cleaner imports and improved data consistency without redundant entries.


JanCizmar (approved):
It's a bit ugly and would be nice to have comment why this is happening in the code, but nevermind.

Anty0:
Yee, definitely not the cleanest solution. 😅

PR #3521, opened 2026-03-11
fix: Merge duplicate screenshot references during key import

Anty0:
## Summary
- Fix `NonUniqueObjectException` crash on the deprecated `POST keys/import-resolvable` endpoint when the same `(key, screenshot)` pair appears more than once in a single import (reported via Sentry).
- `ResolvingKeyImporter.importScreenshots()` now accumulates references and merges duplicates instead of persisting one per entry: positions are combined (identical positions are deduplicated, so a repeated submit stays idempotent) and the first non-null text is kept. Each `(key, screenshot)` reference is persisted once after the whole import is processed.
- Scope: merge applies only to this endpoint; the standard data-import path (`ScreenshotImporter`) is intentionally unchanged.
- Tests cover within-entry merge (position contents + order), cross-entry merge (same key name across separate entries), first-non-null text, a duplicate with no positions, and identical-position deduplication.

## Summary by CodeRabbit

* **Bug Fixes**
  * Importer now merges duplicate screenshot references (including across separate key entries), concatenates position data, avoids duplicate positions, and preserves the first available reference text; duplicates without positions retain existing positions.
* **Chores**
  * Consolidated reference handling during import to ensure a single merged reference per screenshot/key and more reliable cleanup of obsolete references.
* **Tests**
  * Added integration tests validating merged positions, text-selection rules, cross-entry merging, and deduplication behavior.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/KeyControllerResolvableImportTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/KeyControllerResolvableImportTest.kt
@@ -10,6 +10,7 @@ import io.tolgee.fixtures.andIsForbidden
 import io.tolgee.fixtures.andIsOk
 import io.tolgee.fixtures.generateImage
 import io.tolgee.fixtures.node
+import io.tolgee.model.key.screenshotReference.KeyInScreenshotPosition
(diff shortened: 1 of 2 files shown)


JanCizmar (approved):

JanCizmar:
Test would be nice.

Anty0:
@JanCizmar Good point, I'll create some tests in cooldown ^^

dkrizan (approved):

PR #3531, opened 2026-03-13
fix: optimize org stats count queries and fix branching bug

### LINKED ISSUE #3527: Organization stats count queries are slow and incorrectly count branch keys
## Problem

`OrganizationStatsService.getTranslationCount()` and `getKeyCount()` are slow for large organizations because:

1. **Missing index** — The `key.deleted_at IS NULL` filter in the count queries can't efficiently use existing indexes. The existing `key_deleted_at_null` index is an expression index on `(deleted_at IS NULL)` but doesn't include `project_id`. The unique indexes (`key_project_branch_name_*`) include `coalesce(branch_id, -1)` which doesn't align with the `DISTINCT(project_id, name, namespace_id)` in the count query.

2. **Branching bug** — When `project.useBranching = false`, keys under branches are still counted. This means organizations that disable branching are billed for branch keys that shouldn't count.

## Impact

- The `getTranslationCount()` query takes ~4 seconds for large orgs
- This fires on every new translation creation (empty → text) via `StringKeysLimitListeners`
- Incorrect billing for orgs with branching disabled

### Index
Add a partial index optimized for the count queries:
```sql
CREATE INDEX key_project_name_ns_not_deleted
ON key (project_id, name, namespace_id)
WHERE deleted_at IS NULL;
```

### Branching bug
Update `getTranslationCount()` and `getKeyCount()` in `OrganizationStatsService` to exclude branch keys when `project.useBranching = false`:
```sql
AND (p.use_branching = true OR k.branch_id IS NULL)
```

## Relevant files
- `OrganizationStatsService.kt` — the count queries
- `StringKeysLimitListeners.kt` (billing module) — triggers the count on every translation save
- `schema.xml` — existing indexes on key table

dkrizan:
## Summary
- Add partial index `key_project_name_ns_not_deleted` on `key(project_id, name, namespace_id) WHERE deleted_at IS NULL` to speed up org-wide `COUNT(DISTINCT)` queries in `OrganizationStatsService`
- Fix bug where `getKeyCount()` and `getTranslationCount()` counted keys on orphan branches even when `project.useBranching = false`
- Add tests for branching exclusion behavior

Closes #3527

## Test plan
- [x] `OrganizationStatsServiceTest` — all 6 tests pass including new branching exclusion tests
- [x] CI passes
- [ ] Verify on testing instance that PUT translations endpoint is faster for large orgs

## Release Notes

* **Bug Fixes**
  * Corrected organization statistics to accurately reflect key and translation counts when branching is disabled for projects, preventing branch-related items from being included in totals.

* **Chores**
  * Added database index to improve performance of statistics queries.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationStatsService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/organization/OrganizationStatsService.kt
@@ -44,23 +44,31 @@ class OrganizationStatsService(
     return entityManager
       .createNativeQuery(
         """
-        select count(*) from (
-          select distinct k.project_id, k.name, k.namespace_id, t.language_id
-          from translation t
-          join key k on k.id = t.key_id
+        with org_keys as materialized (
+          select k.id, k.project_id, k.name, k.namespace_id
+          from key k
           join project p on p.id = k.project_id and p.deleted_at is null
           left join branch b on b.id = k.branch_id
           where p.organization_owner_id = :organizationId
             and k.deleted_at is null
             and (k.branch_id is null or b.deleted_at is null)
             and (p.use_branching = true or k.branch_id is null or b.is_default = true)
-            and exists (
-              select 1 from language l
-              where l.id = t.language_id
-                and l.deleted_at is null
-            )
+        ),
+        org_translations as materialized (
+          select ok.project_id, ok.name, ok.namespace_id, t.language_id
+          from org_keys ok
+          join translation t on t.key_id = ok.id
             and t.text is not null
             and t.text <> ''
+        )
+        select count(*) from (
+          select distinct ot.project_id, ot.name, ot.namespace_id, ot.language_id
+          from org_translations ot
+          where exists (
+            select 1 from language l
+            where l.id = ot.language_id
+              and l.deleted_at is null
+          )
         ) sub
         """.trimIndent(),
       ).setParameter("organizationId", organizationId)


backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5174,4 +5174,16 @@
             referencedTableName="user_account" referencedColumnNames="id"
             constraintName="fk_key_deleted_by_user_account"/>
     </changeSet>
+
+    <!-- Partial index for org-wide key/translation count queries (fixes #3527) -->
+    <changeSet author="dkrizan" id="1741855200000-1" runInTransaction="false">
+        <sql>
+            create index concurrently if not exists key_project_name_ns_not_deleted
+                on public.key (project_id, name, namespace_id)
```
bdshadow, line 5182: it looks like the same index, but not partial, already exists. It was added in the changeset 1771846650000-1. So it's not clear why the new one is required here.

Maybe, for the usage query it must be for 2 fields: project_id and deleted_at? Not sure, just asking

dkrizan: It improved the performance by 25%, I think due to `where deleted_at is null` .. but I will dive deeper and find a better way.

bdshadow: hm, got it. It's interesting why the existing index isn't taken then... Maybe because of more fields in it 🤔

dkrizan: I think the reason Postgres does not choose it is the condition `WHERE deleted_at IS NOT NULL`, in this case, the opposite is needed.

dkrizan: Looks like this is the best solution + the refactor of SQLs.

bdshadow: Tested it myself. And approving. It's weird. I think we should consider removing the non-partial one

Image

bdshadow (approved):

PR #3535, opened 2026-03-16
test: fix flaky SlackWithAutoTranslationTest race condition

bdshadow:
Both AUTO_TRANSLATE and AUTOMATION batch jobs are non-exclusive, so they run concurrently with no ordering guarantee. When the automation sends the Slack message before auto-translation completes, the chatPostMessage only contains the base translation. The auto-translated text then arrives via chatUpdate. Assert against the latest Slack state (chatUpdate if present, otherwise chatPostMessage) and increase the timeout from 3s to 10s for CI resilience.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for Slack auto-translation integration by implementing more flexible assertion mechanisms and extending timeout thresholds to better handle concurrent operations. Enhanced test validation to accurately verify translation content across different messaging scenarios.


bdshadow:
@coderabbitai review please

PR #3540, opened 2026-03-18
perf: best of both worlds for translation memory suggestion queries

Jbbouille:
### Statement timeout
On large databases, the translation memory similarity search can be extremely slow — in the worst case, several minutes. Rather than leaving the query running indefinitely, we now set `statement_timeout = 550ms` before each TM query. If the timeout fires, the user simply gets no suggestions instead of having the request hang or time out at the HTTP layer. This is a deliberate trade-off: a fast no-result is better than a multi-second freeze.

### REQUIRES_NEW transaction
`SET LOCAL` in PostgreSQL only scopes settings to the current transaction. Without an isolated transaction, the `statement_timeout` and `pg_trgm.similarity_threshold` settings could leak into the caller's transaction and affect unrelated queries. `REQUIRES_NEW` creates a fresh transaction for each TM query, guaranteeing these settings are cleaned up when the method returns. It also ensures that a `QueryTimeoutException` (raised when the 550ms timeout fires) does not mark the outer transaction as rollback-only, which would cause a `TransactionSystemException` on the caller.

### getSuggestionsData query improvement
The previous query used a `CASE WHEN` workaround to force PostgreSQL to evaluate cheap conditions before the trigram similarity check — a hack needed because the planner was underestimating cardinality and evaluating the similarity condition first. The new query replaces this with:

- `baseTranslation.text % :baseTranslationText` — the `%` threshold operator, which the composite GiST index on `(language_id, text gist_trgm_ops)` can evaluate efficiently
- `ORDER BY baseTranslation.text <-> :baseTranslationText` — orders results by actual trigram distance so the most relevant suggestions appear first

This produces the same result (top suggestions ordered by similarity) but leverages the GiST index properly, eliminating the full scan. On our test environment the same query went from **~6800 ms down to ~320 ms**.

## What

- `getSuggestionsData` (UI suggestion panel): rewrote the query — removed the `CASE WHEN` workaround, replaced with `%` filter + `ORDER BY <->`. Added `statement_timeout = 550ms`. Wrapped the public `getSuggestions` entry points in `REQUIRES_NEW` so the `SET LOCAL` settings are properly scoped (note: `getSuggestionsData` is called via self-invocation so Spring AOP cannot proxy it directly — `REQUIRES_NEW` is applied on the public `getSuggestions` overloads instead).
- `getSuggestionsList` (MT/batch pipeline): added `statement_timeout = 550ms`. Kept the no-`ORDER BY` design so the GiST index can stop early at `LIMIT` rows. Upgraded from `@Transactional` to `@Transactional(REQUIRES_NEW)`.

Note: I noticed https://github.com/tolgee/tolgee-platform/pull/3537 too late — I had already started this work independently. Apologies for the overlap.

## Summary by CodeRabbit

* **Refactor**
  * Suggestion query logic simplified and reordered for more consistent and more relevant ranking.
  * Result counting and similarity mapping adjusted to ensure correct totals and similarity values.

* **Bug Fixes / Reliability**
  * Per-request query timeouts enforced to avoid long-running lookups.
  * Timeout errors are caught and logged; suggestion and example endpoints return empty results on timeout instead of failing.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryService.kt
@@ -78,7 +78,7 @@ class TranslationMemoryService(
         keyName = it[2] as String,
         keyNamespace = it[3] as String?,
         keyId = it[4] as Long,
-        similarity = it[5] as Float,
+        similarity = (it[5] as Number).toFloat(),
       )
     }
   }
@@ -171,7 +171,7 @@ class TranslationMemoryService(
         .setFirstResult(offset.toInt())
         .resultList
 
-    val count = (queryResult.firstOrNull() as Array<*>?)?.get(6) as Long? ?: 0L
+    val count = ((queryResult.firstOrNull() as Array<*>?)?.get(6) as Number?)?.toLong() ?: 0L
     return count to
       queryResult.map {
         it as Array<*>
@@ -180,7 +180,7 @@ class TranslationMemoryService(
           baseTranslationText = it[1] as String,
           keyName = it[2] as String,
           keyNamespace = it[3] as String?,
-          similarity = it[5] as Float,
+          similarity = (it[5] as Number).toFloat(),
           keyId = it[4] as Long,
         )
       }
--- a/backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryService.kt
@@ -5,6 +5,7 @@ import io.tolgee.model.Language
 import io.tolgee.model.key.Key
 import io.tolgee.model.views.TranslationMemoryItemView
 import jakarta.persistence.EntityManager
+import jakarta.persistence.QueryTimeoutException
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.PageImpl
 import org.springframework.data.domain.Pageable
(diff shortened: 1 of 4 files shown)


bdshadow (changes requested):
Hi @Jbbouille! Thank you for the pr!
Could you please fix the `TranslationSuggestionControllerTmTest`. Probably, the same problem will be for `suggestions/suggestions.reviewer.cy.ts` and other tests in the `suggestions` directory

backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryService.kt
bdshadow: i guess it needs `Propagation.REQUIRES_NEW` too, since it's public, and can be called from anywhere.

backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryService.kt
```diff
@@ -31,6 +37,9 @@ class TranslationMemoryService(
     targetLanguage: LanguageDto,
     limit: Int = 5,
   ): List<TranslationMemoryItemView> {
+    // Protect against slow queries on large datasets. REQUIRES_NEW ensures this
+    // timeout does not leak into the caller's transaction.
+    entityManager.createNativeQuery("set local statement_timeout to '550ms'").executeUpdate()
```
bdshadow, line 40: @Jbbouille thanks for the fixes and sorry for a delay on review

I think i saw in some previous pr-s, that there was a `try { ... } catch (e: QueryTimeoutException) { emptyList()/emptyPage }`, which is currently not here, so when in the current implementation the `QueryTimeoutException` is thrown it will result as an HTTP 500.

bdshadow (changes requested):

bdshadow:
@Jbbouille i didn't want to disturb you with it more, and i brazenly tried to add a commit to this pr, but it's not possible since i don't have access to your fork.

I've created a git-patch. Please take a look if it's ok, and add it please here 

[0001-fix-wrap-jakarta-QueryTimeoutException-consistently-.patch](https://github.com/user-attachments/files/26828623/0001-fix-wrap-jakarta-QueryTimeoutException-consistently-.patch)

Jbbouille:
Done, I've applied your patch and pushed it to the branch. Thanks for putting it together!

Jbbouille:
Sorry about that, I forgot to run ktlint again 🙈

bdshadow (approved):

PR #3548, opened 2026-03-19
fix: preserve ICU exact-match plural selectors (=0, =1) in export

### LINKED ISSUE #3267: Translation with plurals corrupted/changed after import
**Describe the bug**

When we import translation key with several plurals usage, Tolgee corrupts it and tries to transform using some magic internal rules, but result after transformations is not correct.

**Example**

We have a quite simple translation to present duration (with hours and minutes) in our application:

ORIGINAL translation file we import into Tolgee:
```json
{
  "duration": "{hours, plural, =0 {} =1 {1 h} other {{hours} h}} {minutes, plural, =0 {} =1 {1 min} other {{minutes} min}}"
}
```

Depending on numeric `hours` and `minutes` it prints human-readable value and omit zero parts. 
For example: `1 h 20 min`, `45 min`, `3 h`, etc

When we import such translation key using Tolgee import, it's shown in Web UI and detect only `hours` variable, without case for zero hours

And after export (or by getting it via Content Delivery) this translation key in JSON format it becomes:

RESULT corrupted translation:
```json
{
  "duration": "{hours, plural, other {{hours} h {minutes, plural, =0 {} =1 {1 min} other {{minutes} min}}} =1 {1 h {minutes, plural, =0 {} =1 {1 min} other {{minutes} min}}}}"
}
```

As a result we always have not omitted zero hours parts like `0 h 45 min`, `0 h 5 min`, etc

**To Reproduce**

Prepare JSON file with one translation key:

```json
{
  "duration": "{hours, plural, =0 {} =1 {1 h} other {{hours} h}} {minutes, plural, =0 {} =1 {1 min} other {{minutes} min}}"
}
```

1. Go to 'Import'
2. Select prepared JSON file and import it
3. Find the translation in translations, and check how it looks in Web UI
4. Export imported translation key and compare it with original imported value

**Expected behavior**

One of these options is acceptable:

- Tolgee import should not corrupt correct translation where several plurals are used .
- Or we need to have an ability to disable Tolgee transformation magic for such plurals

**Screenshots**

Here is how imported translation looks in Web UI (there is no option for zero `hours`):

Image

**Versions and environment**

- Tolgee Platform version: self-hosted Jar archive, v3.133.1
- Environment: openjdk 21.0.8 2025-07-15 LTS

peterzen:
Two related issues caused push→pull round-trips to produce file diffs:

1. IcuToGenericFormatMessageConvertor.convert() called toIcuPluralString() without optimize=false, relying on the default optimize=true. This triggered optimizePluralForms + orderPluralForms on every export, silently reordering or removing plural forms that the user authored.

2. orderPluralForms sorted unknown keys (including =0, =1, =2 exact-match selectors) alphabetically after 'other', placing them last instead of first. ICU exact-match selectors should precede keyword selectors and appear in ascending numeric order.

Fix: pass optimize=false in the export path so the stored ICU string is emitted verbatim. Also update orderPluralForms to sort =N selectors before keyword forms in numeric order so that explicit optimize calls produce canonical output consistent with the ICU specification.

Fixes: https://github.com/tolgee/tolgee-platform/issues/3267

## Summary by CodeRabbit

* **Bug Fixes**
  * Exact-match plural selectors (e.g., =0, =1) now sort before keyword plural forms and in ascending numeric order.
  * Plural serialization preserves exact-match selectors and their order (optimization disabled to keep explicit selectors).

* **Tests**
  * Added/expanded tests covering exact-match selector ordering and round-trip serialization of plural forms.


JanCizmar:
Hey! Thanks a lot for the PR. @Anty0 can you please review this?

Anty0:
Hi! Thanks for the PR! ^^
Sorry for the late response.

I took a look and discussed this with a colleague. We can't drop the optimization. It's important for us to, for example, remove empty forms. We treat empty forms as "untranslated". Changing that would be a major breaking change. We also don't natively support multiple plurals within a single key. While that technically works, it won't look nice in the UI and could have some weird side effects. When importing a string, we stretch the first plural we find to cover the whole string. That is intentional, as that way the user can work with the plural in Tolgee UI properly.

Looking at the original issue, the only "problem" I can find is that Tolgee removed the empty `=0` case for hours, which isn't something we can fix. We recommend handling logic in code. In that case, the proper solution would be to write a bit of code that omits hours/minutes when needed - translation should only handle translating strings.

When it comes to ordering plurals, I checked, and it is preferred to have exact (`=`) forms first, so I'll keep that change.

I'll create a follow-up PR based on yours to clean it up. Thanks again for looking into this! ^^

Anty0:
Superseeded by #3681

PR #3560, opened 2026-03-24
feat: add self-hosted plans placeholder

dkrizan:
## Summary
- Show a placeholder with "Self-hosted plan" title, description, and "Contact us" button when no self-hosted plans are available
- Add SVG illustration for the placeholder
- Only affects the self-hosted tab on the Subscriptions page

## Test plan
- [ ] Navigate to Subscriptions > Self-hosted with no public self-hosted plans configured
- [ ] Verify placeholder shows with correct text and contact button
- [ ] Verify plans grid shows normally when plans exist
- [ ] Run billing e2e tests: `self-hosted-no-plans-placeholder.cy.ts`

## Summary by CodeRabbit

* **New Features**
  * Added a placeholder interface for self-hosted plans, displaying informational content and a contact option when no plans are currently available.


webapp/src/ee/billing/Subscriptions/selfHosted/PlansSelfHosted.tsx
```diff
@@ -112,50 +113,57 @@ export const PlansSelfHosted = () => {
...
-                  <T keyName="organization-billing-self-hosted-no-active-subscriptions" />
+      {plansLoadable.data && activeSubscriptionsLoadable.data && period && (
+        <>
+          {hasPlans ? (
+            <>
+              <Box mb={4}>
+                <Typography variant="h6" mb={2}>
+                  {t('organization-billing-self-hosted-active-subscriptions')}
+                </Typography>
+                {activeSubscriptions?.length && (
```
bdshadow, line 129: just a question.
Shouldn't `activeSubscriptions` be shown regardless of `hasPlans` value? I mean, with this changes if there are no plans `activeSubscriptions` won't be shown. I mean if a plan was removed, for example, but a user still has active subscritpion, it won't be shown. Is it correct?

dkrizan: The conditions are essentially the same as before. I just moved a part of the code so that when there are no plans, the placeholder is displayed. The `hasPlans` variable is the same as `plansLoadable.data?._embedded?.plans`, so if what you’re saying is true, it would be broken before.
I believe if a user has active subscriptions, the plans they subscribed to should still exist in `plansLoadable.data` - so it is actually expecteed to hide empty Active subscriptions list when there are no plans.

PR #3568, opened 2026-03-25
fix: avoid loading full keyMetas collection when deleting tagged keys

### LINKED ISSUE #3567: Memory spike in TagService when deleting keys with tags linked to many keys
**Describe the bug**
`TagService.deleteAllTagsForKeys` calls `getTagsWithKeyMetas` which executes `join fetch t.keyMetas`, loading the entire `KeyMeta` collection for every affected tag into the JPA session (including eagerly-fetched `Key` and `Namespace` associations). For projects where tags are linked to thousands of keys, this materializes 10k+ entities per deletion. Under concurrent `DELETE_KEYS` batch jobs, this leads to G1 Old Gen filling up, excessive GC pressure, and liveness probe failures.

The same issue exists in the single-key `remove(key, tag)` path which accesses `tag.keyMetas` directly.

**To Reproduce**
1. Create a project with a large number of keys (10k+) all sharing a common tag
2. Trigger multiple concurrent batch key deletions (e.g. via the API)
3. Observe JVM heap usage — G1 Old Gen fills up, GC overhead increases
4. Liveness probe eventually fails under sustained load

**Expected behavior**
Deleting keys should not load unrelated `KeyMeta` entities into the JPA session. Only the tags that would become empty need to be identified and removed.

**Screenshots**
N/A

**Versions and environment**
- Tolgee Platform version: 3.173.4
- Environment: Kubernetes, Java 21, multiple concurrent batch workers
- Browser: N/A (backend issue)

**Additional context**
Suggested fix: replace the full entity-graph fetch with a pure-ID `EXISTS` subquery (`findTagIdsThatWouldBecomeEmpty`) that returns only tag IDs whose `keyMeta` count would drop to zero, then bulk-delete them via JPQL. A fix is proposed in #3568.

Jbbouille:
Closes #3567

## Summary

- `deleteAllTagsForKeys` used `join fetch t.keyMetas` to load the entire `KeyMeta` collection for every affected tag into the JPA session — 10k+ entities for large projects
- Under concurrent `DELETE_KEYS` batch jobs this caused G1 Old Gen to fill up, excessive GC, and liveness probe failures
- Same issue existed in the single-key `remove(key, tag)` path

## Changes

- Replace `getTagsWithKeyMetas` (full entity-graph fetch) with `findTagIdsThatWouldBecomeEmpty`, a pure-ID `EXISTS` subquery that returns only tag IDs that would become empty
- Remove empty tags via a bulk JPQL `DELETE` — no entity graph loaded
- `flushAutomatically = true` on `deleteByIdIn` is required (not defensive): Hibernate AUTO flush mode does not flush `key_meta_tags` before a `DELETE FROM tag`, which would cause a FK constraint violation
- `getTagsWithKeyMetas` removed as dead code

## Summary by CodeRabbit

* **Refactor**
  * Streamlined tag dissociation to avoid loading full tag associations and perform a single conditional bulk delete of unreferenced tags, reducing DB work and improving performance.
* **Bug Fix**
  * Improved tag lookup and project ownership validation to ensure tags are verified before removal, preventing incorrect deletions.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
@@ -425,7 +425,8 @@ class TagService(
     key: Key,
     tagId: Long,
   ) {
-    val tag = getWithKeyMetasFetched(key.project.id, tagId)
+    val tag = find(tagId)?.takeIf { it.project.id == key.project.id }
+      ?: throw NotFoundException(Message.TAG_NOT_FOUND)
     remove(key, tag)
   }
 }
--- a/backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
@@ -406,7 +406,7 @@ class TagService(
     tagId: Long,
   ) {
     val tag =
-      find(tagId)?.takeIf { it.project.id == key.project.id }
+      tagRepository.findByIdAndProjectId(tagId, key.project.id)
         ?: throw NotFoundException(Message.TAG_NOT_FOUND)
     remove(key, tag)
   }
--- a/backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
@@ -192,12 +192,15 @@ class TagService(
     tag: Tag,
   ) {
     key.keyMeta?.let { keyMeta ->
-      tag.keyMetas.remove(keyMeta)
+      // Check emptiness via query before modifying anything.
+      // Do NOT access tag.keyMetas — it lazy-loads the entire KeyMeta
+      // collection for the tag, which can be 10k+ objects for large projects.
+      val tagWouldBecomeEmpty =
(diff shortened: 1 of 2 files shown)


backend/data/src/main/kotlin/io/tolgee/service/key/TagService.kt
```diff
@@ -412,7 +405,9 @@ class TagService(
     key: Key,
     tagId: Long,
   ) {
-    val tag = getWithKeyMetasFetched(key.project.id, tagId)
+    val tag =
```
bdshadow, line 408: why is this change needed here? `find()` loads here tags which are not related to the current project , the previous call to `getWithKeyMetasFetched` didn't do it. What am i missing here?

Jbbouille: The issue was that `getWithKeyMetasFetched` called `findWithKeyMetasFetched` which does a `left join fetch t.keyMetas` — so for a tag used on many keys, Hibernate would load potentially thousands of `KeyMeta` objects into memory even though none of them were needed here. That's what we wanted to avoid. That said, you're right that we lose the project scope check at the DB level — I'll add a dedicated `findByIdAndProjectId` query without the join fetch to get both.

bdshadow (approved):

PR #3574, opened 2026-03-27
feat: add upload_image MCP tool and screenshot support

Anty0:
## Summary

- Add a `get_image_upload_url` MCP tool that returns a short-lived, signed, **unauthenticated** upload URL, so AI agents upload screenshot bytes out-of-band (e.g. `curl -F image=@file "<uploadUrl>"`) instead of base64-encoding them through the model. The response carries an `uploadedImageId` to pass to `create_keys` / `add_key_screenshots`.
- Add a public `POST /v2/public/image-upload?token=<jwt>` endpoint, authorized solely by the signed `IMG_UPLOAD` ticket (no API key needed). The URL is reusable within its expiry window — default 30 min, configurable via `tolgee.mcp.image-upload-url-expiration-ms`.
- Deprecate the base64 `upload_image` tool (kept for backward-compat) and rename its JSON output key `imageId` → `uploadedImageId` so both upload paths use one spelling matching the consumer field.
- `create_keys` accepts per-key `screenshots` (`uploadedImageId` + optional positions); add `add_key_screenshots` for existing keys; persist associations via `ScreenshotService.saveUploadedImagesForKeys`. Screenshot association is atomic with key creation (one transaction; the `SCREENSHOTS_UPLOAD` permission is checked before any key is created), and the per-key screenshot limit is enforced before image conversion so an over-limit batch leaves no orphaned files.
- **Richer key metadata in a single `create_keys` / import call**, so the auto-translation that runs right after creation has full context (it is not redone if metadata is added later):
  - per-key `description` (already fed the translation prompt), `custom` (arbitrary structured metadata), `comments`, and `codeReferences` (`{path, line}`) — `custom`/`comments`/`codeReferences` also flow through the REST `POST /keys/import` endpoint via the shared `ImportKeysItemDto`.
  - a call-level `relatedKeysInOrder` param that stores **big meta** (key proximity) in the same transaction, so it lands before the post-commit auto-translate batch and feeds the AI translation prompt. Validated metadata limits are owned by the entities; blank entries are skipped and over-long ones rejected with a clean 4xx.
- Map malformed/undecodable images, degenerate image dimensions, and unknown JWT ticket types to a clean 4xx instead of a 500 — these source-level fixes also cover the existing authenticated `/v2/image-upload` and avatar paths.
- Update `webapp/CLAUDE.md` to the URL-based screenshot workflow and remove the app start/stop lifecycle instructions (the developer manages the app lifecycle manually); regenerate the frontend API schema.

## ⚠️ Before merging

- **Set `tolgee.back-end-url` in the deployed configuration.** When it's unset, the issued upload URL falls back to the current servlet-request origin.

## Notes

- The upload URL is **reusable within its expiry window**, not strict single-use (diverges from the issue's literal wording — accepted; lever for tightening: a shorter default lifetime).
- The pre-existing `ImageIO` decompression-bomb risk (full decode before the 3M-px clamp) is now reachable on the unauthenticated route; a pre-decode pixel cap is deferred to its own cross-cutting change.


JanCizmar:
As we discussed. Tracked as #3620 — **MCP for better development with AI agents**.

**1. `create_keys` should accept full key metadata in the same call**
Auto-translation runs right after key creation and uses metadata (especially `description`, but also `comments`, `codeReferences`, `custom`) as context. If we add metadata in a follow-up call, the translations have already been generated without that context. We want everything — metadata *and* screenshots — passed in a single `create_keys` call, so auto-translation sees the full picture.

**2. Leave `webapp/CLAUDE.md` alone re: app startup/shutdown**
The developer manages when the app is started or killed manually. We don't want the MCP tool docs prescribing lifecycle instructions there.

**3. Base64 upload isn't the right shape**
Passing base64 through a tool call is wasteful on tokens and slow. Let's mark `upload_image` as **not recommended** and add an MCP tool that returns a single-use upload URL instead. The client (Claude or any other tool) uploads the image directly to that URL, out-of-band. The URL is single-use, so it doesn't need an API key.

**4. Add a dedicated AI context field on keys**
Separately from `description` (which is for translators), we'd like a first-class `aiContext` field where AI coding tools can stash their own context — code positions, usage hints, whatever is useful for an LLM editing translations later. Today that info would end up in `description` (wrong audience) or `custom` (unstructured). A dedicated field keeps concerns separated and gives AI tools a known place to read/write.

**5. Public skill + video**
We'd like to create a public skill and a short video showing how to use Tolgee with Claude properly, so the workflow is discoverable.

PR #3578, opened 2026-03-30
fix: optimize interaction with redis for batch jobs 

bdshadow:
## Summary by CodeRabbit

* **Refactor**
  * Simplified internal state handling and cleanup logic for batch job processing.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/batch/ProgressManager.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/batch/ProgressManager.kt
@@ -398,8 +398,8 @@ class ProgressManager(
    * we have to unlock the project, otherwise it will be locked forever
    */
   fun finalizeIfCompleted(jobId: Long) {
-    val cached = batchJobStateProvider.getCached(jobId)
-    logger.debug("Checking if job $jobId is completed, has cached value: ${cached != null}")
+    val hasCached = batchJobStateProvider.hasCachedJobState(jobId)
+    logger.debug("Checking if job $jobId is completed, has cached value: $hasCached")
     // Use getJobDtoNoCache to avoid race condition where cache gets stale data
     // during another transaction's commit window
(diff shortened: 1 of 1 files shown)


bdshadow:
@JanCizmar also added our `e2e/scripts` to be run in gh actions

.github/workflows/test.yml
```diff
@@ -480,6 +480,39 @@ jobs:
         run: npm run eslint
         working-directory: ./email
 
+  integration-test:
+    name: Integration testing 🔗
```
JanCizmar: I wouldn't call it integration testing. It's more like, `Custom Script / Multi-node testing`. However, I don't know if I would call it from pipeline, since its designed more to run locally, test performance on it. I don't even know if it returns proper exit codes etc...

What are your reasons behind running this in CI/CD?

bdshadow: The motivation was that it is always run, because i forget to test it locally. You've recently asked if i checked the pr 3566 with this script, and i actually just forgot about it.
We can improve the exit codes and add some basic asserts. 
Also you mentioned testing performace with it - when it's not run, we don't even know the dynamic of performance degradations or improvements. This way we'll at least have it in the gh actions history.

However, i'm not insisting. If you think that it isn't worth supporting and want me to drop the commit, no problem

JanCizmar: I'm a bit worried this won't tell us much in practice. The script was designed to compare two versions as a benchmark. It has a lot of options that are useful for different scenarios.

I'd also prefer not to add more to the CI/CD pipeline. It's already slowing us down, and there's a limit on concurrent jobs on GitHub, so running this every time adds up.

If I remember correctly, I asked about running it for cases where we were potentially affecting multi-node batch job performance. That's pretty rare.

I'd lean towards dropping this part. Instead, I'd invest in mining data from Grafana to monitor regressions properly. That way we'd know what's actually happening in production and what's causing real-world regressions.

PR #3579, opened 2026-03-31
fix: flush merge changes before branch deletion to prevent flaky test

bdshadow:
Move entityManager.flush() to run unconditionally after branchMergeService.applyMerge() instead of only in the keep-branch path. BranchMergeExecutor modifies translations in-memory via Translation.merge() without explicit save, relying on JPA dirty-checking. Without this flush, changes may not be persisted before branch deletion or async listeners.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved branch merge functionality to ensure all translations are properly persisted after merge operations.

* **Tests**
  * Enhanced test infrastructure with improved branch revision synchronization checks to ensure consistency during asynchronous operations.

### DIFF
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/branching/merging/BranchMergeExecutor.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/branching/merging/BranchMergeExecutor.kt
@@ -175,7 +175,7 @@ class BranchMergeExecutor(
   }
 
   private fun persistAfterMerge(key: Key) {
-    key.translations.filter { it.id == 0L }.forEach { translationService.save(it) }
+    key.translations.forEach { translationService.save(it) }
   }
 
   private inline fun BranchMergeChange.withSnapshotKey(
--- a/ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/branching/BranchMergeServiceTest.kt
+++ b/ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/branching/BranchMergeServiceTest.kt
@@ -207,33 +207,17 @@ class BranchMergeServiceTest : AbstractSpringTest() {
   @Test
   fun `apply merge - propagates additions, updates and deletions`() {
     val merge = prepareMergeScenario()
-
-    val changes = merge.changes.map { "${it.change}(src=${it.sourceKey?.name}, tgt=${it.targetKey?.name})" }
-    val featureTranslationBeforeApply =
-      translationService
-        .find(
-          keyService.find(testData.featureKeyToUpdate.id)!!,
-          testData.englishLanguage,
-        ).orElse(null)
-        ?.text
-
     applyMerge(merge)
-
-    val actualTranslation = testData.mainKeyToUpdate.enTranslation()
-    if (actualTranslation != updatedValue) {
-      throw AssertionError(
-        "Expected '$updatedValue' but got '$actualTranslation'. " +
-          "Merge changes: $changes. " +
-          "Feature key translation before apply: '$featureTranslationBeforeApply'. " +
-          "Main key id: ${testData.mainKeyToUpdate.id}, " +
-          "Feature key id: ${testData.featureKeyToUpdate.id}",
-      )
-    }
 
     testData.mainBranch.findKey(additionKeyName).let {
       it.assert.isNotNull()
       it!!.enTranslation().assert.isEqualTo(additionValue)
     }
+
+    testData.mainKeyToUpdate
+      .enTranslation()
+      .assert
+      .isEqualTo(updatedValue)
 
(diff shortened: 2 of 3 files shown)


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/branching/merging/BranchMergeExecutor.kt
```diff
@@ -175,7 +175,7 @@ class BranchMergeExecutor(
   }
 
   private fun persistAfterMerge(key: Key) {
-    key.translations.filter { it.id == 0L }.forEach { translationService.save(it) }
+    key.translations.forEach { translationService.save(it) }
```
dkrizan, line 178: Is this necessary ? Dirty-checking flushes at transaction boundary regardless of whether the translation was saved.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/branching/BranchMergeServiceTest.kt
```diff
@@ -80,6 +81,12 @@ class BranchMergeServiceTest : AbstractSpringTest() {
     }
     testData = BranchMergeTestData()
     testDataService.saveTestData(testData.root)
+
+    // Wait for all async BranchMetaUpdater.snapshot() calls triggered by saveTestData
```
dkrizan, line 85: useless comments, the one above fun is enough

JanCizmar:
Thanks for the fix! However, I think this PR reveals a deeper design issue worth addressing.

**The core problem:** `BranchMetaUpdater.snapshot()` is `@Async`, which means the branch revision increment is eventually consistent. If the async method doesn't complete (crash, thread pool exhaustion, etc.), we get corrupted data — the revision won't reflect reality, and merges could apply on stale data.

**The async is unnecessary.** The method does 1 SELECT + 1 UPDATE on a single row — there's no performance justification for making it async. And firing it per entity change is wasteful — if a transaction modifies 100 translations, we get 100 async revision bumps when all we need to know is "did the branch change since the dry-run?"

**Proposed rewrite:** Instead of the per-entity async Hibernate event listener, listen for `OnProjectActivityEvent` and bump the revision there. This event:
- Fires **once per transaction** (inside `BeforeTransactionCompletionProcess`)
- Is **synchronous** — runs before commit, so the revision bump is flushed atomically with the data changes
- Has `modifiedEntities` which tells you what entity types were changed

This would eliminate the race conditions entirely, remove the need for `waitForBranchRevisionsToSettle` in tests, and be more efficient (one bump per branch per transaction instead of N async calls).

**One thing to verify:** whether `OnProjectActivityEvent` is published for all code paths that modify branch content (imports, batch operations, etc.). If some paths bypass the activity system, those would miss the revision bump.

<details>
<summary>AI prompt for implementation</summary>

```
Rewrite the branch revision tracking in Tolgee so it's synchronous and per-transaction instead of async and per-entity.

## Current behavior (to remove)
- `BranchMetaUpdater.snapshot()` is an `@Async @Transactional` method that increments `branch.revision++`
- `BranchContentEventListener` listens to Hibernate entity lifecycle events (OnEntityPrePersist, OnEntityPreUpdate, OnEntityPreDelete, OnEntityCollectionPreUpdate) and calls `snapshot()` for each BranchMergeableEntity change
- This fires N async calls for N entity changes in a single transaction

## New behavior
- Create a new `@EventListener` that listens for `OnProjectActivityEvent` (see `InterceptedEventsManager.kt` for how it's published — it fires once per transaction inside `BeforeTransactionCompletionProcess`, before commit)
- In the listener, iterate `event.modifiedEntities` to find entities that implement `BranchMergeableEntity`
- Collect the distinct branch IDs from those entities
- Bump `branch.revision++` once per affected branch
- This runs synchronously within the same transaction, so the revision is guaranteed to be flushed with the data changes

## Files to modify
- Remove or gut `BranchMetaUpdater.kt` and `BranchContentEventListener.kt`
- Create a new listener for `OnProjectActivityEvent` that does the revision bump
- Remove all `waitForBranchRevisionsToSettle` calls and the method itself from `BranchMergeServiceTest.kt` and `BranchControllerMergingTest.kt` — they should no longer be needed
- Remove `waitFor` imports that were only used for revision settling

## Important
- Verify that all code paths modifying branch content (imports, batch operations, merge itself) go through the activity system and would trigger `OnProjectActivityEvent`. If any bypass it, those need a separate revision bump.
- The `modifiedEntities` map is keyed by entity class (`KClass<out EntityWithId>`) — you need to check which of those classes implement `BranchMergeableEntity` and extract the branch ID from them.
- Keep the `persistAfterMerge` fix from this PR (saving all translations, not just new ones) — that's a separate valid fix for the JPA dirty-checking issue.
```

</details>

PR #3595, opened 2026-04-08
fix: QueryBase perf optimization 

bdshadow:
## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable translation filtering, sorting, cursor pagination; trash listing now shows who deleted an entry and when; label filters return correct union/OR results.

* **Performance**
  * Translation list queries optimized for large projects and many languages for faster, more stable page loads and benchmarks.

* **Documentation**
  * Added developer notes describing the translation-view query refactor.

* **Other**
  * Language multiselect no longer enforces the previous selectable-languages limit in the UI.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/AbstractPartialFixturePerfTestBase.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/AbstractPartialFixturePerfTestBase.kt
@@ -1,0 +1,154 @@
+package io.tolgee.api.v2.controllers.translations.v2TranslationsController
+
+import io.tolgee.AbstractSpringTest
+import io.tolgee.development.testDataBuilder.builders.ProjectBuilder
+import io.tolgee.development.testDataBuilder.data.BaseTestData
+import io.tolgee.dtos.cacheable.LanguageDto
+import io.tolgee.model.Language
+import io.tolgee.model.enums.TranslationState
+import io.tolgee.repository.LanguageRepository
+import io.tolgee.util.Logging
+import io.tolgee.util.executeInNewTransaction
+import io.tolgee.util.logger
+import org.junit.jupiter.api.BeforeAll
+import org.springframework.beans.factory.annotation.Autowired
+
+/**
+ * Shared base for all diagnostic "partial fixture" perf tests.
+ *
+ * The base sets up a project with [LANGUAGE_COUNT] languages and inserts [KEY_COUNT] keys, each
+ * translated only in the first [TRANSLATED_LANGUAGE_COUNT] languages. Subclasses then implement
+ * [setupFilterSpecificFixture] to add whatever extra data their filter requires (tags, outdated
+ * translations, QA issues, suggestions, …).
+ *
+ * This split lets us **isolate each filter in its own class**: the database for a given test
+ * class contains only the rows relevant to the filter being measured, so query plans, cache, and
+ * timing are not polluted by unrelated fixture data. Running one class is enough to reproduce and
+ * diagnose a perf problem for one specific filter.
+ *
+ * All concrete subclasses should:
+ *  - be annotated `@Disabled` (diagnostic — not for CI)
+ *  - be annotated `@CleanDbBeforeClass`
+ *  - implement [setupFilterSpecificFixture] to populate filter-specific data
+ *  - optionally override [configureProject] to flip project-level flags (e.g. `useQaChecks`)
+ */
+@Suppress("LateinitVarOverridesLateinitVar")
+abstract class AbstractPartialFixturePerfTestBase :
+  AbstractSpringTest(),
+  Logging {
+  companion object {
+    const val KEY_COUNT = 10_000
+    const val LANGUAGE_COUNT = 20
+    const val TRANSLATED_LANGUAGE_COUNT = 2
+    const val PAGE_SIZE = 20
+    const val WARMUP_RUNS = 1
+    const val MEASURED_RUNS = 3
+  }
+
+  @Autowired
+  protected lateinit var languageRepository: LanguageRepository
+
+  protected var projectId: Long = 0L
+  protected lateinit var languages: List<Language>
+
+  @BeforeAll
+  fun setupFixture() {
+    val totalSetup = System.currentTimeMillis()
+    val keyCreateMs = measureMs { createProjectAndKeys() }
+    val extraMs = measureMs { executeInNewTransaction { setupFilterSpecificFixture() } }
+    val totalMs = System.currentTimeMillis() - totalSetup
+
+    logger.info("=".repeat(72))
+    logger.info("${this::class.simpleName} FIXTURE READY")
+    logger.info("  Project id=$projectId")
+    logger.info("  Languages: $LANGUAGE_COUNT (only first $TRANSLATED_LANGUAGE_COUNT translated)")
+    logger.info("  Keys: $KEY_COUNT")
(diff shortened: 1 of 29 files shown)


bdshadow:
@coderabbitai please review

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/AbstractPartialFixturePerfTestBase.kt
```diff
@@ -0,0 +1,203 @@
...
+          this.project = managedProject
+        }
+      entityManager.persist(key)
+      pendingKeys.add(key)
+
+      if (pendingKeys.size >= batchSize) {
+        entityManager.flush()
+        for (k in pendingKeys) {
+          for (language in translatedLanguages) {
+            val translation =
```
JanCizmar: can't we use standard `testData` way of test data creation for this?

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/QueryGlobalFiltering.kt
```diff
@@ -168,70 +246,125 @@ class QueryGlobalFiltering(
   }
 
   private fun filterTag() {
```
JanCizmar, line 248: This method is huge. I don't know understand why the logic had to change so much, but probably there was some good reason. However, I would consider move this whole logic to separate file and split it to multiple nicely-named methods, so it's readable. The same pays for `filterNoTag`. The methods look very similar to me and maybe the code can be somehow reused.

JanCizmar (dismissed):

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/AbstractPartialFixturePerfTestBase.kt
bdshadow: done

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/QueryGlobalFiltering.kt
bdshadow: On main `filterTag` does `root.join(Key_.keyMeta).join(KeyMeta_.tags)`  a left join on the main query that multiplies rows (one key with 3 tags produces 3 rows), which forces the GROUP BY that the whole refactor was designed to eliminate. Here it is repalced with an EXISTS subquery, which keeps the main query row-per-key and removes the GROUP BY dependency. This is what made `filterTag` go from 278ms to 17ms (16x faster).

e2e/scripts/baseQuery/perf-test.sh
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env bash
```
dkrizan, line 1: is this meant to be commited ?

backend/data/src/main/kotlin/io/tolgee/service/queryBuilders/translationViewBuilder/QueryGlobalFiltering.kt
JanCizmar: Then I would at least consider moving it to another file. :)

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translations/v2TranslationsController/TranslationsControllerSnapshotTest.kt
```diff
@@ -0,0 +1,187 @@
...
+ */
+@SpringBootTest
+@Transactional
+class TranslationsControllerSnapshotTest : AbstractSpringTest() {
+  @Autowired
+  private lateinit var translationViewDataProvider: TranslationViewDataProvider
+
+  @Test
+  fun `produces stable snapshot for a comprehensive scenario`() {
+    val testData = TranslationsTestData()
```
dkrizan: create custom TestData to avoid initing it during test

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/KeyTrashFilterTest.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+    // Add a few tagged keys for tag-filter trash tests
+    testData.addFewKeysWithTags()
+    this.projectSupplier = { testData.project }
+  }
+
+  @ProjectJWTAuthTestMethod
+  @Test
+  fun `filterDeletedByUserId returns only keys deleted by the specified user`() {
+    testDataService.saveTestData(testData.root)
```
dkrizan: this can probably be called in `setup()`

webapp/src/views/projects/translations/context/services/useTranslationsService.tsx
```diff
@@ -286,21 +279,6 @@ export const useTranslationsService = (props: Props) => {
   };
 
   const setLanguages = (value: string[] | undefined) => {
-    const limit = config.translationsViewLanguagesLimit;
-    if (value && value.length > limit) {
```
dkrizan, line 290: not sure whether we want to remove this completely 
- either remove the config property too (in TolgeeProperties.kt) 
- or if you just want to disable it in production, set its default value to `null` in TolgeeProperties.kt (and handle nol limit in tsx component properly)

e2e/scripts/baseQuery/perf-test.sh
bdshadow: yes, i think it's convenient to run it like this, just like we have the script for the batch jobs, nexto this one

webapp/src/views/projects/translations/context/services/useTranslationsService.tsx
bdshadow: set it to `-1` in production

ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/TranslationsControllerLabelFilterTest.kt
```diff
@@ -0,0 +1,134 @@
...
+      .andAssertThatJson {
+        node("page.totalElements").isEqualTo(0)
+      }
+  }
+
+  @ProjectJWTAuthTestMethod
+  @Test
+  fun `multiple labels for same language combine with OR semantics`() {
+    // Assign unassignedLabel to the english translation of "second key"
+    testData.secondKeyEnTranslation.labels.add(testData.unassignedLabel)
```
dkrizan: I would suggest creating a new key with assigned labels in TestData, so you can use it directly a no need to call in each test
```
    userAccount = testData.user
```

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/v2KeyController/KeyTrashFilterTest.kt
```diff
@@ -0,0 +1,180 @@
...
+          node("[0].name").isEqualTo("Key with tag")
+        }
+        node("page.totalElements").isEqualTo(1)
+      }
+  }
+
+  @ProjectJWTAuthTestMethod
+  @Test
+  fun `trashed listing combined with filterKeyName returns only trashed matching keys`() {
+    val deletedTarget = keyService.get(testData.project.id, "key 05", null)
```
dkrizan: I would suggest making those keys public so you can access them without calling a service

dkrizan (approved):

PR #3596, opened 2026-04-08
feat: Translation Memory Management

### LINKED ISSUE #3453: Translation Memory Management
### 1. Problem
**Inconsistent translations and wasted effort across projects**

**Core issues:**
Organizations managing multiple projects (web + mobile, Android + iOS, product variants) face critical translation challenges:
- **Duplicate work:** The same phrases are translated repeatedly in each project. "Add to cart" gets translated 3x across 3 projects — pure waste.
- **Inconsistency:** One project uses "Next", another "Continue" for identical source text. Users perceive this as unprofessional and confusing.
- **Isolated translation memories:** Each project maintains its own TM with no sharing. New projects start from zero, ignoring existing translation work.

**Additional enterprise challenges:**
- **Quality degradation:** Poor translations in TM get propagated through AI and auto-translate features, multiplying errors across the system. (This could be addressed by controlling data quality in the shared TM used for reading.)
- **Legacy migration:** Large companies need to import existing TMX files from other tools and make them accessible across all projects.

**Business problem:** 
To be competitive in the enterprise localization market, Tolgee must support cross-project resource sharing. For enterprise clients, this is essential — including TMX import/export capabilities. Without this, we lose deals to established competitors.

---

### 2. Appetite

**One cycle (4 weeks)**
The frontend reuses UI components from Glossaries and the TM translation view, which should significantly reduce UI work.

---

### 3. Solution

Based on competitive analysis of existing localization platforms, we propose the following design — taking the best patterns while keeping complexity low.

#### Pricing & Availability
- **Free plan:** No shared TM. The existing "similar translations" feature stays, but must be renamed and all references to TM removed — suggestions, links, bulk operations (Pre-translate by TM).
- **Business plan:** Full TM available.

#### TM Architecture
TM is a single object type — it behaves the same everywhere. The only difference is how many projects have access to it.
- When a project is created, a **project TM** is automatically created and assigned only to that project.
- At the organization level, users can create any number of additional TMs and assign them to multiple projects — this is how sharing works. The requirement is a matching source language across projects.

#### Organization Level: TM Management
New section in Organization Settings with two tabs:
- **TAB Shared TMs** — shared TMs created at the organization level
- **TAB Project TMs** — project TMs created automatically. Not shown by default, but users can view their project TM and upload data to it independently.

Capabilities:
- Create named Translation Memory instances (e.g., "Main", "Product A", "Legal")
- Browse TM content (source/target langs table with search)
- Import/export TMX
- Delete entries or entire TM
- See which projects use each TM

#### Project Level: TM Configuration
In Project Settings:
- List of connected TMs with priority ordering
- For shared TMs: configure read/write access per TM
- User can set TM priority — by default shared TM has higher priority

**Rules:**
- A project must always have at least 1 TM — the project TM, which is destroyed only when the project is deleted
- Shared TMs can be deleted manually
- When a project is removed from a shared TM, its historical data stays in the TM, new translations stop being written. User gets the choice: keep or delete this project's records from the shared TM.

> **Opt-in sharing model:** We prefer project TM as an isolated base, with sharing as a conscious opt-in. A single org-wide TM shared automatically causes noise in practice — translations from unrelated projects (marketing web, mobile app, developer docs each with different tone of voice) mix together. If we hit major problems during implementation, we can switch approach.

#### Writing to TM
- **Trigger:** Translation is saved to TM automatically on every save. Each record always links to the specific translation (translation ID) — so when a translation changes, the TM record is updated instead of creating duplicates.
- **Project setting — toggle:** "Save to TM only Reviewed translations". Default off — all translations are saved.
- **Where it writes:** Based on the project's TM assignments — records go to TMs that have write access enabled. If the project has multiple write-enabled TMs, it writes to all of them.
- **On translation change:** The TM record is updated via the translation ID link — no duplicates.

#### Reading from TM (Suggestions, AI prompt, Auto-translate)

**Match logic — carried over from current implementation:** suggestions show from 50% match upward. For each suggestion: translated text, source text, match percentage, key name.

1. **Inline suggestions in editor:** When translating a key, TM suggestions show automatically. Clicking applies the suggestion to the editor.
2. **Pre-translate (batch):** Bulk pre-fill of a language or project from TM. Configurable minimum match threshold. Uses existing logic.

**Which TMs are searched:** All TMs assigned to the project. Results are sorted by match percentage. In project settings, user can set TM priority — the prioritized TM is searched first when match percentage is equal.

#### TM Data Format
Import and export TM in **TMX 1.4b** format — the industry standard for localization, compatible with all major CAT tools.

**Each TM record stores:** translation language, source/target text, key name (context), creation date, author, last change date, and last change author.

> Detailed format specification to be explored during Shape-up.

#### TMX Import & Export
- Import TMX files directly into an existing TM. Import is simple — does not resolve conflicts record by record.
- **Confirmation dialog:** Before import, user decides whether to overwrite existing records or ignore them (keep current).
- **Export:** Download the entire TM as a TMX file.

#### Conflict Resolution
When selecting a TM suggestion, the system decides progressively:
1. **Match percentage** — the suggestion with the highest match percentage takes precedence
2. **TM priority** — when match is equal, the TM priority order set in the project decides

---

### 4. Basic Drawings
**Sketches to discuss**
https://www.figma.com/design/H6BDCYRjNFgXymTtnkQriW/Tasks-2026?node-id=1-31&t=CfpgqhcDb4yunQnU-1

Image
Image

---

### 5. Rabbit Holes
- **Performance for batch operations** — Auto-translate 40k keys x 5 languages must not be slower than the current solution
- **Deduplication** — Handled via translation ID linking: TM records update on change instead of creating duplicates
- **Merging TMs** — Admin action to promote translations to shared TM — keep simple
- **Migration** — Strategy for existing data and auto-translate configs
- **Priority logic** — Search all assigned TMs, rank by match percentage first, then by TM priority
- **TMX format specification** — Detailed mapping of Tolgee fields to TMX properties to explore during Shape-up

---

### 6. No-Goes
- Avoid big complexity and scope, keep it simple
- No record-by-record conflict resolution during import (simple overwrite/ignore dialog)
- No contextual/AI-driven match evaluation

---

### 7. Success Criteria

- More business/enterprise customers
- Better AI results (through higher-quality shared TM data)
- Shared TMs are used by our customers
- No performance regression for existing batch operations
- TMX import/export works with standard CAT tools

---

_Research reference: [Translation Memory Platform Comparison 2026](https://claude.ai/public/artifacts/6ebcebea-90f5-436b-8e50-7e6a451df1c1) — competitive analysis of existing localization platforms_

dkrizan:
Adds Translation Memory management: every project gets a project-scoped TM for free; paid orgs additionally get shared TMs they can wire across projects.

## Plan gating

| | Free | Paid (`Feature.TRANSLATION_MEMORY`) |
|---|---|---|
| Project TM (per-project, virtual rows from translations + manual entries) | ✅ | ✅ |
| Shared TM (org-level, multi-project, stored entries) | — | ✅ |
| TM management UI, TM content browser, TMX import/export, batch pre-translate | — | ✅ |
| Suggestion endpoint | ✅ (project TM only) | ✅ (project + shared) |
| Auto-translate / batch pre-translate via TM | ✅ (exact-equality on project translations) | ✅ (also consults stored entries on every readable TM) |

Plan-aware filter lives in `TranslationMemoryManagementService.getReadableTmIdsForSuggestions(projectId, organizationId)`.

## Where the OSS / EE split lives

- `backend/data` (OSS): entities, repositories, project-TM auto-create, suggestion entry point.
  - `TranslationMemoryService` — interface; `TranslationMemoryServiceOssImpl` queries virtual rows only.
  - `TmAutoTranslateProvider` — interface used by `AutoTranslationService`. `TmAutoTranslateProviderOssImpl` does an exact-equality lookup on the project's own translations (matches the pre-feature behavior; keeps batch pre-translate in the millisecond range).
- `ee/backend/app` (EE, `@Primary`): shared-TM functionality.
  - `TranslationMemoryServiceEeImpl` adds the UNION over stored entries with assignment-level penalty.
  - `TmAutoTranslateProviderEeImpl` checks stored entries on every readable TM via exact equality, then falls back to the OSS impl.
  - `TranslationMemoryEntryManagementService` — stored-entries CRUD, virtual-row hydration, project-TM content listing.
  - Controllers: `SharedTranslationMemoryController`, `ProjectTranslationMemoryController`, `TranslationMemoryEntryController`, `TranslationMemoryTmxController`.

## Project TM provisioning

- Liquibase changeset `tm-create-project-tms-for-existing-projects` backfills every existing project on migration.
- `ProjectCreationService.createProject()` and `DemoProjectCreator` create one for new projects.
- `TestDataService.ensureProjectTms()` mirrors the auto-create in test fixtures.
- No lazy/event/batch path — every project always has its own TM.

## Things worth a closer look

- **Suggestion SQL** — `AbstractTranslationMemoryService` plus the OSS / EE `baseSelect` overrides. EE impl penalty + UNION is non-trivial.
- **Project-TM content listing** — `TranslationMemoryEntryManagementService.listEntryRows` paginates STORED buckets + VIRTUAL origins (one row per `(sourceText, project_id, key_name)`). Row IDs come from `rowIdentitiesSql` so search/lang filtering applies uniformly.
- **Select-all** — `GET /…/entries/entryIds` returns representative entry IDs per stored bucket so the bulk-delete UI doesn't pull the entire content list.
- **Atomic base-language change** — `EditProjectRequest.unassignConflictingTms` lets the project-edit endpoint detach mismatched shared TMs in the same transaction, gated by an explicit dialog confirmation on the frontend. No feature-gate on the unassign side so an org that loses the TM feature can still rescue itself.
- **TMX round-trip** — `XmlSecurity`-hardened DOM parser + StAX writer; `tuid` preserved and used for keep/override matching; in-batch dedup so a TMX containing repeated `(tuid, lang)` pairs can't insert duplicates.
- **Empty-state wizard** — `webapp/src/ee/translationMemory/components/content/emptyWizard/`. Three on-ramps; each option is its own component.
- **New-entry dialog** — `TranslationMemoryCreateEntryDialog` uses a header multi-select for target languages (replaces the per-row "Add language" flow).

## Test plan

- [x] CI green
- [x] Free org / new project: suggestion panel works against project TM; management UI hidden; pre-translate-by-TM works against project TM (exact-equality on project translations).
- [x] Paid org: create shared TM, assign to projects, drag priority, set penalty, TMX import/export.
- [x] Free-tier rescue: org loses TM feature → user can still change project base language (confirmation dialog drops the mismatched shared assignments inline).
- [ ] Migration: existing projects show their project TM after deploy without manual action.

## Summary by CodeRabbit

* **New Features**
  * Full Translation Memory (TM) system: project & shared TMs, TM browser (flat/stacked), TM suggestion panel in editor, TM-based auto-translate, TMX import/export, create/edit TM UI, project/org TM management, empty-state wizard, and project settings integration.
* **Tests**
  * Expanded unit/integration/e2e and Cypress coverage for TM flows, suggestions, import/export, permissions and UI behaviors.
* **Chores**
  * .gitignore updated to ignore generated E2E fixtures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3596)


dkrizan:
@coderabbitai review

dkrizan:
### Code review

Found 3 issues:

1. The `tm-create-project-tms-for-existing-projects` Liquibase changeset's `insert into translation_memory_project` omits `id`, `created_at`, `updated_at`, but the table declares all three as `nullable="false"` with no defaults. The `insert into translation_memory` just above correctly supplies `nextval('hibernate_sequence')` and `now(), now()` — this one will fail with a NOT NULL violation when applied to a fresh DB.

https://github.com/tolgee/tolgee-platform/blob/d24601d8a89cc1e626504fb1cacd05d478a7a749/backend/data/src/main/resources/db/changelog/schema.xml#L5505-L5513

2. The OSS suggestion SQL reintroduces `base_t.text % :baseTranslationText` — the trigram operator that PR #3631 explicitly rolled back ten days ago because it caused multi-second per-call latency and multi-hour MT-batch run times on long ICU plural strings. The new `TmAutoTranslateProviderOssImpl.getAutoTranslatedValue` then routes every auto-translate/batch pre-translate key through `getSuggestions(...)` instead of the previous exact-equality `getTranslationMemoryValue`, so this query now runs in the per-key hot path. Worth confirming the `text %` plan on production-shaped data (long plurals, large `translation` tables) before merge.

https://github.com/tolgee/tolgee-platform/blob/d24601d8a89cc1e626504fb1cacd05d478a7a749/backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationMemoryServiceOssImpl.kt#L58-L64

https://github.com/tolgee/tolgee-platform/blob/d24601d8a89cc1e626504fb1cacd05d478a7a749/backend/data/src/main/kotlin/io/tolgee/service/translationMemory/TmAutoTranslateProviderOssImpl.kt#L24-L33

3. `(resultList.firstOrNull() as Array<*>?)?.get(10) as Long?` will throw `ClassCastException` at runtime on PostgreSQL — `count(*) over()` is `bigint` and comes back via JDBC as `java.math.BigInteger` (or `BigDecimal` depending on driver), not `Long`. The same file's `mapRow` already uses the safe `(row[N] as Number).toLong()` form; the same pattern was flagged on PR #3540.

https://github.com/tolgee/tolgee-platform/blob/d24601d8a89cc1e626504fb1cacd05d478a7a749/backend/data/src/main/kotlin/io/tolgee/service/translation/AbstractTranslationMemoryService.kt#L105-L110

webapp/src/ee/translationMemory/views/TranslationMemoryCreateEntryDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+                  fullWidth
+                  data-cy="tm-entry-target-text"
+                />
+              </Box>
+            </div>
+          );
+        })}
+
+        <StyledActions>
+          <Button onClick={onClose}>{t('global_cancel_button')}</Button>
```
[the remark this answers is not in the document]
dkrizan: True, but this should apply only for new keys. This one is existing.

e2e/cypress/compounds/translationMemories/E2TranslationMemoryCreateEditDialog.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  checkBaseLanguage(language: string) {
+    gcy('base-language-select').should('contain', language);
+  }
+
+  toggleAssignedProject(projectName: string) {
+    gcy('tm-add-project-autocomplete').find('input').click();
+    cy.get('li[role="option"]').contains(projectName).click();
+  }
```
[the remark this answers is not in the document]
dkrizan: Fixed in 76a99b191. `AssignedProjectsEditor`'s autocomplete now renders options with `data-cy="tm-add-project-option"` and the e2e helper selects via `gcy('tm-add-project-option').contains(projectName)`. Resolving.

webapp/src/ee/translationMemory/hooks/useTmExport.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  const triggerExport = () => {
+    exportLoadable.mutate(
+      {
+        path: { organizationId, translationMemoryId },
+      },
+      {
+        async onSuccess(response) {
+          return downloadExported(response as unknown as Response, tmName);
```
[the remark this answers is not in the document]
dkrizan: Skipping — the proper fix (overloading `useApiMutation`'s signatures for the `rawResponse: true` case) is a cross-cutting refactor of `tg.service/http/useQueryApi` that affects every caller in the codebase. Out of scope for this PR. The `as unknown as Response` is the documented escape hatch the codebase uses elsewhere when `rawResponse` is set.

webapp/src/ee/translationMemory/views/ProjectTmAssignmentDialog.tsx
```diff
@@ -0,0 +1,112 @@
...
+export const ProjectTmAssignmentDialog: React.VFC<Props> = ({
+  open,
+  onClose,
+  onFinished,
+  projectId,
+  assignment,
+}) => {
+  const { t } = useTranslate();
+  const [readAccess, setReadAccess] = useState(assignment.readAccess);
+  const [writeAccess, setWriteAccess] = useState(assignment.writeAccess);
```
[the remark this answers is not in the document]
dkrizan: Not an issue here — the only caller (`ProjectSettingsTranslationMemory.tsx`) renders the dialog conditionally: `{settingsTmId !== null && settingsAssignment?.type !== 'PROJECT' && !isOrgMaintainer && settingsAssignment && (<ProjectTmAssignmentDialog ... />)}`. The dialog fully unmounts when `settingsTmId` is cleared and remounts with a fresh `useState` initialiser on the next open, so stale state across different assignments isn't reachable. (Same pattern applies to `TmWriteOnlyReviewedDialog` and `TranslationMemoryImportDialog`, which CodeRabbit flagged similarly.)

webapp/src/ee/translationMemory/views/TranslationMemoryCreateEntryDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+        />
+      );
+      onFinished();
+    } catch {
+      messageService.error(
+        <T
+          keyName="translation_memory_create_entry_error"
+          defaultValue="Failed to create entries"
+        />
+      );
```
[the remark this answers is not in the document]
dkrizan: Acknowledged as a follow-up — proper handling needs `Promise.allSettled` plus per-language error UI (mark the still-failing inputs, show partial-success toast, keep their text in place so the user can retry only the failed ones without recreating the succeeded ones). Tracking separately rather than inflating this PR.

webapp/src/ee/translationMemory/views/TranslationMemoryView.tsx
```diff
@@ -0,0 +1,244 @@
...
+import { ProjectLink } from 'tg.component/ProjectLink';
+
+const StyledSharedWith = styled(Box)`
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  color: ${({ theme }) => theme.palette.text.secondary};
+`;
+
+export const TranslationMemoryView = () => {
```
bdshadow, line 29: nitpick. A feature check can be added for this view too. Smth like
```
  const { isEnabled } = useEnabledFeatures();
  if (!isEnabled('TRANSLATION_MEMORY')) {
    return <Redirect to={LINKS.ORGANIZATION_TRANSLATION_MEMORIES.build({ ... })} />;
  }
```

Just in case the view is accessed directly by url, it will be broken. Nothing critical, you can't get here otherwise by navigating in UI

webapp/src/ee/translationMemory/views/TranslationMemoryCreateEntryDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const handleSave = async () => {
+    const payload = selectedTags
+      .map((tag) => ({ tag, text: textByTag[tag]?.trim() ?? '' }))
+      .filter((entry) => entry.text.length > 0);
+    if (!sourceText.trim() || payload.length === 0) return;
+
+    setSaving(true);
+    try {
+      for (const entry of payload) {
+        await createMutation.mutateAsync({
```
bdshadow: Here it is possible that the first failure aborts the loop (await rejection bubbles to the catch). Entries POSTed before the failure are already persisted on the server. If the user retries, the already-persisted entries will get re-POSTed; the backend's (translation_memory_id, source_text, target_language_id, tuid) uniqueness will reject them with translation_memory_entry_already_exists — user sees the same generic toast again with no way to know they're stuck on duplicates. Generic toast "Failed to create entries" - doesn't distinguish which langs succeeded vs. failed.

A bulk endpoint on the backend would be a good solution here. But leaving it to you here, since the probability of it is pretty low (though it's wrapped with try-catch - meaning it was considered a possible scenario)

webapp/src/ee/translationMemory/components/list/TranslationMemoryListItemMenu.tsx
```diff
@@ -0,0 +1,193 @@
...
+      ),
+      message: (
+        <T
+          keyName="delete_translation_memory_confirmation_message"
+          defaultValue="This will permanently delete this translation memory and all its entries. This action cannot be undone."
+        />
+      ),
+      hardModeText: translationMemory.name.toUpperCase(),
+      onConfirm() {
+        deleteMutation.mutate({
```
bdshadow: do we want to show/confirm that it was successfully deleted?

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5348,4 +5348,204 @@
...
+            <sql>DROP INDEX CONCURRENTLY IF EXISTS ix_translation_key_lang;</sql>
+        </rollback>
+    </changeSet>
+
+    <changeSet author="dkrizan" id="tm-entry-tm-source-index" runInTransaction="false">
+        <comment>Composite index for findByTranslationMemoryIdAndSourceTexts during
+            content-browser page hydration. Existing single-column indexes only covered
+            translation_memory_id, leaving the source_text filter as a post-fetch step.</comment>
+        <sql dbms="postgresql">
+            CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_tm_entry_tm_source
```
bdshadow, line 5560: btree on unbounded TEXT - Postgres btree row limit is ~2700 bytes.
DTO `CreateTranslationMemoryEntryRequest`: `@field:Size(max = 10000)` on `sourceText`. The cap is 10,000 chars.

Postgres btree has a hard per-row limit of ~2704 bytes (1/3 of the 8 KB page size). UTF-8 averages:
  - ASCII -> 1 B/char -> safe at <= 2704 chars
  - Latin (accented) -> ~1.5 B/char -> safe at <= ~1800 chars
  - CJK / emoji -> 3-4 B/char -> safe at <= ~680 chars

It means that insert into `ix_tm_entry_tm_source` fails with ERROR: index row size <N> exceeds btree version 4 maximum 2704 for index "ix_tm_entry_tm_source".
Claude offers here:
1. Tighten the DTOs: `@field:Size(max = 2000)` (or 1000 to be CJK-safe) on sourceText/targetText. Plus a parallel length check in TmxParser (drop or error on oversize segments).
2. Change the index expression - the b-tree exists for equality during `findByTranslationMemoryIdAndSourceTexts` lookups. Replace with (translation_memory_id, md5(source_text)) or (translation_memory_id, substr(source_text, 1, 1000)) and adjust the query accordingly.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/translationMemory/TranslationMemoryEntryManagementService.kt
```diff
@@ -0,0 +1,724 @@
...
+
+    val rows =
+      rowIds.map { id ->
+        when (id.kind) {
+          TmRow.Kind.STORED ->
+            TmRow(
+              sourceText = id.sourceText,
+              kind = TmRow.Kind.STORED,
+              originId = id.originId,
+              entries = storedByBucket[id].orEmpty(),
```
bdshadow: here jpa TranslationMemoryEntry entities, reache the controller layer; map to DTO (probably via modelAssebler as we have it in other places)

webapp/src/ee/translationMemory/components/content/TranslationMemoryEntriesList.tsx
bdshadow, line 1: 544 lines is too big. Worth splitting it. Looks like several hooks and a couple of components can be extracted from here.

ee/backend/app/src/main/kotlin/io/tolgee/ee/data/translationMemory/CreateTranslationMemoryEntryRequest.kt
```diff
@@ -0,0 +1,24 @@
+package io.tolgee.ee.data.translationMemory
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Size
+
+class CreateTranslationMemoryEntryRequest {
```
bdshadow: it's absolutely the same as the `UpdateTranslationMemoryEntryRequest`. In case of update, we need to control 2 places. I'd make it one class. Leaving it up to you

ee/backend/app/src/main/kotlin/io/tolgee/ee/data/translationMemory/CreateSharedTranslationMemoryRequest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.Valid
+import jakarta.validation.constraints.Max
+import jakarta.validation.constraints.Min
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Size
+
+class CreateSharedTranslationMemoryRequest {
```
bdshadow: same here.
it's the same as the `UpdateSharedTranslationMemoryRequest`. In case of update, we need to control 2 places. I'd make it one class. Leaving it up to you

bdshadow (changes requested):

webapp/src/ee/translationMemory/views/TranslationMemoryCreateEntryDialog.tsx
dkrizan: thanks, for sure this needs to be processed in batch

backend/data/src/main/kotlin/io/tolgee/repository/translationMemory/TranslationMemoryEntryRepository.kt
```diff
@@ -0,0 +1,138 @@
...
+import io.tolgee.model.translationMemory.TranslationMemoryEntry
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.Pageable
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
+import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.query.Param
+import org.springframework.stereotype.Repository
+
+@Repository
```
bdshadow: annotation can be removed. `JpaRepository` parent is enough.

same in other repos

backend/data/src/main/kotlin/io/tolgee/repository/translationMemory/TranslationMemoryEntryRepository.kt
```diff
@@ -0,0 +1,138 @@
...
+        and (:targetLanguageTags is null or e.target_language_tag = any(string_to_array(:targetLanguageTags, ',')))
+        and (
+          :search is null
+          or lower(e.source_text::text) like lower('%' || :search || '%')
+          or lower(e.target_text::text) like lower('%' || :search || '%')
+        )
+      order by e.source_text, e.target_language_tag, e.id desc
+    """,
+    countQuery = """
+      select count(*) from translation_memory_entry e
```
bdshadow: you can drop countQueries, when they're the same as the actual query. I'd do it - less code, less chance to update it only at one place and have a bug in future

same in other repos

backend/data/src/main/kotlin/io/tolgee/repository/translationMemory/TranslationMemoryEntryRepository.kt
```diff
@@ -0,0 +1,138 @@
...
+        and (:targetLanguageTags is null or e.target_language_tag = any(string_to_array(:targetLanguageTags, ',')))
+        and (
+          :search is null
+          or lower(e.source_text::text) like lower('%' || :search || '%')
+          or lower(e.target_text::text) like lower('%' || :search || '%')
+        )
+    """,
+    nativeQuery = true,
+  )
+  fun findByTranslationMemoryIdPaged(
```
bdshadow: some methods, like `findByTranslationMemoryIdPaged` and `findDistinctSourceTextsPaged` are unused (please check other repositories too).

Please remove redundant code. Such query repos are actually taking pretty a lot of time during the spring initilization phase. If not needed, better to remove

backend/data/src/main/kotlin/io/tolgee/repository/translationMemory/TranslationMemoryEntryRepository.kt
```diff
@@ -0,0 +1,138 @@
...
+      order by e.source_text
+    """,
+    countQuery = """
+      select count(*) from (
+        select distinct e.source_text
+        from translation_memory_entry e
+        where e.translation_memory_id = :translationMemoryId
+          and (
+            :search is null
+            or lower(e.source_text::text) like lower('%' || :search || '%')
```
bdshadow: `lower()` is used in multiple places and leads to full scan. Maybe creating an index make sense here:

`CREATE INDEX ON translation_memory_entry (translation_memory_id,
  lower(source_text) text_pattern_ops)`

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/translationMemory/TranslationMemoryEntryManagementService.kt
bdshadow, line 1: leaving it up to you, but 724 lines is too big. Maybe there is a way to split it better

backend/data/src/main/kotlin/io/tolgee/repository/translationMemory/TranslationMemoryProjectRepository.kt
```diff
@@ -0,0 +1,78 @@
...
+   * reason as [findByProjectIdAndWriteAccessTrue] — callers may invoke this while an outer save
+   * is mid-flight; we don't want the query to trigger a premature flush that exposes transient
+   * entities on the in-progress Translation.
+   */
+  @QueryHints(
+    value = [QueryHint(name = "org.hibernate.flushMode", value = "COMMIT")],
+  )
+  fun findByProjectIdAndReadAccessTrue(projectId: Long): List<TranslationMemoryProject>
+
+  fun findByTranslationMemoryId(translationMemoryId: Long): List<TranslationMemoryProject>
```
bdshadow, line 25: this method is called from multiple places and almost everywere (except controller) it is used to fetch the Project entity. This is a potential N+1 problem. I think more beneficial is to fetch them at once. Smth like:
```kt
@Query("select tmp.project.id from TranslationMemoryProject tmp where
  tmp.translationMemory.id = :tmId and tmp.writeAccess = true")
  fun findWriteAccessProjectIds(@Param("tmId") tmId: Long): List<Long>
```

webapp/src/ee/translationMemory/components/form/AssignedProjectsEditor.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const { preferredOrganization } = usePreferredOrganization();
+  const organizationId = preferredOrganization!.id;
+
+  const [projectSearch, setProjectSearch] = useState('');
+  const [projectSearchDebounced] = useDebounce(
+    projectSearch,
+    PROJECT_SEARCH_DEBOUNCE_MS
+  );
+
+  const projectsLoadable = useApiInfiniteQuery({
```
bdshadow, line 54: Seems like very same project picker is used in the glossary's `AssignedProjectsSelect.tsx`. Please check if it's worth moving to a common place and reuse

webapp/src/ee/translationMemory/components/list/TranslationMemoriesEmptyListMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+type Props = {
+  loading?: boolean;
+  onCreateClick?: () => void;
+};
+
+export const TranslationMemoriesEmptyListMessage: VFC<Props> = ({
+  loading,
+  onCreateClick,
+}) => {
+  if (loading) {
```
bdshadow: Looks like this can be handled by `EmptyState` the same way as it is done in the `GlossariesEmptyListMessage`

webapp/src/ee/translationMemory/views/TranslationMemoriesListView.tsx
```diff
@@ -0,0 +1,236 @@
...
+            {isEmpty ? (
+              <TranslationMemoriesEmptyListMessage
+                loading={tms.isFetching}
+                onCreateClick={canCreate ? onCreate : undefined}
+              />
+            ) : (
+              <>
+                {showToolbar && (
+                  <Box display="flex" gap={2} alignItems="center" mb={2}>
+                    <TextField
```
bdshadow: `BaseOrganizationSettingsView` already accepts an `onSearch` prop and it can be reused instead of creating custom `TextField` here

webapp/src/ee/translationMemory/components/content/TranslationMemoryImportDialog.tsx
```diff
@@ -0,0 +1,221 @@
...
+      <T
+        keyName="translation_memory_import_mode_override_description"
+        defaultValue="New entries are always added. Previously imported entries are replaced with new translations."
+      />
+    ),
+    dataCy: 'tm-import-mode-override',
+  },
+];
+
+export const TranslationMemoryImportDialog: React.VFC<Props> = ({
```
bdshadow, line 70: please check if it can be somehow unified together with `GlossaryImportDialog`. Same is for the `TranslationCell` and `GlossaryListTranslationCell`, they look. very similar

bdshadow (changes requested):

backend/data/src/main/kotlin/io/tolgee/repository/translationMemory/TranslationMemoryEntryRepository.kt
dkrizan: Implemented as GiST trigram indexes on `lower(source_text)` and `lower(target_text)` (matching the existing `tm_entry_source_text_gist_trgm` precedent). All current queries are infix LIKE (`lower(...) like "%search%"`) — `text_pattern_ops` btree only accelerates prefix LIKE, so trigram is the matching tool here. Happy to revisit if you had a specific equality / prefix lookup in mind.

webapp/src/ee/translationMemory/components/list/TranslationMemoriesEmptyListMessage.tsx
dkrizan: I created a reusable component.
Worth to mention - this is only shown when no project is in organization, since this list includes also all project TMs (which are auto-created for all projects in organization).

webapp/src/ee/translationMemory/components/content/TranslationMemoryImportDialog.tsx
dkrizan: The shared skeleton is ~25 lines; the rest is genuinely different per resource, so not worth unifying.

webapp/src/ee/translationMemory/components/form/AssignedProjectsEditor.tsx
dkrizan: Thanks, but sharing the component would mean a handful of render-slot props that roughly equal the original code.

bdshadow (approved):

PR #3599, opened 2026-04-09
fix: save Translation before persisting QA issues

Anty0:
## Summary

- Fix `TransientPropertyValueException` when QA checks run on a key-language pair with no existing translation
- `getOrCreate()` can return a new unsaved `Translation`; the code was saving QA issues referencing it before persisting the translation itself
- Move `translationService.save(translation)` before `replaceIssuesForTranslation()` so the entity is managed first
- Add regression test that reproduces the exact scenario
- Rename `QaBatchServiceTest` → `QaCheckBatchServiceTest` to match the service under test
- Extract `runChecks()` helper to reduce test duplication

## Summary by CodeRabbit

* **Bug Fixes**
  * QA checks now create missing translations and persist QA issues for them.
  * QA recheck behavior adjusted to use existing language data during base-language changes.

* **Tests**
  * Improved QA tests: renamed/organized suite, centralized check invocation, added coverage for missing-translation scenario.
  * API translation list test updated to expect an additional key when no QA filter is applied.

### DIFF
--- a/ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/qa/QaCheckBatchServiceTest.kt
+++ b/ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/qa/QaCheckBatchServiceTest.kt
@@ -56,19 +56,6 @@ class QaCheckBatchServiceTest : AuthorizedControllerTest() {
     userAccount = null
   }
 
-  private fun refetchEntities() {
-    testData.testKey = entityManager.find(Key::class.java, testData.testKey.id)
-    testData.frTranslation = entityManager.find(Translation::class.java, testData.frTranslation.id)
-  }
-
-  private fun runChecks(languageId: Long = testData.frTranslation.language.id) {
-    qaCheckBatchService.runChecksAndPersist(
-      testData.project.id,
-      testData.testKey.id,
-      languageId,
-    )
-  }
-
   @Test
   @Transactional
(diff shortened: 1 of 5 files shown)


ee/backend/tests/src/test/kotlin/io/tolgee/ee/service/qa/QaCheckBatchServiceTest.kt
```diff
@@ -117,28 +117,46 @@ class QaBatchServiceTest : AuthorizedControllerTest() {
...
       val issuesAfter = qaIssueRepository.findAllByTranslationId(testData.frTranslation.id)
       assertThat(issuesAfter).isEmpty()
     }
   }
 
+  @Test
+  fun `creates QA issues for key with no existing translation`() {
+    // Delete the French translation so getOrCreate() will create a new one
+    executeInNewTransaction(platformTransactionManager) {
+      val translation = entityManager.find(Translation::class.java, testData.frTranslation.id)
```
dkrizan: Could be possibly prepared in testData ? 🙏 
We should avoid preparing data in test logic.

Anty0: Fixed 2 times :3

dkrizan (approved):

PR #3605, opened 2026-04-14
fix: bind email-targeted invitations to the invitee's email

bdshadow:
When an invitation is sent to a specific email address, only that user can now accept it (case-insensitive match). Link invitations without an email remain open to anyone.

## Summary by CodeRabbit

* **New Features**
  * Invitations can be bound to a specific email; acceptance verifies the accepting account's email (case-insensitive).
  * Link-style invitations without an email still work for any user.
  * Invitation payload and UI now expose an optional invitee email; sign-up pre-fills email from invitations.
  * New, reusable invitation info/actions UI components.

* **Bug Fixes**
  * Shows a dedicated, translated error when an email mismatch prevents acceptance.

* **Tests**
  * Added controller and end-to-end tests for match, mismatch, and link-invitation flows.


backend/api/src/main/kotlin/io/tolgee/hateoas/invitation/PublicInvitationModel.kt
```diff
@@ -9,4 +9,5 @@ class PublicInvitationModel(
   val createdBy: SimpleUserAccountModel?,
   val projectName: String?,
   val organizationName: String?,
+  val inviteeEmail: String?,
```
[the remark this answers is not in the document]
JanCizmar: It's not publicly available in that sense, you need to have the invitation url to read the e-mail right? So ite seems OK to me.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/V2InvitationControllerTest.kt
```diff
@@ -161,6 +163,96 @@ class V2InvitationControllerTest : AuthorizedControllerTest() {
     type.assert.equalsPermissionType(expectedType)
   }
 
+  @Test
+  fun `rejects project invitation when email does not match`() {
+    val base = dbPopulator.createBase()
+    val project = base.project
+    val code =
+      invitationService
```
JanCizmar: Maybe a method for creating invitation would save few lines of duplications...

webapp/src/component/security/AcceptInvitationView.tsx
```diff
@@ -193,32 +209,45 @@ const AcceptInvitationView: React.FC = () => {
...
-                flexWrap="wrap"
-                justifyContent="center"
-              >
-                <LoadingButton
-                  loading={acceptCode.isLoading}
-                  variant="contained"
-                  color="primary"
-                  onClick={handleAccept}
-                  data-cy="accept-invitation-accept"
+              {isEmailMismatch ? (
```
JanCizmar: The component is getting a bit long. I would suggest splitting it to multiple components, so the JSX is easier to read.

JanCizmar (commented):
Great job ❤️

JanCizmar (dismissed):

webapp/src/component/security/AcceptInvitationView.tsx
```diff
@@ -42,6 +44,97 @@ const StyledPaper = styled(Paper)`
   }
 `;
 
+type InvitationData = components['schemas']['PublicInvitationModel'];
+
+function InvitationInfoText({ data }: { data: InvitationData }) {
```
dkrizan: I would avoid having multiple components in same file

dkrizan (commented):
Should we prefill and lock the email field in the signup form when the invitation is bound to a specific address? Otherwise the user types it manually and only finds out about the mismatch after submitting.

dkrizan (commented):
Could we add e2e coverage for this?

bdshadow:
@dkrizan we can. I just thought that backend testing here is enough, it is the main point of rejection, but i will add an e2e test

bdshadow:
@dkrizan about the prefill and lock. Would you like me to it in this pr? The purpose was to fix the incorrect behavior so i'd not give here. Also, i suppose that often the user is already signed up to the platform when is invited to the org. So it might not be so necessary

webapp/src/component/security/AcceptInvitationView.tsx
bdshadow: done

dkrizan:
> @dkrizan about the prefill and lock. Would you like me to it in this pr? The purpose was to fix the incorrect behavior so i'd not give here. Also, i suppose that often the user is already signed up to the platform when is invited to the org. So it might not be so necessary

I'd prefer to include it in this PR. The email field is already being added here, so prefilling it is a small extra step, not separate work. 
About users already being signed up, I think the opposite. IMO, invitations are mostly how new people join an org, so the invitee is usually a first-time user, not an existing one. For new users, without prefill, they might sign up with the wrong one, hit the mismatch, and have to start over.

dkrizan (approved):

PR #3608, opened 2026-04-16
perf: use readOnly transactions for pure-read service methods

bdshadow:
Mark getter/find methods in Activity, Organization, Project, ProjectStats, Permission, LanguageStats, and Translation services as @Transactional(readOnly = true) to enable Hibernate read-only optimizations (no dirty-checking, read-replica routing).

Also drops readOnly=true from PromptVariablesHelper.getVariables, which transitively runs `SET local pg_trgm.similarity_threshold` via TranslationMemoryService.getSuggestionsList — PostgreSQL disallows SET in a read-only transaction.

## Summary by CodeRabbit

* **Chores**
  * Enhanced backend performance through targeted optimizations to transaction handling and database operations across core services including activity tracking, organization management, project operations, and security permissions.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/activity/ActivityService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/activity/ActivityService.kt
@@ -112,7 +112,7 @@ class ActivityService(
     }
   }
 
-  @Transactional
+  @Transactional(readOnly = true)
   fun findProjectActivity(
     projectId: Long,
     pageable: Pageable,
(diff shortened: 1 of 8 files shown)


dkrizan:
@bdshadow does this actually improve anything ? except `PromptVariablesHelper.getVariables` - that's a real bug

bdshadow:
@dkrizan yes, it does.

There are optimizations on the Hibernate level (skipping dirty-checking, entities are loaded differently to the persistent context) and postgres (different transaction ID assignment - check virtual XIDS).

These optimizations are small, but still make sense.

Also there is a little mistake in the pr descriptions - "PostgreSQL disallows SET in a read-only transaction.". Postgres actually allows it, it failed for me on jpa level when `executeUpdate` is called on entityManager in the read-only transaction

dkrizan (approved):

PR #3610, opened 2026-04-17
feat: deprecate embedded postgres ahead of v4

dkrizan:
## What

Two changes, both preparing for v4 dropping the bundled postgres from `tolgee/tolgee`:

1. **Deprecation warning** — the app logs a `WARN` on startup when it's actually running the bundled postgres, pointing at the external-database docs.
2. **Internal slim image** — `Dockerfile.slim` + the `dockerSlim` Gradle task. Not published. Built on demand by the EE image build and by the smoke test.

## Why

In v4, `tolgee/tolgee` loses the embedded postgres. Self-hosters running the bundled database have to move to an external one before upgrading.

They can already do that today on the current image by setting `postgres-autostart.enabled=false` (this is what the external-database docs section has always described). So the warning gives them a full release cycle of notice, on every boot, for a migration that needs nothing new from us.

The slim image is deliberately **not** published: a public `:slim` tag would only advertise a name that goes obsolete the moment v4 makes it the default, and it buys users nothing they can't already do. It exists so the internal EE image can drop the postgres layers now.

## Notes for review

- No behavior change for existing users beyond the log line. The default `Dockerfile` and the `tolgee/tolgee` tag are untouched.
- The warning is guarded on `enabled=true` **and** `mode=EMBEDDED`, so it stays silent for external-database setups and for Java + `DOCKER` autostart (that mode isn't going away).
- `docker-slim-smoke` boots the slim image against a postgres service and waits for `/actuator/health`.
- Docs: tolgee/documentation#1121 · EE image switch: tolgee/billing#274
- Revives #3193 (stale). Thanks @transparentChange for the original work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -504,7 +504,7 @@ jobs:
       contents: read
     services:
       postgres:
-        image: postgres:13
+        image: postgres:17
         env:
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: postgres
@@ -531,9 +531,6 @@ jobs:
         env:
           SKIP_SERVER_BUILD: true
 
-      # The slim image bundles no postgres server — must connect to the service
-      # container above. `--network host` puts the app on the runner's network
-      # so localhost:5432 resolves to the mapped postgres port.
       - name: Start slim image
         run: |
           docker run -d --name tolgee-slim --network host \
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -500,6 +500,8 @@ jobs:
     name: Slim Docker smoke 🐳
     needs: [backend-build]
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
     services:
       postgres:
         image: postgres:13
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -496,6 +496,69 @@ jobs:
         run: npm run eslint
         working-directory: ./email
 
+  docker-slim-smoke:
+    name: Slim Docker smoke 🐳
+    needs: [backend-build]
+    runs-on: ubuntu-24.04
+    services:
+      postgres:
+        image: postgres:13
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 5s
+          --health-timeout 3s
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -496,6 +496,71 @@ jobs:
...
+        uses: ./.github/actions/download-backend-build
+
+      - name: Build slim Docker image
+        run: ./gradlew dockerSlim
+        env:
+          SKIP_SERVER_BUILD: true
+
+      # The slim image bundles no postgres server — must connect to the service
+      # container above. `--network host` puts the app on the runner's network
+      # so localhost:5432 resolves to the mapped postgres port.
```
Anty0: Feels like kinda useless comment, but we can leave it.

docker/app/Dockerfile.slim
```diff
@@ -0,0 +1,47 @@
+FROM eclipse-temurin:21-jre-alpine
```
Anty0, line 1: I assume it isn't worth templating the Dockerfile? I assume this is mostly a copy of the other Dockerfile we already have. It's possible to create a template and set the base image via a variable, but I guess it's fine as is for now.

gradle/docker.gradle
```diff
@@ -33,6 +33,23 @@ tasks.register('docker') {
     dependsOn("dockerPrepare")
 }
 
+// Builds the slim Docker image for local development.
+// Single platform, no push, tagged as tolgee/tolgee:slim.
+// Uses Dockerfile.slim — JRE-only base, no embedded postgres.
```
Anty0: Useless comment - it mostly states the obvious and adds no value. Though the file seems to contain a lot of these...

Anty0 (dismissed):

docker/app/Dockerfile.slim
dkrizan: I was thinking the same. On the other hand, sometimes it’s better to duplicate the code rather than overcomplicate it by creating a template. I’ll give it some thought and see what I come up with. 
Thanks.

gradle/docker.gradle
dkrizan: Yea, it just followed the file pattern already.

docker/app/Dockerfile.slim
dkrizan: Since this will become the main (and only one) Dockerfile in v4, I would keep it as it is.

Anty0 (approved):

PR #3613, opened 2026-04-20
fix: handle TYPE_CUSTOM images in screenshot upload

### LINKED ISSUE #3260: Some screenshots (valid PNGs) can't be uploaded
**Describe the bug**
Uploading certain screenshots (.PNG format) results in a 500 server error.

**To Reproduce**
Try to upload attached .PNG screenshot.

**Expected behavior**
Valid .PNG screenshots should be uploaded and attached to a phrase.

**Additional context**
We've tried on multiple devices/browsers. Tolgee's support encountered the same issue with this file. We've been told to create this ticket.

Image

dkrizan:
## Summary
- `ImageIO.read` returns `BufferedImage.TYPE_CUSTOM` (0) for images with non-standard color models (16-bit PNG, CMYK, unusual palettes, etc.). Passing that type to `BufferedImage(w, h, type)` in `ImageConverter.getScaledImage` throws `IllegalArgumentException: Unknown image type 0`, failing the screenshot upload.
- Fix: fall back to `TYPE_INT_ARGB` / `TYPE_INT_RGB` (based on whether the source has an alpha channel) when the source is `TYPE_CUSTOM`; `Graphics2D.drawImage` handles the pixel conversion.

## Test plan
- [ ] Upload a 16-bit PNG as a key screenshot against current `main` — confirm it throws `IllegalArgumentException: Unknown image type 0`.
- [ ] Upload the same image with this branch applied — confirm the screenshot is stored and served correctly.
- [ ] Regression check: upload a regular RGB PNG and a PNG with transparency — confirm both still work and alpha is preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image resizing to more reliably preserve quality and format compatibility during scaling.
  * Better handling of images with transparency so scaled outputs keep correct transparency and color fidelity.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/util/ImageConverter.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/util/ImageConverter.kt
@@ -46,7 +46,7 @@ class ImageConverter(
   }
 
   private fun getScaledImage(targetDimension: Dimension): BufferedImage {
-    val resized = BufferedImage(targetDimension.width, targetDimension.height, safeImageType)
+    val resized = BufferedImage(targetDimension.width, targetDimension.height, getSafeImageType())
     val g = resized.createGraphics()
     g.setRenderingHint(
       RenderingHints.KEY_INTERPOLATION,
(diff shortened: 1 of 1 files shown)


JanCizmar (approved):

backend/data/src/main/kotlin/io/tolgee/util/ImageConverter.kt
```diff
@@ -68,6 +68,14 @@ class ImageConverter(
     return resized
   }
 
+  private val safeImageType: Int
+    get() =
+      if (sourceBufferedImage.type == BufferedImage.TYPE_CUSTOM) {
+        if (sourceBufferedImage.colorModel.hasAlpha()) BufferedImage.TYPE_INT_ARGB else BufferedImage.TYPE_INT_RGB
+      } else {
+        sourceBufferedImage.type
```
JanCizmar: Can we intruct Claude to not use elses? Instead to use functions with 2 return statement without the else branch? 🥸

JanCizmar (dismissed):

backend/data/src/main/kotlin/io/tolgee/util/ImageConverter.kt
dkrizan: Refactored to early returns in cc8d3073d.

PR #3621, opened 2026-04-23
test: fix flaky websocket "doesn't subscribe as another user"

JanCizmar:
## Summary

- **Resource leak cleanup**: `anotherUserWebsocket` (class-level) and the locally-created `spyingUserWebsocket` were never stopped. Each test leaked a STOMP session. Now `anotherUserWebsocket.stop()` runs in `@AfterEach` and `spyingUserWebsocket` is wrapped in `try/finally`. `WebsocketTestHelper.stop()` is fully idempotent (clears handler/connection refs, runs `webSocketStompClient.stop()` in `finally`).
- **Track full auth status history, fail loudly on lost frame**: the previous `handleTransportError` workaround silently relabeled any `ConnectionLostException` as `UNAUTHENTICATED`, masking genuinely lost FORBIDDEN rejections. New behavior:
  - Each frame/transport handler appends to an append-only `statusTransitions` list (CopyOnWriteArrayList). `CONNECTION_LOST` is recorded only when the transport closes with no status seen yet.
  - `waitForAuthenticationStatus` now checks `transitions.contains(expected)`. So:
    - **ERROR frame arrives in any order vs close** → list contains the expected status → clean pass, no spurious WARN even when `CONNECTION_LOST` was recorded first.
    - **ERROR frame is truly lost** → `transitions == [CONNECTION_LOST]` → wait times out and the test fails, with the full transitions list logged. The failure points investigators at the server-side flush-before-close path rather than at the test.
- **Diagnostics**: STOMP/transport errors at `ERROR` level with session id and the full transitions list, structured DEBUG logs for frame arrival and status transitions, and `DEBUG` enabled for `io.tolgee.websocket` + Spring STOMP packages in test `application.yaml`. Test failures now carry the rejection path in the HTML test report.

## Context

Tracked by the "Flaky tests" project board. The original failure (`billing/actions/runs/24710980170`, `WaitNotSatisfiedException at waitFor.kt:20` from `spyingUserWebsocket.waitForForbidden()`) and a related one we reproduced on `WebsocketAuthenticationTest.unauthenticated with invalid PAT` both come from the same root cause: Spring's `StompSubProtocolHandler` writes the rejection ERROR frame and closes the WebSocket; under CI load the bytes can be lost in the close race (verified — DEBUG logs from a failing CI run show 1 ms between "subscribing" and "transport error at close: null", with no "Frame received" entry in 5 s of polling).

If the test starts failing on `main` after this lands, that is the signal — the server-side flush-before-close path needs investigation. Frontend clients see the same race; a real product fix there would also stabilize the tests. Out of scope for this PR.

## Test plan

- [x] `WebsocketWithRedisTest`, `WebsocketWithoutRedisTest`, `WebsocketAuthenticationTest` — all 27 tests green locally
- [x] `@RepeatedTest(100)` on the originally flaky test — 100/100 green locally
- [x] 10 consecutive CI reruns on prior commits — our websocket tests passed every time. Failures during those runs were unrelated board-tracked flakes (`BatchJobsGeneralWithRedisTest`, ee-test init errors)

## Summary by CodeRabbit

* **Tests**
  * Improved websocket test reliability and lifecycle handling: tests now always stop created websockets, ensure cleanup on failures, and more robustly exercise forbidden/authentication flows.

* **Chores**
  * Enhanced logging and diagnostics for websocket/messaging: added debug connect/stop traces, stronger error-level reporting for transport/auth issues, clearer handling/reporting of connection-loss states, and reduced verbose payload logging.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/websocket/WebsocketTestHelper.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/websocket/WebsocketTestHelper.kt
@@ -16,6 +16,7 @@ import org.springframework.web.socket.messaging.WebSocketStompClient
 import org.springframework.web.socket.sockjs.client.SockJsClient
 import org.springframework.web.socket.sockjs.client.WebSocketTransport
 import java.lang.reflect.Type
+import java.util.UUID
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.LinkedBlockingDeque
 import java.util.concurrent.TimeUnit
@@ -53,8 +54,17 @@ class WebsocketTestHelper(
     logger.debug("Connecting websocket (userId={}, projectId={}, dest={})", userId, projectId, path)
     receivedMessages = LinkedBlockingDeque()
 
+    // Register the latch BEFORE the connect: the connect future's completion
+    // races against the test worker resuming from .get(), and ConcurrentHashMap
+    // visibility was not enough — we previously saw `awaitSubscribed` find the
+    // latch missing in the same millisecond `register` happened on the
+    // StompSession callback thread. Generating the correlationId here and
+    // registering synchronously guarantees a strict happens-before.
+    val correlationId = UUID.randomUUID().toString()
+    WebsocketTestSubscribeSync.register(correlationId)
+
     webSocketStompClient.messageConverter = SimpleMessageConverter()
-    sessionHandler = MySessionHandler(path, receivedMessages)
+    sessionHandler = MySessionHandler(path, receivedMessages, correlationId)
     connection =
       webSocketStompClient
         .connectAsync(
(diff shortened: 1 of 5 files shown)


backend/app/src/test/kotlin/io/tolgee/websocket/WebsocketTestHelper.kt
```diff
@@ -73,15 +75,22 @@ class WebsocketTestHelper(
...
+    val activeConnection = connection
+    sessionHandler = null
+    connection = null
+    logger.debug("Stopping websocket listener (sessionId={})", activeConnection?.sessionId)
     try {
-      sessionHandler?.subscription?.unsubscribe()
-      connection?.disconnect()
-    } catch (e: IllegalStateException) {
-      logger.warn("Could not unsubscribe from websocket", e)
+      try {
```
dkrizan: probably no need for try in try

backend/app/src/test/resources/application.yaml
```diff
@@ -105,3 +105,8 @@ tolgee:
 #logging:
```
dkrizan: maybe delete this commented logging

dkrizan (dismissed):

PR #3625, opened 2026-04-28
fix: prevent branch merge from dropping translation updates

dkrizan:
## Summary

Branch merges containing both a `DELETE` and an `UPDATE`/`CONFLICT` could silently drop the source-side translation update.

`BranchMergeExecutor.execute` iterated `merge.changes` in non-deterministic order, and `KeyService.hardDelete` calls `entityManager.clear()` (workaround for a Hibernate 6.6 `CHECK_ON_FLUSH` regression). Any in-memory modification made before that `clear()` was silently dropped on commit. Re-running the merge usually masked the bug.

Symptoms:
- Translation update not propagated to the target.
- `merge.mergedAt` stays `NULL` (merge keeps showing as active).
- Stale `branch_merge_change` rows remain (orphan-removal also dropped).

### Fix

- Sort changes so `DELETE`s run last — every other change is flushed by `hardDelete`'s own `flush()` before `clear()`.
- Re-fetch the merge after the loop so `mergedAt` and orphan-removal survive the cleared persistence context.
- New `BranchMerge.markMerged(at: Date)` to finalise the state.

### Out of scope

The root cause is `entityManager.clear()` in `KeyService.hardDelete`. A proper fix needs either a more surgical workaround for the Hibernate 6.6 stale-reference check, or running hard-deletes in `Propagation.REQUIRES_NEW`.

## Test plan

- [x] `BranchMergeServiceTest` — full class passes (was: 49/50 fail under `@RepeatedTest(50)`)
- [x] `io.tolgee.ee.service.branching.*` — all branching service tests pass
- [x] `BranchRevisionsTest` — confirms the `hardDelete` workaround is still load-bearing

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable branch merge completion: merges apply non-delete changes before deletions, then mark the merge as completed only after all changes are persisted to ensure accurate merged-state reporting.
* **Tests**
  * Added verification that applied merges record a completion timestamp and clear pending changes in storage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3625?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/model/branching/BranchMerge.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/branching/BranchMerge.kt
@@ -63,4 +63,9 @@ class BranchMerge :
 
   val isMerged: Boolean
     get() = mergedAt != null
+
+  fun markMerged(at: Date) {
+    changes.clear()
+    mergedAt = at
+  }
 }
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/branching/BranchMergeService.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/branching/BranchMergeService.kt
@@ -98,16 +98,14 @@ class BranchMergeService(
     val targetKeyId: Long?,
   )
 
-  fun applyMerge(merge: BranchMerge) {
-    metrics.branchMergeApplyTimer.record(
-      Runnable {
-        try {
-          branchMergeExecutor.execute(merge)
-        } catch (_: BranchMergeConflictNotResolvedException) {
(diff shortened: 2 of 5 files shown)


JanCizmar (changes requested):
"fix: flaky tests" does not make sense. The "fix: " prefix, will release a patch version. Patch version release means that we fixed a bug or changed the behavior somehow. If we changed the behavior, it should have proper name so it's displayed correctly in changelog. If not, then it should not have the "fix: " prefix.

Anty0 (dismissed):
Might be worth renaming the PR, since it fixes a potentially real issue rather than just the test.

Anty0 (approved):

PR #3630, opened 2026-04-30
fix: open Context Data tab by default on the AI page

Anty0:
## Summary

- Reorder the project AI page tabs so **Context Data** is shown first and opens by default when the user navigates to `/projects/:id/ai`.
- Add `LINKS.PROJECT_AI_PROMPTS = /ai/prompts` so the Prompts tab gets its own URL instead of squatting on the parent `/ai` route. Mirrors the existing `/developer/*` sub-route pattern.
- Update the project menu's AI entry to link directly to `LINKS.PROJECT_CONTEXT_DATA` and use the parent path for `matchAsPrefix`, so the entry stays highlighted across `/ai`, `/ai/prompts`, and `/ai/context-data`.
- Add an exact `<Redirect>` from `/ai` → `/ai/context-data` in `ProjectRouter` for users who type or bookmark the bare URL.
- Refactor `useAiViewItems` to fix a pre-existing rules-of-hooks brittleness from calling the hook inside `.map()`.
- Update e2e helpers (`visitAiSettings` and the AI permissions test) so prompt-focused suites still land on the Prompts UI.
- Drive-by cleanups: rename the long-standing typo `requireExactMath` → `requireExactMatch`, tighten the `Link` constructor JSDoc, simplify `Link.build` with a default parameter, and dedupe `projectIdParams` in `ProjectMenu`.

## Summary by CodeRabbit

* **Bug Fixes**
  * AI section now redirects to the correct default AI sub-view and menu links consistently open the intended sub-section.

* **New Features**
  * Added a dedicated Prompts sub-tab within the AI area and improved tab selection behavior so the active tab is highlighted reliably.

* **Tests**
  * Enhanced end-to-end AI tests to reliably navigate to AI Prompts and verify prompt visibility, edit/delete, and playground actions.

### DIFF
--- a/e2e/cypress/common/prompt.ts
+++ b/e2e/cypress/common/prompt.ts
@@ -1,6 +1,7 @@
 import { gcy, gcyAdvanced } from './shared';
 import { getTranslationCell } from './translations';
 import { components } from '../../../webapp/src/service/apiSchema.generated';
+import { LINKS, PARAMS } from '../../../webapp/src/constants/links';
 import { buildXpath } from './XpathBuilder';
 import { HOST } from './constants';
 
@@ -42,5 +43,9 @@ export function getPromptEditor() {
 }
 
 export function visitAiSettings(projectId: number) {
-  return cy.visit(`${HOST}/projects/${projectId}/ai`);
+  return cy.visit(
+    `${HOST}${LINKS.PROJECT_AI_PROMPTS.build({
+      [PARAMS.PROJECT_ID]: projectId,
+    })}`
+  );
 }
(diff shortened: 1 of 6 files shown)


e2e/cypress/common/prompt.ts
```diff
@@ -42,5 +43,9 @@ export function getPromptEditor() {
 }
 
 export function visitAiSettings(projectId: number) {
-  return cy.visit(`${HOST}/projects/${projectId}/ai`);
+  return cy.visit(
+    `${HOST}${LINKS.PROJECT_AI_PROMPTS.build({
```
dkrizan: I think the URL should be typed as it is, not using the reference on LINKS.
Tests should behave like a user, not like your code.

dkrizan (approved):

PR #3638, opened 2026-05-06
fix: refresh language stats after QA batch job rerun

Anty0:
## Summary

- `LanguageStatsListener.onQaBatchJobFinalized` resolved affected branches from `activity_modified_entity`, but the QA chunk processor sets `disableActivityLogging = true` on every translation/issue it persists (and uses a JPQL bulk delete for replaced issues), so that table is always empty for QA jobs. The listener returned early and never refreshed `language_stats` — visible to users as "Rerun QA Checks" leaving `qaChecksStaleCount` pinned at the pre-rerun value.
- Resolve affected branches from the batch job's target key IDs via `keyRepository.getBranchIdsByIds` (matching the `onActivity` path).
- Drop the now-unused `findDistinctBranchIdsByRevisionId` helper from `ActivityService` and the matching repository query.
- Rewrite `QaLanguageStatsRefreshTest` to drive the listener with realistic `BatchTranslationTargetItem` payloads — the previous test stubbed `activity_modified_entity` rows and so masked the bug.

## Summary by CodeRabbit

* **Refactor**
  * Language-stat refresh now derives affected branches directly from QA job targets, ensuring consistent, branch-scoped updates after QA workflows.

* **Tests**
  * Updated and added tests to validate language-stat refresh for main and multi-branch scenarios using target-aware QA batch-finalization events and improved test helpers for setup and assertions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3638)


backend/data/src/main/kotlin/io/tolgee/component/eventListeners/LanguageStatsListener.kt
```diff
@@ -69,15 +62,44 @@ class LanguageStatsListener(
...
+
+    branchIds.filterNotNull().forEach { branchId ->
+      languageStatsService.refreshLanguageStats(projectId, branchId)
+    }
+  }
+
+  private fun extractKeyIdsFromQaJobTarget(target: List<Any>): List<Long> {
+    if (target.isEmpty()) return emptyList()
+    val type =
+      jacksonObjectMapper().typeFactory.constructCollectionType(
```
bdshadow: ` jacksonObjectMapper()` is constructed twice here (it's actually pretty expensive). `WebConfiguration.objectMapper()` actually declared a bean which can injected and used instead

backend/data/src/main/kotlin/io/tolgee/component/eventListeners/LanguageStatsListener.kt
```diff
@@ -69,15 +62,44 @@ class LanguageStatsListener(
...
+
+      refreshLanguageStatsForBranches(projectId, branchIds)
     }
   }
 
+  private fun refreshLanguageStatsForBranches(
+    projectId: Long,
+    branchIds: Set<Long?>,
+  ) {
+    if (branchIds.contains(null)) {
```
bdshadow: this check is not needed, actually. `refreshLanguageStats` method has a default null value for the `branchId` parameter. So the whole body of the mothod can be the following:

```kt
branchIds.forEach { branchId ->
    languageStatsService.refreshLanguageStats(projectId, branchId)
 }
```

bdshadow (changes requested):

backend/data/src/main/kotlin/io/tolgee/component/eventListeners/LanguageStatsListener.kt
Anty0: Thanks! It was there like this already, I didn't check what's inside the `refreshLanguageStats`, so I didn't realize :D

backend/data/src/main/kotlin/io/tolgee/component/eventListeners/LanguageStatsListener.kt
Anty0: Hmm, `ChunkProcessingUtil` does this too. Might be worth reusing an object mapper instance there as well.

```
  private fun convertChunkToItsType(chunk: List<Any>): List<Any> {
    val type =
      jacksonObjectMapper().typeFactory.constructCollectionType(List::class.java, processor.getTargetItemType())
    return jacksonObjectMapper().convertValue(chunk, type) as List<Any>
  }
```

I'll fix it in the `LanguageStatsListener`, and I'll leave it as it is in the `ChunkProcessingUtil` for now - not really related to this PR.

bdshadow (approved):

PR #3641, opened 2026-05-06
fix: avoid Repeated Words QA check false positives

Anty0:
## Summary

- The Repeated Words QA check now requires the separator between two consecutive word tokens to consist only of regular spaces (U+0020) and/or non-breaking spaces (U+00A0). Tags, placeholders, URLs, punctuation, and line breaks no longer act as valid separators, eliminating the most common false positives (e.g. adjacent `<br><br>` reported as a repeated "br").
- The same blocked-range filter that Spelling/Grammar already use is now applied to Repeated Words results too, so cases like word-like tokens inside HTML attribute values are dropped.
- The blocked-range filter was generic but lived in the `language` subpackage. Moved to the parent `qa.checks` package, renamed to `filterResultsInBlockedRanges`, and split the unrelated glossary-term filter into its own file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced false positives by better ignoring HTML, URLs, placeholders and other blocked ranges during QA checks
  * Improved repeated-word detection to only flag true repeats separated by plain spaces

* **New Features**
  * Glossary-aware filtering: QA results overlapping glossary terms are excluded

* **Tests**
  * Expanded and reorganized tests covering whitespace, HTML, placeholders, URLs and glossary-related scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3641)


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/RepeatedWordsCheck.kt
```diff
@@ -29,24 +29,43 @@ class RepeatedWordsCheck : QaCheck {
...
+    for (i in start until end) {
+      val c = text[i]
+      if (c != SPACE && c != NBSP) return false
+    }
+    return true
   }
 
   companion object {
     private val WORD_REGEX = Regex("""\p{L}[\p{L}\p{N}]*""")
+    private const val SPACE = ' '
```
bdshadow: These are visually identical in source code.
```
  private const val SPACE = '\u0020'
  private const val NBSP = '\u00A0'
```
would be better for reading it

bdshadow (changes requested):

bdshadow (approved):

PR #3629, opened 2026-04-29
fix: performance of import

bdshadow:
## Summary by CodeRabbit

* **New Features**
  * Real-time import progress (imported/total counts) shown in UI and progress bar.
  * End-to-end performance scripts for single- and multi-step imports.

* **Bug Fixes / Performance**
  * Large-import refactor: batched persistence, reduced memory use, manual activity recording, faster throughput.
  * Improved caching and batching for import/update workflows; streaming progress updates.

* **Tests**
  * Added (disabled) import performance test suites.

* **Documentation**
  * Dev notes on import performance refactor and benchmarks.

* **Chores**
  * .gitignore updated to exclude generated logs/JSON.

PR #3644, opened 2026-05-07
perf: single-step import memory + activity-recording follow-ups

JanCizmar:
Follow-ups stacked on top of #3629 to push single-step-import perf and memory further. The branch now sits cleanly as four reviewable commits on top of #3629's HEAD.

### 1. `revert: roll back manual JDBC activity recording`

Drops the bespoke `ImportActivityRecorder` and the JDBC-level snippet in `InterceptedEventsManager` that bypassed Hibernate's activity interceptor. With the listener-side memory work below in place, the natural Hibernate path holds at scale, and rolling this back removes a chunk of import-specific bookkeeping from the activity pipeline.

### 2. `perf: hotspot fixes outside the import flow`

Drive-by hotspots that profiled hot during large imports but live outside the data importer itself:

- **WordCounter**: stopped tokenising the full string just to count words; now samples a bounded prefix for the common case.
- **TranslationService.setQaChecksStale**: changed the bulk update to take a single `bigint[]` array (`= ANY(:ids)`) instead of expanding to thousands of `?` placeholders, and added `AND qa_checks_stale = false` so PostgreSQL skips writes on fresh rows (no heap rewrite, no index churn, no WAL).
- **ProjectHolderConfig**: removed the prototype-proxy indirection on the project holder bean.
- **FormatDetectionUtil**: caps detection sampling and drops the `AtomicLong` counter that was a contention hotspot at parallel-file scale.
- **QaActivityListener**: pre-filters changed entities to base-language ids in Kotlin (via `describingRelations`) so the inner SQL sees only relevant ids.

### 3. `perf: shrink activity-recording memory footprint`

The activity interceptor was the single biggest heap consumer once the import itself was batched. This commit cuts that:

- New `PropertyModifications` map — a wire-compatible replacement for `Map<String, PropertyModification>` that shares field-name strings via a per-entity-class `NameSchema` registry instead of allocating a fresh `String` per recorded modification. Same JSONB shape on disk; new unit tests in `PropertyModificationsTest` verify round-trip equivalence with the legacy `Map` form.
- New `CompactSharedMap`, `DescribingDataMap`, `DescribingRelationsMap`, `ModifiedCollectionKey`, `IntArrayList` — compact data structures backing the `ActivityModifiedEntity` JSONB columns without the per-row map/list overhead.
- `@Immutable` on `ActivityModifiedEntity` so Hibernate skips dirty-checking and per-instance field-tracker allocation. The entity is insert-only, the annotation matches the actual semantics.
- `EntityDescriptionProvider` / `InterceptedEventsManager` cleanups to feed the new map types and reuse them across the listener pipeline.

### 4. `perf: tighten single-step import flow on top of batched persist`

Built on top of #3629's batched persist. Targets the residual memory pressure between batches:

- **Per-language data release.** `ImportDataManager.releaseLanguageData` drops one language's parsed `ImportTranslation` graph as soon as the data importer has extracted from it, so the peak holds at most one language alongside the partially-built keys/translations.
- **Full parsed-data release.** `ImportDataManager.releaseData` is called once all keys and translations have been moved into the importer's working structures. The `~2M` `ImportKey` + `ImportTranslation` graph stops pinning memory through the persist phase.
- **Per-batch tagging.** `tagNewKeys` moves into `flushBatch` and identifies the "new" subset of each batch via `id == 0L` captured before save. The `newKeys` list is cleared right after `addScreenshots` so the persist phase no longer pins 1M `Key` references in `ActivityHolder`.
- **`pluralFlipKeys` instead of pinning ImportTranslation.** `translationsToSave` is now a plain `List<Translation>` — the only two scalars `PluralizationHandler` actually needed from the source are captured in a sparse `pluralFlipKeys` map keyed by (namespace, name). At zero plural keys this stays empty.
- **Nested maps instead of `Pair` keys.** `keysToSave` and `ImportDataManager.storedKeys` are nested `HashMap`s so saved-key lookups don't allocate one `Pair` per key.
- **ImportKey: drop inverse `translations` @OneToMany.** Replaced with a single `findByKeyId` derived query on `ImportTranslationRepository` where actually needed; eliminates a hibernate bag wrapper per ImportKey.
- **Tighter ArrayList sizing.** `populateStoredTranslations` and `getStoredTranslations` allocate initial-capacity-1 lists for the common 1:1 per-(lang, key) case.
- **Configurable batch size.** `ImportProperties.flushBatchSize` (default 10,000) — lower this in environments that OOM at scale.

### Test coverage

- New `SingleStepImportActivityTest` — end-to-end coverage of activity recording across batches (creates, updates, deletes, plurals, tagging, namespace creation). Verifies the natural Hibernate interceptor path produces the same `ActivityModifiedEntity` rows we'd expect on a non-batched import, including a many-key (25-key) case that exercises per-batch flushing.
- `SingleStepImportPerformanceTest` (`@Disabled`) — verifies via `keyRepository.countByProjectId` rather than loading every Key into memory (the verification step was the second OOM, after the import itself).
- Existing `V2ImportControllerAddFilesTest`, `SingleStepImportControllerTest`, `V2ImportControllerApplicationTest`, `V2ImportControllerPluralizationTest`, and the rest of the `v2ImportController` suite all stay green.

### `CleanDbTestListener` test-isolation fix

Test-isolation bug surfaced while bringing the new SingleStepImportActivityTest online: `CleanDbTestListener` was issuing a session-scoped `SET statement_timeout = 5000` on a Hikari-pooled connection. Without `DISCARD ALL` / `RESET` on connection return, the 5s timeout leaked into application code on subsequent borrowings and made unrelated bulk operations abort. Switched to `SET LOCAL` so it's confined to the truncate transaction.

## Test plan

- [x] `:server-app:test --tests SingleStepImportActivityTest` (9 tests) — green
- [x] `:server-app:test --tests V2ImportControllerAddFilesTest` — green
- [x] `:server-app:test --tests SingleStepImportControllerTest` — green
- [x] `:server-app:test --tests V2ImportControllerApplicationTest` — green
- [x] `:server-app:test --tests V2ImportControllerPluralizationTest` — green
- [ ] CI pipeline on the platform repo


backend/data/src/main/kotlin/io/tolgee/util/WordCounter.kt
```diff
@@ -7,12 +7,20 @@ object WordCounter {
   private val NON_WORD = """[\p{P} \t\n\r~!@#$%^&*()_+{}\[\]:;,.<>/?-]""".toRegex()
   private val LANGUAGE_PART = "^([A-Za-z]+).*".toRegex()
 
+  // BreakIterator construction is expensive (~ms per call) and the result is not
+  // thread-safe. Keep one instance per thread per language tag — bulk imports
+  // call countWords millions of times with a small set of locales.
+  private val iteratorCache: ThreadLocal<MutableMap<String, BreakIterator>> =
+    ThreadLocal.withInitial { mutableMapOf() }
```
bdshadow, line 14: `ThreadLocal` is considered bad in modern java, especially with virtual threads. You can google about it a lot of articles. Same problem can be with kotlin coroutines. If countWords is ever called from a coroutine that suspends and resumes on a different thread, the `ThreadLocal` cache would miss (harmless but defeats the purpose) or worse, a `BreakIterator` could be shared between two coroutines on the same thread concurrently

JanCizmar: Feel free to fix it. :)

PR #3653, opened 2026-05-11
fix: preserve ICU exact-match plural selectors (=0, =1) in export

### LINKED ISSUE #3267: Translation with plurals corrupted/changed after import

peterzen:
Two related issues caused push→pull round-trips to produce file diffs:

1. IcuToGenericFormatMessageConvertor.convert() called toIcuPluralString() without optimize=false, relying on the default optimize=true. This triggered optimizePluralForms + orderPluralForms on every export, silently reordering or removing plural forms that the user authored.

2. orderPluralForms sorted unknown keys (including =0, =1, =2 exact-match selectors) alphabetically after 'other', placing them last instead of first. ICU exact-match selectors should precede keyword selectors and appear in ascending numeric order.

Fix: pass optimize=false in the export path so the stored ICU string is emitted verbatim. Also update orderPluralForms to sort =N selectors before keyword forms in numeric order so that explicit optimize calls produce canonical output consistent with the ICU specification.

Fixes: https://github.com/tolgee/tolgee-platform/issues/3267

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plural form ordering to correctly prioritize exact-match selectors (e.g., `=0`, `=1`) before standard keywords, ensuring they appear in ascending numeric order for more predictable and accurate localization handling.
  * Enhanced format conversion to preserve exact-match selectors during export and round-trip operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3653)

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/generic/IcuToGenericFormatMessageConvertor.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/generic/IcuToGenericFormatMessageConvertor.kt
@@ -30,6 +30,7 @@ class IcuToGenericFormatMessageConvertor(
     val formsResult = converted.formsResult ?: return ""
     return formsResult
       .toIcuPluralString(
+        optimize = false,
         addNewLines = false,
         argName = converted.argName ?: DEFAULT_PLURAL_ARGUMENT_NAME,
       )
--- a/backend/data/src/main/kotlin/io/tolgee/formats/pluralFormsUtil.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/pluralFormsUtil.kt
@@ -41,14 +41,17 @@ fun populateForms(
 
 fun orderPluralForms(pluralForms: Map<String, String>): Map<String, String> {
   return pluralForms.entries
-    .sortedBy {
-      val formIndex = formKeywords.indexOf(it.key)
-      if (formIndex == -1) {
-        "A_$it"
-      } else {
-        formIndex.toString()
-      }
-    }.associate { it.key to it.value }
+    .sortedWith(
+      compareBy(
+        // Exact-match selectors (=0, =1, =2, …) sort first, in ascending numeric order
+        { if (it.key.startsWith("=")) it.key.substring(1).toLongOrNull() ?: Long.MAX_VALUE else Long.MAX_VALUE },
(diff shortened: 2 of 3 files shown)


Anty0:
Duplicate of #3548

PR #3657, opened 2026-05-12
fix: add tooltip and responsive wrapping for long language names (#3611)

### LINKED ISSUE #3611: Long language names cut off in the language selector dropdown
**Describe the bug**
In the language selector dropdown longer language names are truncated and cut off on the right side of the dropdown. This makes it impossible to read the full name of languages with longer labels, resulting in a poor user experience.

**To Reproduce**
Go to a project's Translations view
Click on the language selector dropdown 
See that languages with longer names (including region/country qualifiers) are visually truncated and cannot be read in full

**Expected behavior**
The dropdown should display the full language name for each option. The user should always be able to clearly read which language they are selecting.

Browser: Safari

Image

Rkarama26:
- Add tooltip on hover for full language name
- Expand dropdown width (minWidth: 200px, maxWidth: 350px)
- Use ellipsis on large screens, wrapping on small screens

Screenshot (284)

browser-chrome, brave

Fixes #3611

## Summary by CodeRabbit

* **Style**
  * Improved language dropdown sizing with responsive minimum and maximum widths.
  * Added full-name tooltips for language options when labels are truncated.
  * Enhanced label wrapping and truncation behavior for clearer display across screen sizes.

### DIFF
--- a/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
+++ b/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
@@ -52,7 +52,8 @@ export const LanguagesSelect: FunctionComponent<Props> = (props) => {
     variant: 'menu',
     PaperProps: {
       style: {
-        width: 250,
+        minWidth: 200,
+        maxWidth: 350,
       },
     },
     id: `language-select-${props.context}-menu`,
--- a/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
+++ b/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
@@ -51,10 +51,10 @@ export const LanguagesSelect: FunctionComponent<Props> = (props) => {
   const menuProps: Partial<MenuProps> = {
     variant: 'menu',
     PaperProps: {
-      style: {
+      sx: (theme) => ({
         minWidth: 200,
-        maxWidth: 350,
-      },
+        maxWidth: 300,
+      }),
     },
     id: `language-select-${props.context}-menu`,
     anchorOrigin: {
--- a/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
+++ b/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
@@ -51,10 +51,9 @@ export const LanguagesSelect: FunctionComponent<Props> = (props) => {
   const menuProps: Partial<MenuProps> = {
     variant: 'menu',
     PaperProps: {
-      sx: (theme) => ({
-        minWidth: 200,
-        maxWidth: 300,
-      }),
+      sx: {
+        width:250,
+      }
     },
     id: `language-select-${props.context}-menu`,
     anchorOrigin: {
--- a/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
+++ b/webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
@@ -52,8 +52,9 @@ export const LanguagesSelect: FunctionComponent<Props> = (props) => {
     variant: 'menu',
     PaperProps: {
       sx: {
-        width:250,
-      }
+        minWidth: 200,
+        maxWidth: 250,
+      },
     },
     id: `language-select-${props.context}-menu`,
     anchorOrigin: {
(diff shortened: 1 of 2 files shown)


Rkarama26:
As seen above some backend tests are failing, 
But i have not made any changes in backend code i worked only in frontend code @bdshadow

webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
```diff
@@ -51,9 +51,10 @@ export const LanguagesSelect: FunctionComponent<Props> = (props) => {
   const menuProps: Partial<MenuProps> = {
     variant: 'menu',
     PaperProps: {
-      style: {
-        width: 250,
-      },
+      sx: (theme) => ({
```
bdshadow: a `theme` variable is unused here, so `sx` can be a plain object `sx: { minWidth: 200,
  maxWidth: 300 }`

webapp/src/component/common/form/LanguagesSelect/getLanguagesContent.tsx
```diff
@@ -67,16 +73,33 @@ export const getLanguagesContent = ({
...
-      value={lang.tag}
-      data-cy="translations-language-select-item"
-      onClick={handleLanguageChange(lang.tag)}
-      disabled={disabledLanguages?.includes(lang.id)}
-    >
-      <Checkbox checked={value?.includes(lang.tag)} size="small" />
-      <ListItemText primary={lang.name} />
-    </MenuItem>
+    <Tooltip key={lang.tag} title={lang.name} placement="right">
+      <span style={{ display: 'inline-block', width: '100%' }}>
```
bdshadow: Keyboard arrow navigation and selected highlighting is broken here. Please, try to open the dropdown, use arrow keys, press Enter to select language).

Probably keeping `<MenuItem>` as the direct child of the Select and rendering the `<Tooltip>` around the inner content (`<ListItemText>`) will solve it (i haven't checked it myself though)

bdshadow (changes requested):
@Rkarama26 thank you for the pull request

Failing backend tests - are just flaky; we're working on improving them. 

However, there are some other real concerns about the changes. Please check the individual code comments.

Also, it would be great to show the tooltip only when the language name doesn't fit the width, and not show it for short names. Probably you can achieve this by deferring the measurement to `onMouseEnter` - compute `scrollWidth > clientWidth` lazily and toggle the tooltip open.

webapp/src/component/common/form/LanguagesSelect/LanguagesSelect.tsx
bdshadow: @Rkarama26 
`PaperProps` is a plain object now, but the value went back to width: 250- exactly what main already has via style. So this file's diff is a no-op, and the "Expand dropdown width (minWidth 200 / maxWidth 350)" claim in the PR description is no longer true. Should either be reverted or actually set minWidth/maxWidth

webapp/src/component/common/form/LanguagesSelect/getLanguagesContent.tsx
bdshadow: @Rkarama26 
The `<Tooltip>` was moved inside, around `<ListItemText>` (good), but the `<span>` wrapper is still the root of the `.map()`, so `MenuItem` is still not a direct child of the `Select`.

- `MenuList.moveFocus` walks the list's DOM element siblings and skips anything without a tabindex attribute. The direct children are now `<span>`s with no tabindex, so arrow keys find nothing focusable.
- `SelectInput` clones each direct child with selected. The `<span>` has no value, so selected is always false, lands on a plain `<span>` instead of the `MenuItem`, so no `Mui-selected` highlight, and `role="option"/data-valu`e are wrong.
- Side note: it also produces `<ul>` -> `<span>` -> `<li>`, which is invalid markup.

Removing the `<span>` and putting `key={lang.tag}` back on MenuItem is all that's needed now that the Tooltip is on the inner content.

bdshadow:
@Rkarama26 i'm very sorry for such a delay in reply. It was a busy summer.

Please check my comments above. 

Also, the requirement in my comment above is still not addressed: 
> show the tooltip only when the name doesn't fit; measure `scrollWidth` > `clientWidth` on `onMouseEnter`

And one more thing, please make sure that "Frontend static check" is not failing. Run ` npm run eslint -- --fix` from the `webapp` directory

Rkarama26:
Ohk I'm looking into it

bdshadow (approved):

bdshadow:
@Rkarama26 thank you for the contribution!

PR #3664, opened 2026-05-13
fix: handle ICU plural # in QA checks

Anty0:
## Summary

- **Inconsistent placeholders** check now treats the ICU plural `#` (REPLACE_NUMBER) as a placeholder inside plural variants. Missing or extra `#` between base and translation is reported, with position + replacement on the extra path so editors can offer to delete it. Escaped `'#'` is correctly ignored (the variant body is wrapped in a synthetic plural so the ICU parser handles quoting), and `#` nested inside a select-inside-plural stays a literal — matching ICU's actual semantics.
- **Special character mismatch** check now skips `#` for projects with ICU placeholder support enabled (mirrors how brackets checks already use `params.icuPlaceholders`). This avoids double-reporting once Inconsistent placeholders owns `#`.
- **`QaPluralCheckHelper.runPerVariant`** signals whether each invocation is inside a plural variant body, so the extractor knows when to switch on REPLACE_NUMBER detection. All other QA checks pick up the new lambda parameter.
- Tooltip for "Special character mismatch" updated in Tolgee Cloud (branch `fix-qa-icu-hash-handling`) to explain the new ICU-aware behavior.

### Tradeoff worth flagging

When ICU is enabled, `#` is dropped from the Special character mismatch check *everywhere* — even outside any plural (e.g. literal `#hashtag` in regular text). Simpler than locating `#` only inside plurals and avoids any risk of double-reporting.

## Release Notes

* **New Features**
  * Enhanced QA checks to better validate plural form consistency, including improved detection of missing or extra placeholders in plural variants.
  * Improved ICU plural placeholder support with more accurate argument tracking across variant forms.

* **Tests**
  * Expanded test coverage for plural form validation and special character detection in QA checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3664)

### DIFF
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/InconsistentPlaceholdersCheck.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/InconsistentPlaceholdersCheck.kt
@@ -13,8 +13,8 @@ class InconsistentPlaceholdersCheck : QaCheck {
   override val type: QaCheckType = QaCheckType.INCONSISTENT_PLACEHOLDERS
 
   override fun check(params: QaCheckParams): List<QaCheckResult> {
-    return QaPluralCheckHelper.runPerVariant(params) { text, baseText ->
-      checkVariant(text, baseText)
+    return QaPluralCheckHelper.runPerVariant(params) { text, baseText, isVariant ->
+      checkVariant(text, baseText, isVariant)
     }
   }
 
@@ -31,13 +31,13 @@ class InconsistentPlaceholdersCheck : QaCheck {
   private fun checkVariant(
     text: String,
     baseText: String?,
+    isPluralVariant: Boolean,
   ): List<QaCheckResult> {
     val base = baseText ?: return emptyList()
-    if (base.isBlank()) return emptyList()
-    if (text.isBlank()) return emptyList()
+    if (base.isBlank() || text.isBlank()) return emptyList()
 
-    val baseArgs = deduplicateNestedArgs(extractArgs(base) ?: return emptyList())
-    val textArgs = deduplicateNestedArgs(extractArgs(text) ?: return emptyList())
+    val baseArgs = deduplicateNestedArgs(extractArgs(base, isPluralVariant) ?: return emptyList())
+    val textArgs = deduplicateNestedArgs(extractArgs(text, isPluralVariant) ?: return emptyList())
 
     val baseCounts = baseArgs.groupingBy { it.name }.eachCount()
     val textCounts = textArgs.groupingBy { it.name }.eachCount()
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/PlaceholderExtractor.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/PlaceholderExtractor.kt
@@ -8,11 +8,28 @@ data class ArgInfo(
   val positionEnd: Int? = null,
 )
 
-fun extractArgs(text: String): List<ArgInfo>? {
+/**
+ * Virtual placeholder name used for the ICU plural `#` token (REPLACE_NUMBER).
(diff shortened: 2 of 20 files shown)


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/InconsistentPlaceholdersCheck.kt
```diff
@@ -31,13 +31,13 @@ class InconsistentPlaceholdersCheck : QaCheck {
...
+    isPluralVariant: Boolean,
   ): List<QaCheckResult> {
     val base = baseText ?: return emptyList()
-    if (base.isBlank()) return emptyList()
-    if (text.isBlank()) return emptyList()
+    if (base.isBlank() || text.isBlank()) return emptyList()
 
-    val baseArgs = deduplicateNestedArgs(extractArgs(base) ?: return emptyList())
-    val textArgs = deduplicateNestedArgs(extractArgs(text) ?: return emptyList())
+    val baseArgs = deduplicateNestedArgs(extractArgs(base, isPluralVariant) ?: return emptyList())
```
bdshadow, line 39: Since we're solving performanace of the QA checks, it seems that we can easily avoid double parsing by `MessagePatternUtil.buildMessageNode`.  The follwoing can be added to the `QaCheckParams`:
```kt
  val textIcuParseResult: Result<MessagePatternUtil.MessageNode> by lazy {
    runCatching { MessagePatternUtil.buildMessageNode(text) }
  }
  val baseTextIcuParseResult: Result<MessagePatternUtil.MessageNode>? by lazy {
    baseText?.let { runCatching { MessagePatternUtil.buildMessageNode(it) } }
  }
```
then call it from `IcuSyntaxCheck` and `InconsistentPlaceholdersCheck`. Please, check the patch, if it makes sense for you

[qa-perf-ast-cache.patch](https://github.com/user-attachments/files/28011465/qa-perf-ast-cache.patch)

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/PlaceholderExtractor.kt
```diff
@@ -21,11 +38,29 @@ fun extractArgs(text: String): List<ArgInfo>? {
   }
 }
 
+private fun buildMessageNode(
+  text: String,
+  isPluralVariant: Boolean,
+): MessagePatternUtil.MessageNode? {
+  if (!isPluralVariant) {
```
bdshadow, line 45: i think, here the condition can be `(!isPluralVariant || '#' !in text)` to skip when it's possible. I also added it to the patch file from the comment above. Maybe i'm wrong here, please check

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/InconsistentPlaceholdersCheck.kt
Anty0: I remember thinking about something like this. Since we already parse the ICU to extract plurals, I was looking at reusing that, but it isn't really possible without refactoring, so I decided against it. Doing another ICU parsing when building `QaCheckParams` doesn't really make sense, since it would be useless if the user disables ICU-related QA checks.

Currently, something like >90% of the time spent on QA Checks comes from DB, Hibernate and other data processing steps. Actual checks (except LT checks) run very fast - it's probably not worth optimizing this for now.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/checks/PlaceholderExtractor.kt
Anty0: Hmm, I'm not completely sure if there aren't more differences between the text inside and outside of the plural. It feels safer the way it is now, and I wouldn't expect any significant speed improvement from this.

bdshadow (approved):

PR #3667, opened 2026-05-14
feat: rework QA filter to support per-type + per-language scoping

Anty0:
## Summary

- `filterQaCheckType` query param now accepts `languageTag,checkType` pairs (mirrors `filterState`). The frontend builds one entry per `(selected language) × (selected check type)`; per-language predicates are OR-ed. Breaking change to the param shape — the param was new and unreleased.
- The Translations view QA submenu adds a language-scope radio (default **All languages**, with **All but base** and per-language options behind a chevron). "Any QA issue" becomes a UI shortcut that ticks every known check type.
- QA check categories are now embedded in `/v2/public/initial-data` (static enum metadata). `QaCheckCategoryModel` + `QaCheckCategoryModelAssembler` moved from `:ee-app` to `:api`. Frontend gets new `useQaCategories()` and `useQaCheckTypes()` helpers reading from the global initial-data context — no per-project fetch in the QA submenu, and external entry-points (QA dashboard, project list QA badge) can build the "all check types" deep-link synchronously.
- New backend `TranslationFilterByQaCheckType` parser modeled on `TranslationFilterByState` (uses `BaseFilterByKeyValue.parseList`). New `Message.FILTER_BY_VALUE_QA_CHECK_TYPE_NOT_VALID` for malformed values.
- `applyQaFilters` rewritten to mirror `applyLabelFilter`: parse → group-by-language → emit one `EXISTS` subquery per language → OR them into a single wrapper predicate.

## Summary by CodeRabbit

* **New Features**
  * Language-scoped QA check-type filtering with per-language OR semantics.
  * QA check categories included in initial app data; global hooks to access categories/types.
  * QA filter UI: language-scope controls, “select all / no base” options, and updated “Show all” navigation.

* **Bug Fixes / Validation**
  * Malformed QA filter values return a clear 400 validation error.

* **Tests**
  * Expanded tests for language-scoped QA filtering and bad-value handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3667)


webapp/src/globalContext/helpers.tsx
```diff
@@ -63,6 +58,14 @@ export const useOrganizationUsage = () => {
   return useGlobalContext((v) => v.organizationUsage!);
 };
 
+export const useQaCategories = () =>
+  useGlobalContext((c) => c.initialData.qaCheckCategories ?? []);
```
dkrizan: Tiny nit — `?? []` hands back a fresh array each call. Since `useGlobalContext` is `useContextSelector` under the hood (Object.is comparison), every consumer re-renders on every unrelated global-context change while `qaCheckCategories` is still undefined. Initial-data loads quickly so it's barely noticeable in practice, but a module-level constant fixes it cleanly:

```tsx
const EMPTY_QA_CATEGORIES: QaCheckCategoryModel[] = [];

export const useQaCategories = () =>
  useGlobalContext((c) => c.initialData.qaCheckCategories ?? EMPTY_QA_CATEGORIES);
```

Same story with `useQaCheckTypes` — the `flatMap` allocates a new array each render. Fine today (only used in render and click handlers), but worth a `useMemo` if it ever becomes a hook dependency.

backend/api/src/main/kotlin/io/tolgee/hateoas/qa/QaCheckCategoryModelAssembler.kt
```diff
@@ -13,4 +12,7 @@ class QaCheckCategoryModelAssembler {
   ): QaCheckCategoryModel {
     return QaCheckCategoryModel(category = category, checkTypes = checkTypes)
   }
+
+  fun toModelsForAllCategories(): List<QaCheckCategoryModel> =
+    QaCheckType.CATEGORIES.map { (category, checkTypes) -> toModel(category, checkTypes) }
```
dkrizan, line 17: Could we add a small safety-net test that `CATEGORIES` lists every `QaCheckType`? Now that the "any QA issue" shortcut is built from the flattened categories, a new enum value someone forgets to add to `CATEGORIES` would silently vanish from the filter menu, the shortcut, and the settings UI.

Something like:

```kotlin
@Test
fun `CATEGORIES covers every QaCheckType`() {
  val inCategories = QaCheckType.CATEGORIES.values.flatten().toSet()
  assertThat(inCategories).containsExactlyInAnyOrderElementsOf(QaCheckType.entries)
}
```

webapp/src/service/apiSchemaTypes.ts
```diff
@@ -0,0 +1,16 @@
...
+ * we extract them manually here.
+ */
+
+import {
+  QaSettingsRequest,
+  QaIssueModel,
+  SelfHostedEePlanModel,
+} from 'tg.service/apiSchemaTypes.generated';
+
+export type Feature = SelfHostedEePlanModel['enabledFeatures'][number];
```
dkrizan, line 13: Do we actually need this file?

backend/api/src/main/kotlin/io/tolgee/hateoas/qa/QaCheckCategoryModelAssembler.kt
Anty0: I'm like 99% there is a test for that 🤔
...

Found it:
```kotlin
  @Test
  fun `QaCheckType entries order matches flattened CATEGORIES order`() {
    val flattenedCategoryOrder = QaCheckType.CATEGORIES.values.flatten()

    assertThat(flattenedCategoryOrder)
      .`as`("QaCheckType.entries order must match the order of CATEGORIES.values.flatten()")
      .containsExactlyElementsOf(QaCheckType.entries)
  }
```

This already exists on main :3

webapp/src/service/apiSchemaTypes.ts
Anty0: It felt much better than having secondary "imports" in the form of `type Something = something['derived']['from']['api']`. Initially (in a different PR some time ago), I created `apiSchemaTypes.generated.ts`, which automatically exports everything in `components['schemas']`, but it doesn't cover types that exist only inside other exported types, so I've made this "standard place" for such type definitions.

Should I remove it?

dkrizan (approved):

PR #3669, opened 2026-05-18
perf: filter entity pre-commit events by entity type at dispatch

bdshadow:
Make EntityPreCommitEvent and its concrete subclasses generic on the entity type and implement ResolvableTypeProvider so Spring's event multicaster filters listeners by the runtime entity class. Listeners declaring a specific entity type (e.g. EntityPreCommitEvent<Key>) are no longer invoked for unrelated entity writes, skipping transaction-scoped bean resolution and @Transactional proxy entry for entities they do not care about.

EeKeyCountLimitListener is narrowed to EntityPreCommitEvent<Key>.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced event system with improved type safety for event listeners and routing. Events now carry stronger type information, enabling more precise event handler matching and reducing potential type-related issues in event processing infrastructure.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3669?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/events/EntityPreCommitEvent.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/events/EntityPreCommitEvent.kt
@@ -2,7 +2,7 @@ package io.tolgee.events
 
 import io.tolgee.activity.iterceptor.PreCommitEventPublisher
 
-interface EntityPreCommitEvent {
+interface EntityPreCommitEvent<T : Any> {
   val source: PreCommitEventPublisher
-  val entity: Any?
+  val entity: T?
 }
--- a/backend/data/src/main/kotlin/io/tolgee/events/OnEntityCollectionPreUpdate.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/events/OnEntityCollectionPreUpdate.kt
@@ -2,10 +2,15 @@ package io.tolgee.events
 
 import io.tolgee.activity.iterceptor.PreCommitEventPublisher
 import org.springframework.context.ApplicationEvent
+import org.springframework.core.ResolvableType
+import org.springframework.core.ResolvableTypeProvider
 
-class OnEntityCollectionPreUpdate(
+class OnEntityCollectionPreUpdate<T : Any>(
   override val source: PreCommitEventPublisher,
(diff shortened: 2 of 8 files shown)


dkrizan (commented):
Does this bring real perf improvement ? Do you have any before/after stats ? These changes may save a few sub-microseconds, but overall, this is more of a refactor than a perf improvement to me.

bdshadow:
@dkrizan yes, it is a real perf improvement. Less transactions, less db connections, much less events in the local spring  queues. 

For example you create a key with a tag and a screenshot. It means Key, Translation, KeyMeta, Tag, KeyMetaTag, Screenshot, KeyScreenshotReference - 7 writes.

Before - every one of those 7 writes hits all three listeners:
  - StringKeysLimitListeners.onActivity (transaction-scoped bean lookup + `@Transactional` proxy + work, only to fall through when for 5 of 7)
  - SeatUsageListener.permissionAndRolesListener (`@Transactional` proxy + entity-type check fall-through for all 7)
  - SeatUsageListener.userListener (`@Transactional` proxy + entity-type check fall-through
 for all 7)
  - EeKeyCountLimitListener (cheap, falls through for 6 of 7)

That's ~21 transactional-proxy listener invocations, of which only 2 actually do meaningful work (Key, Translation).

After - Spring's multicaster filters by entity class via ResolvableType before any listener is touched:
  - Key write > 1 invocation (onKeyChange) + 1 (EeKeyCountLimitListener).
  - Translation write > 1 invocation (onTranslationChange).
  - KeyMeta, Tag, KeyMetaTag, Screenshot, KeyScreenshotReference → zero listenerinvocations. No `@Transactional` proxy entry, no tx-scoped bean materialization, no activityHolder toggling, no getUsageIncreaseAmount call.

So for that one operation: ~21 transactional listener calls > ~3. 

It really reduces the amount of noise happening around each entity.

bdshadow:
@dkrizan if you prefer to mark this pr as a refactor instead of perf, i'm fine with it. Just let me know

JanCizmar (approved):

PR #3670, opened 2026-05-18
fix: speed up QA batch checks (branch copy, LT cache, chunk batching)

Anty0:
## Summary

Performance and reliability improvements for QA batch checks, motivated by slow runs on real projects (especially those with multiple branches and/or LanguageTool enabled).

### Performance

- **Branch creation copies QA state forward.** New branches inherit `translation_qa_issue` rows and `qa_checks_stale` from their source branch, so freshly created branches no longer trigger a redundant project-wide QA recheck on first activity.
- **LanguageTool responses cached in Redis.** `/v2/check` results are now cached via Spring `@Cacheable` so the cache is shared across backend nodes and survives restarts. The HTTP call is wrapped in a small retry to ride out transient failures (e.g. a LanguageTool instance restarting mid-request).
- **Chunk-batched QA processing.** The QA batch processor now reads inputs and persists issues for a whole chunk at once: bulk-loaded translations, keys, base translations, and existing issues per chunk; checks run outside the DB transaction so network-bound checks (LanguageTool) do not hold row locks; new issues persisted via a single batched insert per chunk.

### Reliability — make sure translations don't get stuck stale

- **After branch creation,** if QA is enabled on the project and any translation on the new branch is `qa_checks_stale = true`, a targeted QA batch job is enqueued for those translations.
- **After every QA batch job finalizes,** if there are still stuck-stale translations in the project (and no other QA job is already queued for it), a follow-up QA batch job is enqueued for them and a Sentry warning is logged. Catches edge cases where a translation might otherwise sit unrechecked.

### Benchmark

Two workloads:

- **50k**: 1000 keys × 50 languages = 50 000 translations.
- **1k**: 100 keys × 10 languages = 1 000 translations.

In both: SPELLING/GRAMMAR off (LT tested separately below), `tolgee.batch.concurrency = 1`, one run per cell, localhost, single backend node, Caffeine for the non-LT cells.

| variant                                  | wall 50k  | wall 1k  | vs main (50k) | vs main (1k) |
|------------------------------------------|----------:|---------:|--------------:|-------------:|
| main, default knobs (chunk=1000, jdbc=100) | 1 428 s | 28.6 s   | 1.00×         | 1.00×        |
| perf branch, chunk=1000, jdbc=100        | 1 225 s   |   3.0 s  | 1.17×         | 9.5×         |
| perf branch, **chunk=100, jdbc=1000** ★  | **496 s** | **2.6 s**| **2.88×**     | **11×**      |

The same-config gap between main and the perf branch is much larger at the small scale (~9.5×) than at large scale (~1.17×). Per-translation runtime drops from ~27 ms (main, chunk=1000) to ~1.5 ms (perf branch, chunk=100) at 1k — at 50k the per-translation cost grows because of dataset-size-dependent effects (more existing rows scanned per chunk, larger working set).

Driven by sweeping the two new tunables.

**Chunk size (jdbc=1000 at 1k; jdbc=100 at 50k):**

| chunk | wall 50k (s) | wall 1k (ms) |
|------:|-------------:|-------------:|
| 50    | 595          | 2 480        |
| **100** | **509**    | **2 601**  ← default |
| 250   | 899          | 2 614        |
| 1000  | 1 225        | 3 001        |
| 2000  | 1 236        | 2 600        |
| 5000  | 2 401        | 4 425        |

Per-item cost grows with chunk size, and the growth is much steeper at the 50k scale than at 1k (chunk=5000 is 4.7× slower than chunk=100 at 50k vs only 1.7× at 1k). Likely culprits (not measured precisely, candidates ordered by likelihood): Postgres planner switching from index scan to seq scan once the bulk-`SELECT … WHERE id IN (…)` list crosses a threshold; N²-ish in-memory work merging existing-vs-new issues; longer-held Hibernate session for one fat transaction. Sweet spot is 100 on both scales; smaller (50) starts paying back per-chunk overhead.

**JDBC bulk-insert batch (chunk=100):**

| jdbc | wall 50k (s) | wall 1k (ms) |
|-----:|-------------:|-------------:|
| 100  | 509          | 2 547        |
| 500  | 519          | 2 619        |
| 1000 | **496**      | **2 601** ← default |

Marginal effect, but free improvement — bumped to 1000 to align with `hibernate.jdbc.batch_size`.

**Glossary impact** (50k, relative to no-glossary, best knobs):

| glossary terms | overhead |
|---------------:|---------:|
| 100            | +20 %    |
| 500            | +52 %    |

`QaCheckBatchServiceImpl.findGlossaryTerms` currently queries per (text, language) for every translation. Scales linearly with `translations × matching_terms`. Acceptable today; future opportunity is to hoist the lookup to a per-chunk batch query if a customer with a large glossary reports slow QA.

**LanguageTool** (1 000 translations, SPELLING + GRAMMAR on):

| variant                             | wall time | vs main |
|-------------------------------------|----------:|--------:|
| main baseline                       |    640 s  | 1.00×   |
| perf branch, cold LT cache          |    512 s  | 1.25×   |
| perf branch, **warm Redis cache**   |  **1.3 s**| **~490×** |

S0 → S1 is the chunk-batching speedup (LT round-trips dominate). S1 → S2 is the new Redis-backed LT cache — once a text has been checked, every subsequent check of the same text is a cache hit.

### Recommended defaults

Landed in this PR:

- `QaCheckChunkProcessor.getChunkSize`: **100**
- `QaIssueService.insertIssues` `jdbcTemplate.batchUpdate` batch size: **1000**

### Caveats on the numbers

Single-machine, localhost, single backend node, one run per cell — directionally correct, ±10 % noise expected (DB state also accumulated across cells). 50k and 1k runs were separate sessions, so absolute numbers across the two scales are comparable directionally but not to better than ~15 %.

### Tests

- New tests for branch QA issue/flag carry-over, branch-creation stale recheck, LanguageTool retry behaviour, the multi-translation chunk path, and the self-heal listener.
- Existing QA and branching test suites continue to pass.

### Breaking change — frontend websocket payload

`qa-issues-updated` events now carry `data` as a list of per-translation updates instead of a single object (the backend emits one event per batch-job chunk). The webapp consumer has been updated accordingly.

## Summary by CodeRabbit

* **New Features**
  * QA checks run as chunked batches for faster, more efficient processing.
  * LanguageTool integration added with retry and configurable timeouts.
  * Automatic recheck of stale translations when branches are created and self-heal rechecks on batch finalization.

* **Improvements**
  * QA issues are copied during branch operations and preserved correctly.
  * Stale-translation queries improved with a targeted DB index.
  * Websocket payload handling updated to support multiple QA issue updates per event.

* **Tests**
  * New integration and unit tests covering QA batching, LanguageTool retries, and branch QA copying.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3670?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
@@ -215,9 +215,6 @@ interface TranslationRepository : JpaRepository<Translation, Long> {
     languagesIds: List<Long>,
   ): List<Translation>
 
-  /**
-   * Eagerly fetches `key` and `language`
-   */
   @Query(
     """
     from Translation t
--- a/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
@@ -248,6 +248,17 @@ interface TranslationRepository : JpaRepository<Translation, Long> {
 
   @Query(
     """
+    select new io.tolgee.dtos.queryResults.qa.KeyLanguagePairView(k.id, t.language.id)
+    from Translation t
+    join t.key k
+    where k.project.id = :projectId
+      and t.qaChecksStale = true
+    """,
+  )
+  fun getStaleKeyLanguagePairsByProject(projectId: Long): List<KeyLanguagePairView>
+
+  @Query(
+    """
     from Translation t
     join t.key k
     where k.project.id = :projectId
--- a/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
@@ -214,9 +214,20 @@ interface TranslationRepository : JpaRepository<Translation, Long> {
     languagesIds: List<Long>,
   ): List<Translation>
 
-  fun getAllByKeyIdInAndLanguageIdIn(
-    keyIds: List<Long>,
-    languageIds: List<Long>,
+  /**
+   * Eagerly fetches `key` and `language`
+   */
+  @Query(
+    """
+    from Translation t
+    join fetch t.key
+    join fetch t.language
+    where t.key.id in :keyIds and t.language.id in :languageIds
+    """,
+  )
+  fun findAllWithKeyAndLanguageByKeyIdInAndLanguageIdIn(
+    keyIds: Collection<Long>,
+    languageIds: Collection<Long>,
   ): List<Translation>
 
   @Query(
--- a/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
@@ -1,5 +1,6 @@
 package io.tolgee.repository
 
+import io.tolgee.dtos.queryResults.qa.KeyLanguagePairView
 import io.tolgee.jobs.migration.translationStats.StatsMigrationTranslationView
 import io.tolgee.model.Project
 import io.tolgee.model.key.Key
@@ -229,6 +230,21 @@ interface TranslationRepository : JpaRepository<Translation, Long> {
     keyIds: Collection<Long>,
     languageIds: Collection<Long>,
   ): List<Translation>
+
+  @Query(
+    """
+    select new io.tolgee.dtos.queryResults.qa.KeyLanguagePairView(k.id, t.language.id)
+    from Translation t
+    join t.key k
+    where k.project.id = :projectId
+      and k.branch.id = :branchId
+      and t.qaChecksStale = true
+    """,
+  )
+  fun getStaleKeyLanguagePairsByBranch(
+    projectId: Long,
+    branchId: Long,
+  ): List<KeyLanguagePairView>
 
   @Query(
     """
--- a/backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
@@ -416,7 +416,10 @@ class LanguageService(
     )
   }
 
-  fun findByIdIn(ids: Iterable<Long>): List<Language> {
+  fun findByIdIn(ids: Collection<Long>): List<Language> {
+    if (ids.isEmpty()) {
+      return emptyList()
+    }
     return languageRepository.findAllById(ids)
   }
 
--- a/backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
@@ -416,10 +416,7 @@ class LanguageService(
     )
   }
 
-  fun findByIdIn(ids: Collection<Long>): List<Language> {
-    if (ids.isEmpty()) {
-      return emptyList()
-    }
+  fun findByIdIn(ids: Iterable<Long>): List<Language> {
     return languageRepository.findAllById(ids)
   }
 
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/component/qa/QaActivityListener.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/component/qa/QaActivityListener.kt
@@ -6,6 +6,7 @@ import io.tolgee.batch.BatchJobService
 import io.tolgee.batch.data.BatchJobType
 import io.tolgee.batch.data.BatchTranslationTargetItem
 import io.tolgee.batch.events.OnBatchJobFinalized
+import io.tolgee.batch.processors.QaCheckChunkProcessor
 import io.tolgee.batch.request.QaCheckRequest
 import io.tolgee.component.enabledFeaturesProvider.EnabledFeaturesProvider
 import io.tolgee.component.eventListeners.BypassableActivityListener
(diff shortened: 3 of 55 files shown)


backend/data/src/main/kotlin/io/tolgee/repository/TranslationRepository.kt
```diff
@@ -202,11 +203,48 @@ interface TranslationRepository : JpaRepository<Translation, Long> {
     languagesIds: List<Long>,
   ): List<Translation>
 
-  fun getAllByKeyIdInAndLanguageIdIn(
-    keyIds: List<Long>,
-    languageIds: List<Long>,
+  /**
+   * Eagerly fetches `key` and `language`
```
dkrizan: unnecessary comment

backend/data/src/main/kotlin/io/tolgee/service/language/LanguageService.kt
```diff
@@ -416,7 +416,10 @@ class LanguageService(
     )
   }
 
-  fun findByIdIn(ids: Iterable<Long>): List<Language> {
+  fun findByIdIn(ids: Collection<Long>): List<Language> {
+    if (ids.isEmpty()) {
```
bdshadow: the check is redundant. `findAllById` does it as the first step - check `SimpleJpaRepository`

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/qa/QaActivityListener.kt
```diff
@@ -99,6 +99,16 @@ class QaActivityListener(
     }
   }
 
+  @EventListener
+  fun onQaBatchJobFinalizedSelfHeal(event: OnBatchJobFinalized) {
+    if (bypass) return
+    if (event.job.type != BatchJobType.QA_CHECK) return
+    val projectId = event.job.projectId ?: return
+    runSentryCatching {
+      qaRecheckService.recheckStuckStaleTranslationsInProject(projectId)
```
bdshadow, line 115: could this cause an infinite loop here? 

For example, a QA check throws on a specific translation, the stale flag is never cleared, the next finalize enqueues another job for it, which throws, which enqueues another…

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/LanguageToolApiClient.kt
```diff
@@ -0,0 +1,134 @@
...
+  val isConfigured: Boolean
+    get() = tolgeeProperties.languageTool.url.isNotBlank()
+
+  fun checkConfigured() {
+    if (!isConfigured) {
+      throw LanguageToolNotConfiguredException()
+    }
+  }
+
+  @Cacheable(Caches.LANGUAGE_TOOL_RESULTS, key = "{#resolvedTag, #text}")
```
bdshadow: on long text parameter, the key in redis can be in fact very long - which is not recommended. In cloud translation is limit is 10k chars.
Maybe smth like `key = "{#resolvedTag, T(...).sha256(#text)}"` is better

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/QaCheckBatchServiceImpl.kt
```diff
@@ -45,101 +47,182 @@ class QaCheckBatchServiceImpl(
...
+    val missingLanguageIds = languageIds.filter { it !in tagsFromTranslations }
+    val missingLanguageTags = languageService.findByIdIn(missingLanguageIds).associate { it.id to it.tag }
+    val languageTagById = tagsFromTranslations + missingLanguageTags
 
-      // Disable activity logging — no content changes are happening here.
-      // Without this, when we create a new empty translation (so we can reference it from QA issues),
-      // it gets logged.
-      translation.disableActivityLogging = true
-      translationService.save(translation)
+    val keyById = keyRepository.findAllByIdIn(keyIds).associateBy { it.id }
```
bdshadow, line 108: just a quick check. Can it be possible be called here for more than 65k keys? If yes, then we'll end up with a failed PreparedStatement construction.

There are actually more new in-queries in the `TranslationQaIssueRepository`. Please check them too

bdshadow (changes requested):

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/qa/QaActivityListener.kt
Anty0: Yup, technically yes. I did try to isolate QA checks, though - they shouldn't throw; instead, they'll generate an issue telling the user there was an error running the check.

If we encounter a situation where the system would outright fail, it could loop.
To make it a bit safer, I'll add a flag to the batch job params whether it should do this check at the end - if set, we would skip this, so it would run at max one more time every time there is a QA batch job - sounds more reasonable.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/qa/QaCheckBatchServiceImpl.kt
Anty0: All places that create QA batch jobs limit the batch job size to 32k items (combinations of key and language), so there can't be more than 32k keys, I believe.

But it does assume we never create a bigger batch job - I wanted to ask whether we don't want to add a similar batch job size limit to the batch processing system itself instead. Since the way it's written right now, it has to parse the whole JSON for every chunk it wants to process - if we're creating bigger batch jobs that that, we might start hitting slowdowns from just parsing the JSON list of all items.

bdshadow: Recently, I have added a batch size of 5000 for all jobs that had had it unlimited: https://github.com/tolgee/tolgee-platform/pull/3633. Default is still infinite, but by now, i think, we're safe.

bdshadow (dismissed):

bdshadow:
approved. There is just a compilation error in `QaCheckChunkProcessorTest `

bdshadow (approved):

PR #3688, opened 2026-05-22
fix: translation memory UI issues

dkrizan:
## Translation Memory UI fixes

- **Language filter** — no longer pre-selects every org language. The selection is seeded once from the languages on the first page of entries, then persisted and kept sticky across visits (manual changes stick too).
- **Virtual entry checkboxes** — project-mirrored entries now show a *disabled* checkbox with an explanatory tooltip, instead of no checkbox at all.
- **"Base" badge** — the header's "Source" caption next to the base translation is replaced by a grey "Base" badge.
- **New TM dialog** — the base language field has no default selection, forcing the user to choose one explicitly (avoids accidentally picking the wrong base language).
- **Base-language validation** — the required error now appears after a submit attempt even when the field was selected and then cleared.
- **Select-all checkbox** — the header checkbox can now deselect all entries (previously it could only ever select).

### Notes

Two new translation keys are used — `translation_memory_header_base` and `tm_entry_select_disabled_virtual`. Both have `defaultValue`s in code and still need to be registered in Tolgee.

## Summary by CodeRabbit

* **New Features**
  * Auto-seed language filters from visible entries; newly created entry languages are added to the filter.
  * Create-entry flow now returns created language tags so the UI can update filters.

* **UI/UX Improvements**
  * Base language shown as a compact chip in the translation memory header.
  * Language selection is explicit (no “all languages” mode).
  * Select-all toggles select/unselect all; virtual/synced rows show explanatory tooltips.
  * Option to disable auto-selecting the first language in selectors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3688?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/webapp/src/component/languages/BaseLanguageSelect.tsx
+++ b/webapp/src/component/languages/BaseLanguageSelect.tsx
@@ -61,6 +61,7 @@ type Props = {
   disabled?: boolean;
   label?: React.ReactNode;
   minHeight?: boolean;
+  autoSelectFirst?: boolean;
 } & Omit<ComponentProps<typeof Box>, 'children' | 'minHeight'>;
 
 export const BaseLanguageSelect: React.VFC<Props> = ({
@@ -69,6 +70,7 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
   disabled,
   label,
   minHeight,
+  autoSelectFirst = true,
   ...boxProps
 }) => {
   const { preferredOrganization } = usePreferredOrganization();
@@ -76,8 +78,12 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
   const { t } = useTranslate();
   const [field, meta] = useField(name);
   const value = field.value as SelectedLanguageModel | undefined;
-  // Formik returns error as object - schema is incorrect...
-  const error = (meta.error as any)?.tag;
+
+  const error = React.isValidElement(meta.error)
+    ? meta.error
+    : (meta.error as any)?.tag;
+
+  const showError = (meta.touched || context.submitCount > 0) && error;
 
   const [search, setSearch] = useState('');
   const [searchDebounced] = useDebounce(search, 500);
@@ -129,7 +135,13 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
   // The ref prevents re-triggering after setSelected updates the Formik field.
   const hasAutoSelected = useRef(false);
   useEffect(() => {
-    if (!value && data && data.length > 0 && !hasAutoSelected.current) {
+    if (
+      autoSelectFirst &&
+      !value &&
+      data &&
+      data.length > 0 &&
+      !hasAutoSelected.current
+    ) {
       hasAutoSelected.current = true;
       setSelected(data[0]);
     }
@@ -216,7 +228,7 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
         labelItem={labelItem}
         inputComponent={FlaggedInput}
         label={label ?? t('field_base_language', 'Base language')}
-        error={meta.touched && error}
+        error={showError}
         searchPlaceholder={t('language_search_placeholder')}
         disabled={disabled}
         minHeight={minHeight}
--- a/webapp/src/component/languages/BaseLanguageSelect.tsx
+++ b/webapp/src/component/languages/BaseLanguageSelect.tsx
@@ -79,7 +79,13 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
   const [field, meta] = useField(name);
   const value = field.value as SelectedLanguageModel | undefined;
 
+  // meta.error shape depends on which Yup rule fired:
+  //   - object-level `.required()` (field cleared / undefined) → string
+  //   - shape `{ tag: ... }` validation → { tag: string }
+  //   - localized message → ReactElement
   const error = React.isValidElement(meta.error)
+    ? meta.error
+    : typeof meta.error === 'string'
     ? meta.error
     : (meta.error as any)?.tag;
 
--- a/webapp/src/ee/translationMemory/components/content/TmEntriesListHeader.tsx
+++ b/webapp/src/ee/translationMemory/components/content/TmEntriesListHeader.tsx
@@ -48,10 +48,8 @@ export const TmEntriesListHeader: React.VFC<Props> = ({
           checked={selectionService.isAllSelected}
           indeterminate={selectionService.isSomeSelected}
           disabled={selectionService.isLoading}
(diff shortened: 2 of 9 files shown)


webapp/src/component/languages/BaseLanguageSelect.tsx
```diff
@@ -69,15 +70,20 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
...
   const [field, meta] = useField(name);
   const value = field.value as SelectedLanguageModel | undefined;
-  // Formik returns error as object - schema is incorrect...
-  const error = (meta.error as any)?.tag;
+
+  const error = React.isValidElement(meta.error)
+    ? meta.error
+    : (meta.error as any)?.tag;
+
+  const showError = (meta.touched || context.submitCount > 0) && error;
```
Anty0, line 92: It might be worth implementing the `submitCount` check for other forms in future PRs, too. We don't do it anywhere else, but it kind of makes sense to show the errors if the user tries to submit.

webapp/src/component/languages/BaseLanguageSelect.tsx
```diff
@@ -69,15 +70,20 @@ export const BaseLanguageSelect: React.VFC<Props> = ({
...
   const context = useFormikContext();
   const { t } = useTranslate();
   const [field, meta] = useField(name);
   const value = field.value as SelectedLanguageModel | undefined;
-  // Formik returns error as object - schema is incorrect...
-  const error = (meta.error as any)?.tag;
+
+  const error = React.isValidElement(meta.error)
+    ? meta.error
+    : (meta.error as any)?.tag;
```
Anty0, line 90: AI agent believes that this might not work for cases when the value is set to `undefined`, because in that case, the validation doesn't fail on the `tag` but on the parent itself - the `error` will be a string. Needs to be validated.

webapp/src/ee/translationMemory/components/content/TmEntriesListHeader.tsx
```diff
@@ -47,7 +48,15 @@ export const TmEntriesListHeader: React.VFC<Props> = ({
           checked={selectionService.isAllSelected}
           indeterminate={selectionService.isSomeSelected}
           disabled={selectionService.isLoading}
-          onChange={selectionService.toggleSelectAll}
+          // `toggleSelectAll` keys off `isAllSelected`, which can never be true here:
+          // `totalCount` counts virtual rows but only stored entries are selectable, so the
+          // selected count never reaches it. Toggle on "anything selected" instead — a click
+          // clears the selection when something is selected, otherwise selects all.
```
Anty0: I... I don't follow the comment. It talks about properties that aren't around here. Feels like leftover AI blabber.

The `onChange` is mostly self-explanatory. A note about "we don't try to figure out whether everything is selected" could be fair, but I couldn't parse such info from the comment.

webapp/src/ee/translationMemory/components/content/TmEntriesListHeader.tsx
```diff
@@ -62,9 +71,12 @@ export const TmEntriesListHeader: React.VFC<Props> = ({
...
       </span>
-      <Typography variant="caption" color="text.secondary">
-        <T keyName="translation_memory_header_source" defaultValue="Source" />
-      </Typography>
+      <DefaultChip
+        data-cy="tm-entries-header-base-badge"
+        label={
+          <T keyName="translation_memory_header_base" defaultValue="Base" />
+        }
+      ></DefaultChip>
```
Anty0: nit: probably could be self-closed

webapp/src/ee/translationMemory/components/content/TmEntriesToolbar.tsx
```diff
@@ -170,14 +165,10 @@ export const TmEntriesToolbar: React.VFC<Props> = ({
...
-            <Box data-cy="tm-entries-language-filter">
+            <Box data-cy="tm-entries-language-filter" sx={{ width: 250 }}>
               <InfiniteMultiSearchSelect
                 items={languages}
-                selected={
-                  isAllSelected
-                    ? (languages ?? []).map((l) => l.tag)
-                    : selectedLanguages
-                }
+                selected={selectedLanguages ?? []}
```
Anty0: nit: Reminds me of the "new array every render" I got on a review from you :3
I assume it's probably fine here - I don't see it causing many re-renders here.

webapp/src/ee/translationMemory/components/content/TranslationMemoryEntriesList.tsx
```diff
@@ -105,6 +105,28 @@ export const TranslationMemoryEntriesList: React.VFC<Props> = ({
...
+    if (!firstPage) return undefined;
+    const tags = new Set<string>();
+    firstPage.forEach((r) =>
+      r.cells.forEach((c) => tags.add(c.targetLanguageTag))
+    );
+    return [...tags];
+  }, [entries.data]);
+
+  // One-time seed: when nothing is stored for this TM, set the selection from the first
+  // page's languages and persist it.
```
Anty0: nit: Not sure if it's worth keeping this one. "one time seed" isn't hard to read from code. "persist it" could easily go stale, since the persistence isn't handled here.

webapp/src/ee/translationMemory/components/content/TranslationMemoryEntriesList.tsx
```diff
@@ -156,7 +178,9 @@ export const TranslationMemoryEntriesList: React.VFC<Props> = ({
         editingLang={editingLangForRow}
         canManage={canManage}
         layout={layout}
-        selectionService={row.editable ? selectionService : undefined}
+        // Passed for every row — the row itself renders a real checkbox for editable
+        // rows and a disabled one (with tooltip) for virtual rows.
```
Anty0: nit: Seems kind of unrelated here. I'm not sure I'd be looking here to figure out how the logic works - probably would go straight inside the item. Also, same as the other one - since the logic isn't here, I can easily see it going stale.

webapp/src/ee/translationMemory/hooks/useTmLanguageFilter.ts
```diff
@@ -75,30 +76,37 @@ export function useTmLanguageFilter({
     }
   };
 
+  // One-time seed of the persisted selection from the first page of entries. The single
+  // caller guards on `selectedLanguages === undefined`, so this only runs when nothing is
```
Anty0: nit: "caller guards..." if that's a "limitation" I'd document it as such, but probably not worth keeping the comment. As is, it's more of a "will become stale in the future" again.

Anty0 (dismissed):

PR #3690, opened 2026-05-25
fix: chunk soft delete to avoid PostgreSQL parameter limit

bdshadow:
## Summary

- `KeyService.softDeleteMultiple` issued a single `findAllByIdIn(ids)`, which expanded into an IN clause with one parameter per id.
- PostgreSQL's wire protocol caps prepared-statement parameters at 65,535, so soft-deleting that many keys at once fails with `PreparedStatement can have at most 65,535 parameters`.
- Chunk the operation at 10,000 ids per query (constant `SOFT_DELETE_CHUNK_SIZE`) to stay safely under the cap. Same pattern as the existing `getByIds` helper in this file.

## Test plan

- [ ] Soft-delete >65k keys via the batch endpoint or service call and verify it completes without a JDBC parameter-count error.
- [ ] Existing soft-delete tests (`SoftDeleteTest`, `BatchDeleteKeysTest`, `KeyTrashPurgeSchedulerTest`) still pass.

## Summary by CodeRabbit

* **Refactor**
  * Optimized bulk key deletion operations to process in smaller batches, improving performance and reducing memory consumption during large-scale deletion tasks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3690?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/key/KeyService.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/key/KeyService.kt
@@ -373,13 +373,15 @@ class KeyService(
     deletedBy: UserAccount? = null,
   ) {
     val actualDeletedBy = deletedBy ?: authenticationFacade.authenticatedUserEntityOrNull
-    val keys = keyRepository.findAllByIdIn(ids.toList())
     val now = currentDateProvider.date
-    keys.forEach {
-      it.deletedAt = now
-      it.deletedBy = actualDeletedBy
+    ids.chunked(SOFT_DELETE_CHUNK_SIZE).forEach { chunk ->
+      val keys = keyRepository.findAllByIdIn(chunk)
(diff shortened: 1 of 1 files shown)


dkrizan (commented):
@bdshadow 
Can you clarify whether this is fixing a real failure or being proactive? I can't see a path where `softDeleteMultiple` actually receives more than 65k IDs today.

JanCizmar (approved):

bdshadow:
@dkrizan  look, i'm 100% sure i replicated it locally earlier.... but, i wasn't able to do it now....

TBH, i would do it everywhere, maybe as an extension function, for example. We've got this exception pretty often recently.

I will leave it up to you here. Feel free to close it. I would merge it now, and later would think of a more general solution

PR #3694, opened 2026-05-26
feat: support gettext msgctxt in PO import/export and UI

### LINKED ISSUE #3053: Store `msgctxt` from Gettext (.po) files

Anty0:
Supersedes #3459
Fixes #3053

## Summary
- **Import:** the PO parser now captures the optional `msgctxt` field. On import, Tolgee stores the pair `(msgctxt, msgid)` as a single key name using the U+0004 EOT separator — the same convention GNU Gettext uses internally in MO files — so two entries with the same `msgid` but different `msgctxt` remain distinct keys.
- **Export:** the PO exporter splits the key name on U+0004 and writes the `msgctxt` line back before `msgid`. Plural fallback now uses the split `msgid` instead of the raw key name.
- **Parser cleanup:** header detection now requires both `msgid` and `msgctxt` to be empty, so a malformed msgctxt-only entry can't be misread as the header block. The legacy `PO_MSGCTXT_NOT_SUPPORTED` file issue is no longer emitted (the enum value stays in place for back-compat with previously stored issues).
- **UI display:** a new `<KeyName>` component renders the `msgctxt` portion as a styled chip in front of the `msgid` across all key-name display sites (translations grid, simple/list view, single-key page breadcrumb, activity log, import dialogs, translation-memory rows and tooltips, branching/merge header, screenshot label tooltips). The window title intentionally still shows the raw key — the chip cannot render in a `document.title` string.
- **UI edit:** the `<Editor>` component gains a `keyName` mode that wires a new CodeMirror plugin (`KeyNamePlugin` from `@tginternal/editor`) decorating the U+0004 with an atomic widget. Display and edit are now visually identical — the previous substitution-based workaround (`keyNameForEditing`/`keyNameFromEdited` + `␄` visible-char) is removed.
- **Vite config:** added `@codemirror/state`, `@codemirror/view`, `@codemirror/lint` to `resolve.dedupe` to avoid the "multiple instances of @codemirror/state" runtime error introduced when `preserveSymlinks: true` is combined with the linked `@tginternal/editor` worktree.

## Notes
- Re-importing a `.po` file that was imported before this change (when msgctxt was silently dropped) will create new keys for the `(msgctxt, msgid)` pairs instead of updating the existing `msgid`-only keys. Accepted breakage — most users import once.
- An empty `msgctxt ""` collapses to a plain key without a chip, matching gettext semantics.

## Companion PRs
- editor: https://github.com/tolgee/editor/pull/8
- tolgee-js: https://github.com/tolgee/tolgee-js/pull/3523
- documentation: https://github.com/tolgee/documentation/pull/1111

## Summary by CodeRabbit

* **New Features**
  * Full PO msgctxt support: import/export preserves and uses context to disambiguate identical strings.
  * Key display shows visual context indicators with a tooltip linking to docs via a new KeyName component.
  * Key editor gains a new "keyName" mode for viewing/editing keys with context.
  * Import ignores stray context-only entries to avoid header mis-parsing.

* **Tests**
  * Added unit and end-to-end tests for msgctxt parsing, escaping, import/export, fixtures and UI behavior.

* **Chores**
  * Editor package updated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tolgee/tolgee-platform/pull/3694?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/PoMsgctxtTestData.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/PoMsgctxtTestData.kt
@@ -1,0 +1,19 @@
+package io.tolgee.development.testDataBuilder.data
+
+class PoMsgctxtTestData : BaseTestData() {
+  init {
+    projectBuilder.apply {
+      addKey {
+        name = "plain.key"
+      }.build {
+        addTranslation("en", "Plain key without msgctxt")
+      }
+
+      addKey {
+        name = "menu\u0004Open"
+      }.build {
+        addTranslation("en", "Open file from the menu")
+      }
+    }
+  }
+}
--- a/backend/data/src/main/kotlin/io/tolgee/formats/po/contsants.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/po/contsants.kt
@@ -1,3 +1,5 @@
 package io.tolgee.formats.po
 
 const val PO_FILE_MSG_ID_PLURAL_CUSTOM_KEY = "_poFileMsgIdPlural"
+
+const val PO_MSGCTXT_KEY_SEPARATOR = "\u0004"
--- a/backend/data/src/main/kotlin/io/tolgee/formats/po/in/PoFileProcessor.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/po/in/PoFileProcessor.kt
@@ -6,6 +6,7 @@ import io.tolgee.formats.ImportFileProcessor
 import io.tolgee.formats.MessageConvertorResult
 import io.tolgee.formats.importCommon.ImportFormat
 import io.tolgee.formats.po.PO_FILE_MSG_ID_PLURAL_CUSTOM_KEY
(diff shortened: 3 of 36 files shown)


JanCizmar (changes requested):
Good job! Thanks 👍

- Missing e2e tests

JanCizmar (approved):

JanCizmar:
Awesome job! Thx.

PR #3696, opened 2026-05-26
perf: batch TM lookups per chunk in pretranslate/auto-translate jobs

bdshadow:
## Summary

The `pretranslate_by_tm` and `AUTO_TRANSLATE` batch jobs were issuing one TM-lookup SQL round-trip per chunk item — plus a couple of lazy entity loads behind it — for ~3 DB round-trips per key. For a 2000-key pretranslate run at chunk size 10, that's ~6000 round-trips and ~7.5s of pure DB time on a local Postgres.

This PR does two things:

1. **Batches the TM lookup** so all `(keyId, targetLanguageId)` pairs in a chunk are resolved in a single statement via a LATERAL join with `LIMIT 1` per input pair. Semantics are preserved exactly — the LATERAL query mirrors the existing `TranslationRepository.getTranslationMemoryValue` HQL (cross-project, base-language exact match, `k <> req_key` self-exclusion) for OSS, and the EE-side UNION ALL stored+virtual query (with `tmIds` + penalty + `write_access` gates) for EE.
2. **Adds a hash index on `translation.text`** so the planner can drive the virtual half from the text equality predicate instead of fanning out across every key in every contributing project. Without this index, prod EXPLAIN ANALYZE showed both the per-item (~825 ms) and batched (~4360 ms for 10) paths bottlenecked on a missing index — the algorithmic win from (1) isn't realized on a real Cloud project until (2) is in place.

## Changes

- **`TmAutoTranslateProvider`** — new default method `getAutoTranslatedValuesForChunk(items, keysById, languagesById)`. Default impl loops the per-item method; intended to be overridden.
- **`TmAutoTranslateProviderOssImpl`** — overrides with a single native LATERAL query that mirrors the existing HQL semantics. `unnest(reqKeyIds, reqLangIds)` zips the actual pairs into the inputs CTE so the SQL processes exactly `items.size` input rows, not the cross-product of distinct keys × distinct langs.
- **`TmAutoTranslateProviderEeImpl`** — overrides with a batched UNION ALL (stored entries + virtual rows) over readable `tmIds` with the same penalty / `write_access` gates as the per-item path. Items not matched by the EE-aware half are delegated to the OSS batched fallback (via an unmetered helper so the batch metric isn't double-recorded). Bounded at 3 DB round-trips per chunk total (tmIds lookup + EE batched + OSS fallback) regardless of chunk size.
- **`AutoTranslationService`** — `autoTranslateSync`, `softAutoTranslate`, `autoTranslateUsingTm`, and the private `getAutoTranslatedValue(config, translation)` gain an optional `precomputedTmMatches: Map<Pair<Long, Long>, TranslationMemoryItemView>?` parameter. When present, the TM branch reads from the map; otherwise the per-item provider call runs unchanged (preserves UI / single-translation callers).
- **`GenericAutoTranslationChunkProcessor`** / **`AutoTranslateChunkProcessor`** — precompute the chunk's TM map once via `tmAutoTranslateProvider.getAutoTranslatedValuesForChunk(...)` and pass it down to each per-item call. Items where `useTranslationMemory = false` skip the precompute entirely.
- **`Metrics.recordTranslationMemoryLookupBatch`** — new metric capturing chunk size, hit count, and duration.
- **Liquibase migration** — adds `translation_text_hash_idx` (PostgreSQL hash index on `translation.text`, built `CONCURRENTLY`). See "Index migration" below for the rationale.
- **Tests** — 8 unit tests in `TranslationMemoryServiceBatchTest` covering empty input, single match, self-exclusion, no-match, shared `(baseText, targetLangId)` across chunk items, mixed target languages, and unrequested-pair filtering. Plus a `@Disabled` `TranslationMemoryServiceBenchmark` that runs the per-item and batched paths back-to-back on the same fixture (drop `@Disabled` locally to compare numbers).
- **Test mock stubs** — `BatchJobTestUtil`, `AbstractBatchJobConcurrentTest`, and `BatchJobManagementControllerTest` updated for the new optional parameter on `autoTranslateSync`.

## Index migration

The auto-translate path filters `WHERE base_t.text = ?`. Three trigram indexes already exist on `translation.text` (`translation_text_gin_trgm_index`, `translation_text_unaccent_gin_trgm_index`, `translation_lang_text_gist_trgm`) — but they all use `_trgm` opclass which serves `%` / `LIKE` / similarity, **not** exact equality. The planner has no usable index for `text = ?`.

Without that index, the virtual half's join order ends up as:
```
tm_virt (10 from :tmIds)
  → tmp_w (12 contributing projects)
  → key k (3170 keys per project = 38k keys total)
  → target_t (probe per key)
  → base_t (probe per key) + Filter: text = ?  ← evaluated LAST, drops 99.99%
```

Production EXPLAIN ANALYZE on a project with shared TMs across many others:

| | per-item | batched-10 |
|---|---:|---:|
| Execution time | **825 ms** | **4360 ms** |
| Shared buffer hits | 402,952 | 5,414,211 |
| Keys visited in virtual half | 38,037 | 380,370 (10× same) |

The batched query is doing the same wasteful work N times in one round-trip — the algorithmic win is real (10 chunk-items would be ~8.3 s if run serially, batched does it in 4.4 s), but neither baseline is acceptable.

`CREATE INDEX CONCURRENTLY translation_text_hash_idx ON translation USING hash (text)` flips the join order: the planner probes by `text = ?` first (highly selective, single-digit rows), then walks outward to project / TM / target. Expected drop: per-item to single-digit ms, batched-10 to ~tens of ms.

Hash chosen over btree because the auto-translate path only ever uses `=`. Hash indexes are smaller and have lighter write amplification on a hot table (which already carries 3 trigram indexes on the same column). Built `CONCURRENTLY` so the migration doesn't lock writes; `runInTransaction="false"` on the changeSet with an explicit `<rollback>` for idempotency.

## Benchmark results (local, no index needed because the test fixture is small)

20,000 TM source keys, 2,000 request keys to translate, ~5,000 distinct base texts (~4 candidates per lookup), German target, local Postgres, JDK 21:

| chunk size | per-item (= main) | batched (this PR) | speedup |
|-----------:|------------------:|------------------:|--------:|
| 10         | avg 8230 ms       | avg 268 ms        | **30.7×** |
| 50         | avg 8546 ms       | avg 218 ms        | **39.2×** |

Production `pretranslate_by_tm` uses chunk size 10 → ~30× win on this fixture, but the real prod win comes from the index migration (see above). EE-side benchmark not measured here because the fixture doesn't populate `translation_memory_entry` rows; the algorithmic improvement is the same shape (1-3 round-trips per chunk vs ~3 per item).

## Per-item path is untouched

The optional `precomputedTmMatches` parameter defaults to `null`, in which case `AutoTranslationService` falls through to the existing per-item `tmAutoTranslateProvider.getAutoTranslatedValue(...)`. UI calls, single-translation auto-translate, the retry loop in `autoTranslateSyncWithRetry`, and any caller that doesn't have a precomputed map continue to work exactly as before. The new index also benefits the per-item path — same `text = ?` predicate.

## Test plan

- [x] `:server-app:test --tests "io.tolgee.service.translation.TranslationMemoryServiceBatchTest"` — all 8 tests pass on fresh DB (including a re-run after the migration was added to verify clean Liquibase application)
- [x] `:ee-test:test --tests "io.tolgee.ee.service.translationMemory.TmAutoTranslateProviderEeImplTest"` — all 5 tests pass
- [x] Local benchmark run — numbers above
- [x] Prod EXPLAIN ANALYZE captured for both per-item and batched without the index → confirmed the missing-index root cause
- [ ] CI green
- [ ] Apply the migration in a non-peak window; `CREATE INDEX CONCURRENTLY` on `translation` is online but takes time on a busy table
- [ ] After the migration lands, re-run the prod EXPLAIN ANALYZE to confirm the planner flips join order and execution time drops to single-digit ms
- [ ] Manual smoke on a sizable project: trigger `pretranslate_by_tm` over hundreds of keys, confirm runtime drops and resulting translations are identical to a pre-change baseline run

## Notes for review

- The OSS batched query intentionally keeps the same scope as the existing HQL (cross-project, no `tmIds` / penalty filter). Even on EE, the fallback path remains unrestricted to preserve current behavior — tightening that scope is a separate semantic decision and not part of this perf-only PR.
- The OSS LATERAL query uses `LIMIT 1` per input pair because the self-key exclusion (`k.id <> inputs.req_key_id`) is enforced inside the LATERAL body — no need for the windowed top-K approach that the original plan considered.
- The EE batched query similarly uses `LIMIT 1` per pair via `order by case kind when 'stored' then 0 else 1 end`, mirroring the per-item EE method's preference for stored entries over virtual rows.
- EE's OSS-fallback uses a separate unmetered helper (`ossDelegate.fetchUnmetered`) so the batch metric is recorded exactly once per chunk — calling `getAutoTranslatedValuesForChunk` directly would double-record.

## Summary by CodeRabbit

* **New Features**
  * Added batched Translation Memory (TM) exact-match lookups to precompute results per chunk and feed them into auto-translation.
  * Auto-translation now supports optional reuse of precomputed TM matches.
  * Both OSS and EE TM providers now expose batched lookup paths optimized with native SQL.

* **Metrics**
  * Added new TM batch lookup metrics (duration and hit/miss counts).

* **Database**
  * Added a concurrent hash index on translation text to speed exact-match lookups.

* **Tests**
  * Added new TM batch lookup tests and a disabled benchmark; updated mocks to match expanded TM APIs.

### DIFF
--- a/backend/data/src/main/resources/db/changelog/schema.xml
+++ b/backend/data/src/main/resources/db/changelog/schema.xml
@@ -5775,4 +5775,13 @@
         ]]></sql>
         <rollback/>
     </changeSet>
+    <changeSet author="bdshadow" id="translation-text-hash-idx-for-tm-exact-match" runInTransaction="false">
+        <sql dbms="postgresql">
+            CREATE INDEX CONCURRENTLY IF NOT EXISTS translation_text_hash_idx
+            ON translation USING hash (text);
+        </sql>
+        <rollback>
+            DROP INDEX CONCURRENTLY IF EXISTS translation_text_hash_idx;
+        </rollback>
+    </changeSet>
 </databaseChangeLog>
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/translationMemory/TmAutoTranslateProviderEeImpl.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/translationMemory/TmAutoTranslateProviderEeImpl.kt
@@ -11,6 +11,7 @@ import io.tolgee.service.translationMemory.TranslationMemoryManagementService
 import jakarta.persistence.EntityManager
 import org.springframework.context.annotation.Primary
 import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
 
 /**
  * EE auto-translate resolver. Mirrors the suggestion panel's notion of "matches the project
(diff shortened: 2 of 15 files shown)


bdshadow:
@coderabbitai review, please

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/translationMemory/TmAutoTranslateProviderEeImpl.kt
```diff
@@ -70,6 +71,203 @@ class TmAutoTranslateProviderEeImpl(
...
+  ): Map<Pair<Long, Long>, TranslationMemoryItemView> {
+    // Pass the actual (keyId, targetLangId) pairs as two positionally-aligned arrays.
+    // Postgres `unnest(arr1, arr2)` zips them into rows, so the input set is exactly
+    // O(items) instead of O(uniqueKeys × uniqueLangs).
+    val reqKeyIds = items.map { it.first }.toTypedArray()
+    val reqLangIds = items.map { it.second }.toTypedArray()
+
+    val sql =
+      """
+      with inputs as (
```
dkrizan, line 177: I believe I have already seen this query. I am worried about duplicating same queries :( next time someone wants to rework the logic, they need to make sure the same change is applied to all its twins.

Could we atleast add a test that asserts the batched result matches the per-item result over the same fixture, for both OSS and EE?

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/translationMemory/TmAutoTranslateProviderEeImpl.kt
```diff
@@ -70,6 +71,203 @@ class TmAutoTranslateProviderEeImpl(
...
+    metrics.recordTranslationMemoryLookupBatch(
+      chunkSize = items.size,
+      hitCount = result.size,
+      durationMs = System.currentTimeMillis() - startedAt,
+    )
+    return result
+  }
+
+  /**
+   * Returns EE-aware matches for the given chunk in one SQL round-trip. Pairs without
```
dkrizan, line 154: I would avoid adding noisy comments like this (and in other changes too)

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5604,4 +5604,13 @@
             DROP INDEX CONCURRENTLY IF EXISTS translation_qa_checks_stale_partial_idx;
         </rollback>
     </changeSet>
+    <changeSet author="bdshadow" id="translation-text-hash-idx-for-tm-exact-match" runInTransaction="false">
+        <sql dbms="postgresql">
+            CREATE INDEX CONCURRENTLY IF NOT EXISTS translation_text_hash_idx
```
dkrizan, line 5780: You mention that adding this is a real prod win, but it's not measured... personally, I don't have good experience with blindly believing an index will speed things up in production :D
Can you actually test this? e.g. on a preview instance

bdshadow:
Re-measured this PR's win against a prod-shape DB (11M translations, 2.4M keys, 24k projects, 23k TMs). The local-fixture `TranslationMemoryServiceBenchmark` numbers in the PR description (23-37× at chunk size 10) hold up for the specific scenario it benchmarks but don't translate to real-world workloads. Three measured prod-shapescenarios, each 1 warmup + 3 timed rounds, median wall-clock from `psql \timing`:

| Workload | Per-call | Speedup |
|---|---:|---:|
| Microbenchmark (local, single project, ~4 ms/call) | ~3-4 ms | 23-37× |
| OSS deployment on a 20k-key project | ~87 ms | 1.2× |
| EE Cloud project without shared TMs | ~149 ms | 1.5× |
 | EE Cloud project with assigned shared TMs | ~797 ms | no win (slight regression) |

For comparison, the local fixture has ~3-4 ms per-call cost. That's the scenario where round-trip + Hibernate entity hydration savings dominate and you see 23×.

**Why the gap**

Server-side per-call cost grows ~20-200× from the local fixture to real prod data:
- Local fixture: 20k keys, single project, fresh DB, simple OSS query -> ~3-4 ms/call. Batching's wins (1800 round-trips eliminated, 1800 Hibernate entity loads avoided, planning amortized) dominate the time budget.
- Prod OSS, no shared TMs: ~87 ms/call. Round-trip + planning savings are ~5-15 ms each × 9 = ~100 ms total - modest fraction of 870 ms.
- Prod EE with shared TMs: per-call dominated by the planner's join order in the virtual half (~80k candidate keys walked per pair before the text = ? filter is evaluated). Each call costs ~800 ms regardless of batching. The LATERAL wrapper adds ~500 ms of outer-query setup tax, hence the small regression.

Confirmed via EXPLAIN ANALYZE: same bad plan shape (tm_virt -> tmp_w -> key k filter on text with 80k inner-loop iterations) in both per-item and batched on the shared-TM scenario. I also tried a hash index on translation(text) to flip the planner - PG ignored it across all unforced runs (idx_scan = 0), and forced plans were slower (3.3s with enable_nestloop = off). The hash-index commit was dropped in this PR for that reason.

By now i'm closing this PR as the current state doesn't provide the benefits it was initially created for. It can be used later as a reference of a better implementation.

PR #3705, opened 2026-05-29
fix: stop leaking untranslated glossary terms into AI translations

### LINKED ISSUE #3288: Untranslated glossary terms are breaking AI translations
**Describe the bug**
When a term is not translated in the Glossary, it will be used in its default language and form, as in the glossary, when using AI to translate a string.

**To Reproduce**
1. Go to Glossary
2. Create new term "Test"
3. Do not translate it
4. Go to project
5. Add key "This is a test."
6. Translate it by machine translation to different lanugage.
7. Translation will include word "Test" with capital letter in EN lanugage.

**Expected behavior**
If terms are not yet translated to target language in Glossary, they should be translated as regular words.

**Versions and environment**
- Tolgee Platform version: Public as of this date.
- Browser: Chrome on Mac

Anty0:
## Summary

Closes #3288.

- Glossary terms matched in the source text were always sent to the AI translation prompt, even with no translation in the target language. The prompt told the model to use them, so the source-language form leaked into the output untranslated.
- `GlossaryTermService.getGlossaryTerms` now includes a matched term only when it carries a usable signal for the target language: a target translation, a description, or the non-translatable / forbidden flags. Bare untranslated terms are dropped.
- The glossary prompt fragment gains a per-field legend (`source` / `target` / `description` / `isAbbreviation`) and clarifies that an absent `target` means the term should be translated naturally rather than copied verbatim.
- Adds `GlossaryTermsForAiTranslationTest` covering all include/exclude cases.

## Summary by CodeRabbit

* **New Features**
  * AI translation glossary now omits entries that would yield empty/meaningless target translations while still including terms with descriptions or special flags (non‑translatable, forbidden, case‑sensitive, abbreviation).
  * Glossary prompt fragment expanded with clearer instructions and explicit fields for source, target, description, case sensitivity and abbreviation handling.
  * Added helper to detect glossary terms with no flags.

* **Tests**
  * Added integration tests and seed data validating glossary term selection for AI-assisted en→fr translations.
  * Updated E2E assertion text for glossary option in AI Playground tests.

### DIFF
--- a/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/glossary/GlossaryTermService.kt
+++ b/ee/backend/app/src/main/kotlin/io/tolgee/ee/service/glossary/GlossaryTermService.kt
@@ -14,6 +14,7 @@ import io.tolgee.ee.repository.glossary.GlossaryTermRepository
 import io.tolgee.exceptions.NotFoundException
 import io.tolgee.model.glossary.Glossary
 import io.tolgee.model.glossary.GlossaryTerm
+import io.tolgee.model.glossary.GlossaryTerm.Companion.hasNoFlags
 import io.tolgee.model.glossary.GlossaryTermTranslation
 import io.tolgee.model.glossary.GlossaryTermTranslation.Companion.WORD_REGEX
 import io.tolgee.service.machineTranslation.MtGlossaryTermsProvider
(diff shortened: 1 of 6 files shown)


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/glossary/GlossaryTermService.kt
```diff
@@ -292,13 +292,27 @@ class GlossaryTermService(
...
+        val target = term.translations.find { it.languageTag == targetLanguageTag }?.text
+        val description = term.description
+
+        if (
+          target.isNullOrBlank() &&
+          description.isBlank() &&
+          !term.flagNonTranslatable &&
+          !term.flagForbiddenTerm &&
+          !term.flagCaseSensitive &&
+          !term.flagAbbreviation
```
dkrizan: nitpick: how about adding val in `GlossaryTerm` for grouping condition for these 4 flags
e.x.:
```kt
val noFlags: Boolean
get() = (!flagNonTranslatable && !flagCaseSensitive && !flagAbbreviation && !flagForbiddenTerm)
```

dkrizan (approved):

PR #3715, opened 2026-06-01
fix: Show meaningful error when Content Delivery storage prune fails

### LINKED ISSUE #3339: Handle unsupported prune for S3 buckets better
**Describe the bug**
When a user configures an S3 bucket that doesn't support pruning directories, they only see `unexpected_error_occurred` in the UI when trying to publish to the bucket. We should display a clearer error message so the user can resolve this issue themselves without needing access to the server logs.

**To Reproduce**
1. Create an S3 bucket using a non-official 3S protocol implementation that does not support pruning.
2. Enable pruning for publishing files into that bucket.
3. Try to publish manually from the UI.
4. Observe the `unexpected_error_occurred` error.

**Expected behavior**
The error message should indicate something like "Prune of S3 bucket failed."

1795771535y-cell:
## Problem
When publishing to Content Delivery with 'Prune before publish' enabled and the S3-compatible storage doesn't support directory listing, the server throws a generic \unexpected_error_occurred\ with no hint about the root cause.

## Fix
- Catch prune exceptions in \ContentDeliveryUploader\ and wrap in \BadRequestException\ with new \CONTENT_DELIVERY_PRUNE_FAILED\ error code
- Add user-facing error message suggesting to disable 'Prune before publish'
- Register the error code in the frontend \useErrorTranslation\

Closes #3339

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for content delivery pruning so users see a clear, translated error when pruning fails.
* **Localization**
  * Added a user-facing translation for the prune failure message and updated error translation logic to display it.
* **UX**
  * Error message now suggests disabling “Prune before publishing” in content delivery settings when relevant.


JanCizmar:
Love this! Thanks ❤️

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
```diff
@@ -104,7 +104,7 @@ private const val PROJECT_PERMISSIONS_MAIN = """
 interface UserAccountRepository : JpaRepository<UserAccount, Long> {
   fun findByUsername(username: String?): Optional<UserAccount>
 
-  @Query("from UserAccount ua where ua.username = :username and ua.deletedAt is null and ua.disabledAt is null")
+  @Query("from UserAccount ua where lower(ua.username) = lower(:username) and ua.deletedAt is null and ua.disabledAt is null")
```
Anty0: Quite a big breaking change. As far as I know, we do have conflicting user accounts on production - we probably can't apply this without some additional changes to resolve the conflicts.

backend/data/src/main/kotlin/io/tolgee/component/contentDelivery/ContentDeliveryUploader.kt
```diff
@@ -60,7 +60,14 @@ class ContentDeliveryUploader(
     storage: FileStorage,
   ) {
     if (config.pruneBeforePublish) {
-      storage.pruneDirectory(config.slug)
+      try {
+        storage.pruneDirectory(config.slug)
+      } catch (e: Exception) {
+        throw io.tolgee.exceptions.BadRequestException(
+          io.tolgee.constants.Message.CONTENT_DELIVERY_PRUNE_FAILED,
```
Anty0: Import instead of using FQ names.

Anty0 (changes requested):
The error message for the CD is something we could add. The same can't be said about making the email case-insensitive. That could be quite an undertaking. Any ideas @JanCizmar? Should I make a follow-up PR with only the CD error handling changes?

1795771535y-cell:
Thanks for the review @Anty0 and @coderabbitai! Here's what I changed:

1. ? **Fixed FQ names** ? added import statements for BadRequestException and Message in ContentDeliveryUploader.kt (replaced io.tolgee.exceptions.BadRequestException and io.tolgee.constants.Message with short names)
2. ? **Added defaultValue** to 	('content_delivery_prune_failed') in useErrorTranslation.ts per coderabbitai's suggestion
3. ? **Removed email case-insensitivity changes** - reverted UserAccountRepository.kt, SignUpService.kt, UserAccountService.kt, and ThirdPartyUserHandler.kt back to upstream. Per Anty0's feedback, this is too large a breaking change and should be its own follow-up PR with careful conflict resolution.

The PR now only contains the CD error handling improvements.

1795771535y-cell:
### Review feedback addressed

**@Anty0:** 
1. ? **Email case-insensitive changes** - completely reverted from this PR (the 4 files: UserAccountRepository.kt, UserAccountService.kt, UserPreferencesService.kt, SignUpService.kt are now at upstream versions). The case-insensitive normalization needs a separate discussion/PR as a breaking change.
2. ? **FQ names ? imports** - ContentDeliveryUploader.kt now uses proper imports (io.tolgee.exceptions.BadRequestException, io.tolgee.constants.Message).
3. ? **CD error handling** - try-catch around pruneDirectory with BadRequestException + translation key.

**@coderabbitai:** 
The defaultValue is already correctly passed as the 2nd positional argument: 	('content_delivery_prune_failed', 'Content delivery prune failed'). This follows the Tolgee React API pattern. The full descriptive message is in en.json.

This PR now contains only the CD prune error handling (4 files). Ready for re-review!

Anty0:
Thanks @1795771535y-cell! I'll clean up the translations in a separate PR and merge it.

Anty0:
Superseded by #3720

PR #3716, opened 2026-06-01
fix: make email case-insensitive across login, signup, and invitations

### LINKED ISSUE #1382: Minor: user email is case sensitive
I've locked my account a couple times because I was trying to use `jdimeo@email.org` instead of `JDimeo@email.org`, which apparently was the exact casing I used during registration. Maybe this is a feature rather than a bug for security reasons?

dkrizan:
## Summary

Emails are treated case-insensitively by virtually all providers, but Tolgee matched them case-sensitively — a user registered as `AB@email.org` could not log in as `ab@email.org`.

This normalizes usernames (emails) to lowercase on every write, makes lookups case-insensitive, enforces case-insensitive uniqueness at the DB level, and migrates existing data (normalize + de-duplicate) so the change is safe on Cloud and self-hosted.

Closes #1382.

> ⚠️ Overlaps with #3461 (disabled-account handling) — same files. See "Coordination with #3461" below.

## Approach

- **Normalize on write:** the email is lowercased on signup, email change (incl. the verification flow), OAuth/SSO, and initial-user creation, so stored usernames are always lowercase.
- **Case-insensitive lookups:** `findActive` uses `lower(ua.username) = lower(:username)` — covers login, password reset, OAuth matching, and duplicate checks, and stays correct for any non-lowercased rows (test/dev/legacy data).
- **Hard DB guarantee:** a `unique (lower(username)) where deleted_at is null` index replaces the old case-sensitive `(username)` one. Scoped to `deleted_at is null` (so it also covers disabled accounts → a disabled account keeps reserving its email), and `findActive` is now index-backed.
- **Invitation duplicate-checks** use `lower()` on both `i.email` and `u.username`.

## Code changes

| File | Change |
|------|--------|
| `SignUpService` / `UserAccountService.updateUserEmail` / `ThirdPartyUserHandler` / `EmailVerificationService` / `InitialUserCreatorCommandLineRunner` | Lowercase the email on every write path |
| `UserAccountRepository.findActive` | `lower(ua.username) = lower(:username)` (case-insensitive, robust to non-lowercased rows) |
| `InvitationRepository` | Case-insensitive duplicate-invite detection |

## Migration (`schema.xml`) — 3-step swap, no uniqueness gap

Runs automatically on Cloud and every self-hosted upgrade; deterministic and non-destructive. A unique index is live at every committed point (the new one is created before the old one is dropped):

1. **Dedup (data, transactional):** for each group of non-deleted accounts (`deleted_at IS NULL`, `username NOT NULL`) colliding on `lower(username)`, keep a survivor (an **active** account if present, then most-recent activity, tie-broken by oldest `created_at` then lowest id); **retire the losers** by setting `disabled_at` **and** renaming `username` to a unique `disabled-<n>-<original>`. Then lowercase the surviving active usernames. (Disabled, not deleted → self-hosted admins can re-enable; renamed → frees the email so there's exactly one non-deleted row per email.) NULL usernames excluded; no-op rollback.
2. **Create** `unique (lower(username)) where deleted_at is null` `CONCURRENTLY`, while the old index still protects writes.
3. **Drop** the old case-sensitive `(username)` index (auto-named, located by definition); no-op rollback.

### On Tolgee Cloud (49 affected groups)
The migration makes Cloud correct automatically. For the few groups where the same person used **both** accounts heavily, a manual merge of the disabled account's data into the survivor can be done as a follow-up (not automated, since per-account merge decisions can't be made blindly across self-hosted instances).

## Coordination with #3461

#3461 ("improve handling of disabled accounts") touches the same files and adds `findActiveOrDisabled`. To merge cleanly:
- `findActiveOrDisabled` must be **case-folded** (`lower(ua.username) = lower(:username)`), consistent with the lowercase model.
- Because losers here are **renamed** (not left sharing the survivor's email), there is exactly one non-deleted row per email — so `findActiveOrDisabled` stays single-result (no `NonUniqueResultException`).

## Tests

- `AuthTest` — token generated regardless of email casing (case-insensitive login)
- `PublicControllerTest` — username stored lowercased on signup; signup with the same email in different casing is rejected
- `V2UserControllerTest` — email stored lowercased on update
- `EmailVerificationTest` — verified new email stored lowercased
- `OrganizationControllerInvitingTest` — case-insensitive duplicate-invite detection
- `UserAccountUsernameUniqueIndexTest` — the unique index rejects active+active and active+disabled case-variants, and allows a deleted case-variant

All passing against a freshly-migrated database.

## Summary by CodeRabbit

* **Bug Fixes**
  * Email/username handling is now case-insensitive: sign-up, login, email updates, invitations, and auth work correctly regardless of email casing.
  * Sign-up and invitations reject duplicates that differ only by letter casing.

* **Chores**
  * Database migration enforces case-insensitive uniqueness of active usernames and normalizes stored usernames to lowercase.

* **Tests**
  * Added coverage ensuring casing behavior is consistent across flows.


backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5604,4 +5604,115 @@
...
+                AND tablename = 'user_account'
+                AND indexdef ILIKE '%UNIQUE%'
+                AND indexdef ILIKE '%(username)%'
+                AND indexdef ILIKE '%deleted_at IS NULL%'
+                AND indexdef NOT ILIKE '%lower(%';
+              IF idx IS NOT NULL THEN
+                EXECUTE 'DROP INDEX ' || quote_ident(idx);
+              END IF;
+            END $$;
+        ]]></sql>
```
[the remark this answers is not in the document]
dkrizan: Good catch. Reworked the migration into 3 ordered changesets so a unique index is always live: disable duplicate losers (old index still protecting) → build the new `lower(username)` index `CONCURRENTLY` → drop the old index and lowercase survivors last. No committed window without DB-level uniqueness now. Fixed in 33dc49ebb.

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5604,4 +5604,115 @@
...
+                       PARTITION BY c.lemail
+                       ORDER BY a.ts DESC NULLS LAST, c.created_at ASC, c.id ASC
+                     ) AS rn
+              FROM cand c
+              LEFT JOIN last_activity a ON a.author_id = c.id
+            )
+            UPDATE user_account ua
+            SET disabled_at = now()
+            FROM ranked r
+            WHERE ua.id = r.id AND r.rn > 1;
```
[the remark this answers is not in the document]
dkrizan: Done — added `username IS NOT NULL` to the dedup CTEs so NULL usernames aren't grouped together and wrongly disabled. Fixed in 33dc49ebb.

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5604,4 +5604,115 @@
...
+
+        <rollback>
+            <!-- Schema-only rollback: restores the old case-sensitive partial unique index.
+                 The data changes (disabled duplicates, lowercased usernames) are not reverted -
+                 the original casing and which accounts were disabled are not recorded. -->
+            <sql dbms="postgresql"><![CDATA[
+                CREATE UNIQUE INDEX IF NOT EXISTS user_account_username_idx
+                  ON public.user_account (username)
+                  WHERE deleted_at IS NULL;
+            ]]></sql>
```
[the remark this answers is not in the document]
dkrizan: Agreed — recreating the case-sensitive index could fail against the normalized data (active + disabled rows sharing the same lowercased username). Switched the data changesets to a no-op rollback, since the migration isn't cleanly reversible anyway (original casing / disabled set aren't recorded). Fixed in 33dc49ebb.

backend/data/src/main/resources/db/changelog/schema.xml
```diff
@@ -5604,4 +5604,122 @@
...
+
+    <changeSet author="dkrizan" id="user-account-lower-username-unique-index" runInTransaction="false">
+        <comment>
+            Replace the case-sensitive username uniqueness with a case-insensitive one. Built
+            CONCURRENTLY here, while the old case-sensitive index is still live, so writes are
+            protected throughout (the old index is dropped in the next changeSet, after this one
+            succeeds - there is never a window without a unique index).
+        </comment>
+        <sql dbms="postgresql">
+            CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS user_account_lower_username_uidx
```
Anty0, line 5733: **P1 — migration can end with NO username uniqueness enforced.** `CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS`: if the CONCURRENTLY build is interrupted (lock timeout, duplicate hit mid-build, killed deploy), Postgres leaves an **INVALID** index of this name. On Liquibase retry, `IF NOT EXISTS` matches it by name, skips re-creation, records the changeSet as applied — then the next changeSet drops the old, still-valid case-sensitive index. Net: no enforcing unique index, silently, which breaks this changeSet's own "never a window without a unique index" claim.

Fix: before create, drop any `pg_index.indisvalid = false` index of this name (a `DO` block); and gate the drop-old-index changeSet on the new one existing AND `indisvalid = true`.

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
```diff
@@ -104,13 +104,16 @@ private const val PROJECT_PERMISSIONS_MAIN = """
 interface UserAccountRepository : JpaRepository<UserAccount, Long> {
   fun findByUsername(username: String?): Optional<UserAccount>
 
-  @Query("from UserAccount ua where ua.username = :username and ua.deletedAt is null and ua.disabledAt is null")
+  @Query(
+    "from UserAccount ua where lower(ua.username) = lower(:username) " +
```
Anty0, line 109: **P1 — JVM `.lowercase()` vs Postgres `lower()` diverge for non-ASCII → login lockout on the default self-hosted image.** Writes fold via Kotlin `String.lowercase()` (Locale-invariant Unicode); this query, the unique index, and the dedup all fold via SQL `lower()`, which is LC_CTYPE-dependent and ASCII-only under musl/C. The shipped self-hosted DB is `postgres:13.23-alpine` (musl); tests run on glibc `postgres:16.3`, so this is invisible in CI. `@Email`/`EmailValidator` accepts Unicode local parts + IDN, so non-ASCII usernames are reachable.

Concrete repro (musl/C cluster): sign up `ΛDMIN@example.com` → stored `λdmin@example.com`; login `ΛDMIN@example.com` → `lower('λdmin…') = lower('ΛDMIN…')` leaves `Λ` untouched → **401**, user must manually type the lowercased form. (`UserCredentialsService.checkUserCredentials` passes the raw input straight into `findActiveOrDisabled`.)

Fix: make one side authoritative. Simplest — normalize the lookup INPUT in Kotlin too and compare against the plain stored column (`ua.username = :normalizedUsername`, index on plain `username`), so DB collation drops out; or restrict usernames to ASCII / NFKC-fold before persist. Add a non-ASCII case-variant test.

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
```diff
@@ -104,13 +104,16 @@ private const val PROJECT_PERMISSIONS_MAIN = """
 interface UserAccountRepository : JpaRepository<UserAccount, Long> {
   fun findByUsername(username: String?): Optional<UserAccount>
```
Anty0: **P3 — `findByUsername` left case-sensitive; latent footgun.** It is the only username lookup still doing an exact match; `findActive`/`findActiveOrDisabled` were converted to `lower()`-based. No production callers remain (only `AllOrganizationOwnerJobTest`, `DbPopulatorTest`). With usernames now expected lowercase everywhere, a future caller could reach for `findByUsername` expecting the same case-insensitive semantics as its siblings and silently get a case-sensitive miss.

Fix: remove it (switch the two tests to `findActive`/`findActiveOrDisabled`), make it `lower()`-based, or rename to `findByExactUsername` to make the semantics explicit.

backend/data/src/main/kotlin/io/tolgee/service/security/SignUpService.kt
```diff
@@ -64,6 +64,6 @@ class SignUpService(
 
   fun dtoToEntity(request: SignUpDto): UserAccount {
     val encodedPassword = passwordEncoder.encode(request.password!!)
-    return UserAccount(name = request.name, username = request.email, password = encodedPassword)
+    return UserAccount(name = request.name, username = request.email.lowercase(), password = encodedPassword)
```
Anty0, line 62: **P3 — normalization invariant duplicated across 5 write sites, and a 6th already omits it.** "Stored username = lowercased email" is an inline `.lowercase()` here, in `ThirdPartyUserHandler.createUserEntity`, `EmailVerificationService.setNewEmailIfChanged`, `UserAccountService.updateUserEmail`, and `InitialUserCreatorCommandLineRunner`. There is no shared normalizer. `UserAccountService.createDemoUser` (UserAccountService.kt:151, outside this diff) builds a `UserAccount` and saves it with **no** `.lowercase()` — harmless only because demo emails are hardcoded lowercase, but it is exactly the failure mode this scattering invites, and the read-side `lower()` would mask the symptom.

Fix: enforce at one chokepoint — normalize in the `UserAccount.username` setter, or a shared `normalizeUsername()` — and route all sites (incl. `createDemoUser`) through it; drop the five inline calls.

backend/data/src/main/kotlin/io/tolgee/repository/InvitationRepository.kt
```diff
@@ -36,8 +36,8 @@ interface InvitationRepository : JpaRepository<Invitation?, Long?> {
     select count(p) from Permission p
     left join p.invitation i
     left join p.user u
-    where 
-        (i.email = :email or u.username = :email) and
+    where
+        (lower(i.email) = lower(:email) or lower(u.username) = lower(:email)) and
```
Anty0, line 40: **P3 — inconsistent normalization strategy for the same concept.** Usernames are normalized on write; invitation emails are stored verbatim (`InvitationService.getInvitationInstance` sets `invitation.email = params.email`) and case-folded only at read time (these `lower()` queries + `ignoreCase = true` validation). Two opposite conventions side by side: a future reader can't tell from the model which email fields are guaranteed lowercase, and a future query against `invitation.email` that forgets `lower()` silently misses case-variant rows.

Fix: pick one. Simplest is to also lowercase the invitation email on write (via the same shared normalizer), making the read-side `lower()` belt-and-suspenders rather than the sole guarantee.

backend/data/src/main/kotlin/io/tolgee/security/thirdParty/ThirdPartyUserHandler.kt
```diff
@@ -98,7 +98,7 @@ class ThirdPartyUserHandler(
 
   private fun createUserEntity(data: ThirdPartyUserDetails): UserAccount {
     val user = UserAccount()
-    user.username = data.username
+    user.username = data.username.lowercase()
```
Anty0, line 101: **P2 — OAuth/SSO lowercasing has no test.** This `.lowercase()` is one of the PR's five enumerated write paths and the only one with zero casing coverage. Existing OAuth tests (`AuthTest` github/google/oauth2) all use already-lowercase provider emails, so they'd still pass if this call were deleted. Nothing asserts that a mixed-case provider email is stored lowercased, nor that a follow-up OAuth login with differently-cased provider email resolves to the same account (`createUser` looks up via `findActiveOrDisabled(data.username)`).

Fix: add a test driving OAuth login with a mixed-case provider email (via the Git/Google/OAuth2 auth utils) asserting the stored username is lowercased, plus a second asserting case-variant re-login matches the same account.

backend/data/src/main/kotlin/io/tolgee/service/security/UserAccountService.kt
```diff
@@ -572,19 +572,23 @@ class UserAccountService(
...
+      return
+    }
 
-      this.findActive(dto.email)?.let { throw ValidationException(Message.USERNAME_ALREADY_EXISTS) }
-      if (tolgeeProperties.authentication.needsEmailVerification) {
-        emailVerificationService.resendEmailVerification(userAccount, request, dto.callbackUrl, dto.email)
-      } else {
-        userAccount.username = dto.email
-      }
+    if (!emailValidator.isValid(dto.email, null)) {
```
Anty0: **P3 (cleanup) — validate the normalized value.** This block introduces `val newEmail = dto.email.lowercase()` and uses it for the early-return, the `findActive` lookup, the verification call, and the final assignment — but the validity check still passes raw `dto.email`. Lowercasing can't change validity, so it's not a bug, but mixing `dto.email` and `newEmail` in the same short block makes a reader pause. Use `emailValidator.isValid(newEmail, null)` for consistency.

backend/app/src/test/kotlin/io/tolgee/repository/UserAccountUsernameUniqueIndexTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+@SpringBootTest
+class UserAccountUsernameUniqueIndexTest : AbstractSpringTest() {
+  @Autowired
+  lateinit var userAccountRepository: UserAccountRepository
+
+  @Test
+  @Transactional
+  fun `rejects a second active account colliding on lower(username)`() {
+    userAccountRepository.saveAndFlush(UserAccount(name = "a", username = "ciunique@test.com", password = "x"))
```
Anty0: **P3 — direct JPA-entity instantiation violates `.claude/rules/test-data-conventions.md`.** All three methods build `UserAccount(...)` inline and persist via `saveAndFlush`, which the rule forbids ("Never create test entities directly… Always use a dedicated *TestData class"). This is the only backend test file doing so (a grep for `UserAccount(name` matches only this file).

backend/app/src/test/kotlin/io/tolgee/repository/UserAccountUsernameUniqueIndexTest.kt
```diff
@@ -0,0 +1,58 @@
...
+      userAccountRepository.saveAndFlush(UserAccount(name = "b", username = "CiUnique@test.com", password = "x"))
+    }.isInstanceOf(DataIntegrityViolationException::class.java)
+  }
+
+  @Test
+  @Transactional
+  fun `rejects a disabled account colliding with an active one - disabled still reserves the email`() {
+    userAccountRepository.saveAndFlush(UserAccount(name = "a", username = "ciunique2@test.com", password = "x"))
+
+    // the index is scoped to deleted_at IS NULL (includes disabled), so a disabled case-variant is rejected
```
Anty0: **P3 (nit) — comments restate behavior already in the names + describe remote state.** This comment and the one at line 49 narrate what the (very descriptive) test names already state, and the mechanism they cite (`index scoped to deleted_at IS NULL` / `deleted rows excluded`) is the partial-index definition that lives in schema.xml — it will rot silently if that predicate changes. Per `.claude/rules/comments.md`, drop both, or replace with a single pointer to the `user_account_lower_username_uidx` changeSet rather than restating its WHERE clause.

Anty0 (dismissed):
While reviewing the change set, I kept wondering whether there are any issues with lowercasing, since it uses both Kotlin's `lowercase` and PostgreSQL's `lower`. The review notes below are the result of the CC review workflow I've been testing. I've flagged the casing issue but otherwise left it for a full review. I've only left comments that I think are worth considering, but we don't have to fix everything noted there.

Severity legend: **P1** correctness/data-integrity, **P2** destructive-migration & test coverage, **P3** maintainability/conventions.

backend/data/src/main/resources/db/changelog/schema.xml
dkrizan: Applied in 77ab7d024.

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
dkrizan: Deferring (not blocking). The trigger is a non-ASCII **uppercase** email local part on a musl DB — vanishingly rare, and not a regression (ASCII case-insensitivity, the #1382 ask, works; non-ASCII just isn't folded on musl). Every fix has a tradeoff (Kotlin-folding the lookup input reintroduces a "callers must pre-fold" contract; restricting usernames to ASCII is a product change). Tracking as a follow-up.

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
dkrizan: Applied in 011f0967e.

backend/data/src/main/kotlin/io/tolgee/service/security/SignUpService.kt
dkrizan: Deferring. `createDemoUser` is harmless (demo emails are hardcoded lowercase), and a forgotten `.lowercase()` elsewhere is masked by the read-side `lower()` + the unique index. The cleanest chokepoint (the `username` setter) would lowercase test fixtures on persist and break tests, so it's not a clear win.

backend/data/src/main/kotlin/io/tolgee/repository/InvitationRepository.kt
dkrizan: Deferring — low impact. Read-side `lower()` already matches case-variant invitation emails and invitations are ephemeral; pure consistency, can lowercase on write in a follow-up.

backend/data/src/main/kotlin/io/tolgee/security/thirdParty/ThirdPartyUserHandler.kt
dkrizan: Applied in 270b74776.

backend/data/src/main/kotlin/io/tolgee/service/security/UserAccountService.kt
dkrizan: Applied in 011f0967e.

backend/app/src/test/kotlin/io/tolgee/repository/UserAccountUsernameUniqueIndexTest.kt
dkrizan: Keeping as-is — deliberate exception: this is a low-level test of the DB unique index itself, where building the colliding rows directly via the repository is the point; a *TestData graph would obscure the assertion.

backend/app/src/test/kotlin/io/tolgee/repository/UserAccountUsernameUniqueIndexTest.kt
dkrizan: Applied in 011f0967e.

backend/app/src/test/kotlin/io/tolgee/repository/UserAccountUsernameUniqueIndexTest.kt
dkrizan: On reflection you're right, applied in 6e9c99d49.

PR #3718, opened 2026-06-02
feat: support Python brace-format placeholders in PO

### LINKED ISSUE #3589: Support .PO Brace Formatting
EDIT: This may actually be a separate bug/issue related to https://github.com/tolgee/tolgee-platform/issues/2375 - it seems that the incorrect formatting listed below (when importing with `format: PO_ICU`) only occurs with plurals, but works correctly with single text values

Currently, exporting to .po format with brace formatting (i.e. `python-brace-format`) is not possible. Brace formatting is best practice when translating in python due to allowing parameter re-ordering, which is often times required in languages with different sentence ordering.

For example, given the imported .po file:
```
#, python-brace-format
msgid "1 item"
msgid_plural "{n} items"
msgstr[0] "1 item"
msgstr[1] "{n} items"
```

the subsequent export will result in a .po file with the following:
```
msgid "1 item"
msgid_plural "{n} items"
msgstr[0] "1 item"
msgstr[1] "'{'n'}' items"
```

Proper support for the python flavor of .po gettext, as described in [PEP 3103](https://peps.python.org/pep-3101/).

writing a conversion layer to convert our .po files to ICU and vice-versa in our application code. we'll likely do this anyway while this feature request is pending :(

Anty0:
Closes #3589

## Summary
- Add `PO_PYTHON_BRACE` import format (gettext `python-brace-format` flag) and `PYTHON_BRACE` export message format for Python `str.format` style placeholders (`{name}`, `{0}`, `{x:.2f}`, …), mirroring the existing `PO_PYTHON` / `PYTHON_PERCENT` pair.
- Import↔export are symmetric for every supported placeholder: `{x:d}` ↔ `{x, number}`, `{x:.Nf}` ↔ `{x, number, .0..0}`, `{x:f}` ↔ precision 6, `{x:e}` ↔ scientific, and the plural `{count:d}` count ↔ ICU `#` ↔ `{0:d}`.
- Unsupported specs (`{x!r}`, `{x:>10}`, attribute/index access) and literal braces (`{{`, `}}`) degrade to escaped ICU literals and remain byte-stable on every subsequent import → export. Literal `#` inside plurals is ICU-escaped so it survives export verbatim.
- Brace heuristic detection weighted below ICU so plain `{name}` files without the flag still resolve to `PO_ICU` (no regression to existing detection).
- Frontend: new option in the PO export-format group, dropdown label, and e2e map.

## Release Notes

* **New Features**
  * Added Python Brace Format support for exporting translations
  * Added Python Brace Format support for importing PO files containing brace placeholders
  * Extended available message format options for export configuration

* **Tests**
  * Added test coverage for Python Brace Format detection, conversion, and round-trip operations

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/paramConvertors/in/PythonBraceToIcuPlaceholderConvertor.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/paramConvertors/in/PythonBraceToIcuPlaceholderConvertor.kt
@@ -1,0 +1,104 @@
+package io.tolgee.formats.paramConvertors.`in`
+
+import io.tolgee.formats.ToIcuPlaceholderConvertor
+import io.tolgee.formats.escapeIcu
+
+class PythonBraceToIcuPlaceholderConvertor : ToIcuPlaceholderConvertor {
+  private var autoIndex = 0
+  private var placeholderCount = 0
+
+  override val pluralArgName: String = "0"
+
+  override val regex: Regex
+    get() = PYTHON_BRACE_PLACEHOLDER_REGEX
+
+  override fun convert(
+    matchResult: MatchResult,
+    isInPlural: Boolean,
+  ): String {
+    val value = matchResult.value
+    if (value == "{{") return "{".escapeIcu(isInPlural)
+    if (value == "}}") return "}".escapeIcu(isInPlural)
+
+    val field = matchResult.groups[GROUP_FIELD]?.value ?: ""
+    val name = resolveName(field) ?: return value.escapeIcu(isInPlural)
+    placeholderCount++
+
+    val conversion = matchResult.groups[GROUP_CONVERSION]?.value
+    if (conversion != null) {
+      // not supported
+      return value.escapeIcu(isInPlural)
+    }
+
+    val formatSpec = matchResult.groups[GROUP_FORMAT_SPEC]?.value
+    if (formatSpec == null) {
+      return "{$name}"
+    }
+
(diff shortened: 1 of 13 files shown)


backend/data/src/main/kotlin/io/tolgee/formats/paramConvertors/in/PythonBraceToIcuPlaceholderConvertor.kt
dkrizan, line 1: Small thing on the plural # logic: it assumes the count is the first placeholder, but anything before it (like a name) throws it off. So `{name} has {count:d} dogs` comes out as `{name} has {count,  number}` dogs instead of using `#`. 

What do you think? 🤔

Anty0: Hmm, I was looking at it a bit. It's a potentially interesting issue. We already have the same behavior for other formats like PO, so changing it only here doesn't really make sense. It also doesn't break the import/export round trip. But it can be pretty annoying when importing multiple languages - when different locales have different placeholder order, different translations could be converted differently.

Probably the only "proper" way to fix this would be to disable conversion to `#` - import for this format would never generate a `#` placeholder (or keep it enabled only when there is a single placeholder).

I'd leave it be for now, since other formats already have the same behavior. Or do you think it's worth fixing?

dkrizan: If this issue is generic, not related to this PR, I agree - leave it.

dkrizan (approved):

PR #3723, opened 2026-06-04
feat: pass key context to DeepL machine translation

dkrizan:
## TL;DR

DeepL's API has a free, untranslated [`context` parameter](https://www.deepl.com/en/blog/deepl-api-context-parameter) that disambiguates short strings. We now feed it from data Tolgee already has — the **key description** + **source texts of neighbouring keys** (big meta). Backend-only, opt-in per provider (`supportsContext`), context is part of the MT cache key, price unchanged. Verified live: "bank" (EN→ES) → `banco` normally, but `orilla` once a river-related description **or** river-related neighbour keys are present.

## What

DeepL's API accepts an optional [`context` parameter](https://www.deepl.com/en/blog/deepl-api-context-parameter): surrounding text that is **not translated and not billed**, but helps the model disambiguate short or ambiguous source strings (UI labels, single words, etc.).

Until now Tolgee sent DeepL only the raw text + languages + formality. This PR builds a context payload from the **key description** and the **source texts of neighbouring keys** (big meta) and passes it to DeepL.

## How

- `DeeplApiService.translate(...)` gains an optional `context` form field (added only when non-blank).
- `MtValueProvider` gets a `supportsContext` capability flag (default `false`, `true` for DeepL) so the extra work is only done for providers that actually use it.
- `MtBatchTranslator` delegates context assembly to `MetadataProvider.getContext()`, co-located with the existing `getCloseItems()` close-keys logic (reuses big meta rather than duplicating it). Building context is best-effort — a `DataAccessException` while fetching neighbours yields no context rather than failing the translation.
- `context` is threaded `TranslationParams` → `ProviderTranslateParams` and included in the MT **cache key**, so cached results invalidate when the context changes.
- Price is unchanged: DeepL does not bill context characters.

Only DeepL benefits today (Google/AWS/Baidu have no equivalent parameter); the LLM provider already had its own richer context path, which is untouched.

## Testing

- Unit test asserts the `context` value reaches `DeeplApiService` and that DeepL declares `supportsContext`.
- Full `TranslationSuggestionControllerMtTest` suite passes.
- Manually verified end-to-end against the live DeepL API (same source word "bank", EN→ES):
  - **key description** path — no context → `banco`; description "the land alongside a river" → `orilla`.
  - **big-meta neighbours** path (no description) — surrounding the key with river-themed neighbour keys → `orilla`; money-themed neighbours → `banco`.

The quality-difference check is manual by nature (non-deterministic, needs a live key); the deterministic guarantee is the unit test on the request payload.

## Summary by CodeRabbit

* **New Features**
  * Added optional context support for machine translation. Context is derived from key descriptions and nearby translations and is passed only to providers that declare support (improves translation relevance). Cache keys now account for context.
* **Tests**
  * Added provider tests for context support and forwarding; updated mocks to cover nullable context handling.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MetadataProvider.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MetadataProvider.kt
@@ -5,10 +5,14 @@ import io.tolgee.dtos.cacheable.LanguageDto
 import io.tolgee.service.bigMeta.BigMetaService
 import io.tolgee.service.translation.TranslationMemoryService
 import jakarta.persistence.EntityManager
+import org.slf4j.LoggerFactory
+import org.springframework.dao.DataAccessException
 
 class MetadataProvider(
   private val context: MtTranslatorContext,
 ) {
+  private val log = LoggerFactory.getLogger(MetadataProvider::class.java)
+
   fun getCloseItems(
     sourceLanguage: LanguageDto,
     targetLanguage: LanguageDto,
@@ -37,6 +41,31 @@ class MetadataProvider(
       .setParameter("sourceLanguageId", sourceLanguage.id)
       .setParameter("closeKeyIds", closeKeyIds)
       .resultList
+  }
+
+  /**
+   * Builds a plain-text context payload for providers that accept one
+   * Best-effort: context only improves the result, so a database failure while fetching the
+   * neighbouring keys yields no context rather than failing the translation itself.
+   */
+  fun getContext(
+    sourceLanguage: LanguageDto,
+    targetLanguage: LanguageDto,
+    metadataKey: MetadataKey,
+    keyDescription: String?,
+  ): String? {
+    try {
+      val parts = mutableListOf<String>()
+      keyDescription?.takeIf { it.isNotBlank() }?.let { parts.add(it) }
+      getCloseItems(sourceLanguage, targetLanguage, metadataKey)
+        .map { it.source }
+        .filter { it.isNotBlank() }
+        .forEach { parts.add(it) }
+      return parts.joinToString("\n").ifBlank { null }
+    } catch (e: DataAccessException) {
+      log.warn("Failed to build MT context for key ${metadataKey.keyId}", e)
+      return null
+    }
   }
 
   /**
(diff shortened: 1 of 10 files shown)


backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MetadataProvider.kt
```diff
@@ -39,6 +43,31 @@ class MetadataProvider(
...
+    keyDescription: String?,
+  ): String? {
+    try {
+      val parts = mutableListOf<String>()
+      keyDescription?.takeIf { it.isNotBlank() }?.let { parts.add(it) }
+      getCloseItems(sourceLanguage, targetLanguage, metadataKey)
+        .map { it.source }
+        .filter { it.isNotBlank() }
+        .forEach { parts.add(it) }
+      return parts.joinToString("\n").ifBlank { null }
```
JanCizmar, line 64: This seem like it may confuse the translator actually. We don't say what's the key description and what's the close items. This would need some testing, otherwise I am afraid that it might actually produce worse results. 🤷‍♂️

backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MtBatchTranslator.kt
```diff
@@ -176,6 +176,8 @@ class MtBatchTranslator(
         item.promptId,
       )
 
+    val providerContext = buildContext(item, baseTranslationText, provider.supportsContext)
```
JanCizmar: Do we need to pass the boolean to the method? We can simply decide here, whether it's null or we need the context, right?

JanCizmar (changes requested):

backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MetadataProvider.kt
dkrizan: I don't think so, as mentioned in docs - "The context parameter is not designed for LLM-style instructions", adding any labels could **actually confuse** the Deepl translator.

backend/data/src/main/kotlin/io/tolgee/service/machineTranslation/MtBatchTranslator.kt
dkrizan: correct, addressed

JanCizmar (approved):

PR #3731, opened 2026-06-08
fix: persist plural arg name on key import

### LINKED ISSUE #2651: Editing a translation, changes the variable name.
**Describe the bug**
I have a base text with variable name 'count'.
 `User has {count, plural, =0 {no points} one {one point} other {{count} points}}`

If I try to edit it, the UI changes the variable name to 'value', and saves it as following, breaking the translation.
 `User has {value, plural, =0 {no points} one {one point} other {{count} points}}`

**Screenshots**

Screenshot 2024-10-30 at 21-25-04 Translations Tolgee

Screenshot 2024-10-30 at 21-25-20 Translations Tolgee

**Versions and environment**
- Tolgee Platform version: 3.80.6

Anty0:
## Summary

- Fixes #2651: editing a synced plural translation rewrote the plural variable (e.g. `{count, ...}` → `{value, ...}`), breaking interpolation.
- **Root cause:** the auto-detecting import paths (`KeysImporter`, `ResolvingKeyImporter` — used by `tolgee sync` and single-step-resolvable import) set `isPlural=true` but discarded the detected `pluralArgName`, storing it NULL. The webapp editor then fell back to the default `value` argument name and rewrote the outer variable.
- **Backend:** both importers now persist the detected argument name; the existing-already-plural branch backfills a missing one and keeps a set name stable.
- **Frontend:** the editor now prefers the stored argument name, then the name parsed from the ICU string, before the `value` default — which also heals keys already stored with a NULL `pluralArgName`.
- **Tests:** extended the backend import pluralization tests to assert the persisted `pluralArgName`, with added re-import and single-step coverage. Also adds a webapp unit test for the plural argument-name resolution (`utils.test.ts`); it starts running once the test-runner PR lands.
- **Note:** the webapp unit-test runner setup (vitest + jsdom, the `frontend-test` CI job, the jest/CRA-stack cleanup, and the stale `qaUtils` assertion fixes) has been split out into #3739 to keep this PR focused on the #2651 fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plural translation handling during imports to properly preserve and populate plural argument names ("count" by default).
  * Fixed plural argument name backfilling when re-importing existing plural keys.
  * Enhanced plural parameter resolution in the translation editor for better consistency.

* **Tests**
  * Added comprehensive test coverage for plural translation import scenarios and plural parameter resolution logic.


.github/workflows/test.yml
```diff
@@ -295,6 +295,29 @@ jobs:
           npm run build
         working-directory: ./webapp
 
+  frontend-test:
```
dkrizan: I guess adding frontend tests workflow should be a seaparate PR, what do you think ?

dkrizan: I guess adding frontend tests workflow should be a seaparate PR, what do you think ?

Anty0: I was considering splitting it up, but with how much we talk recently about optimizing delivery time, it felt like lost time doing that. But sure, I'll split it off. ^^

Anty0:
Done, split off into: #3739

dkrizan (approved):

PR #3738, opened 2026-06-09
feat: add QA Checks and Translation Memory announcement banner

Anty0:
## Summary

- Add new `FEATURE_QA_CHECKS_AND_TRANSLATION_MEMORY` announcement (expires 2026-08-01), appended as the latest entry so it is the one served
- Since the backend serves only the single latest announcement, this combined banner promotes both QA Checks and Translation Memory management, so the active TM banner is not dropped when announcing QA Checks
- Banner links to QA Checks and Translation Memory docs via the shared `DOCS_LINKS` map; also migrates the existing TM case to the same constant (single source of truth)
- Regenerate frontend API schema for the new enum value
- Translation key `announcement_qa_checks_and_translation_memory` created in the Tolgee project

## Release Notes

* **New Features**
  * Added support for displaying an announcement about QA checks and translation memory features, with activation scheduled for August 2026.

* **Chores**
  * Centralized documentation link management across the application for improved maintainability and consistency of help resources.


backend/data/src/main/kotlin/io/tolgee/model/enums/announcement/Announcement.kt
```diff
@@ -23,6 +23,7 @@ enum class Announcement(
   FEATURE_IMPROVED_FIGMA_ANDROID_AND_IOS(parseTime("2025-10-01 00:00 UTC")),
   FEATURE_BRANCHING(parseTime("2026-05-01 00:00 UTC")),
   FEATURE_TRANSLATION_MEMORY_MANAGEMENT(parseTime("2026-07-01 00:00 UTC")),
```
dkrizan, line 25: I’m not sure how it works when there are two overlapping announcements, but we can probably change the date to “today” for this one (`FEATURE_TRANSLATION_MEMORY_MANAGEMENT`).

Anty0: Only the latest one is taken into account; the rest are ignored. No need to update the date of the previous announcement. ^^

dkrizan (approved):

PR #3759, opened 2026-06-17
feat: allow organization owners to disable and re-enable managed users (SSO users)

### LINKED ISSUE #3275: Add support for re-enabling disabled managed users by organization owners
Following the merge of PR #3263, we currently support disabling managed users. However, the system does not yet provide a clear way for organization owners to re-enable disabled users.

Solution:
- Add dedicated endpoints to disable and enable managed accounts by organization owners (instead of convoluting the leave endpoint as was done as a hot fix in the last PR – revert that)
- Keep disabled users visible in the organization member list but visually distinguish them (e.g., gray them out).
- Add UI controls to enable/disable users for organization owners if the organization manages them.
- Add tests for disabling and re-enabling managed users.

Anty0:
Closes #3275. Companion billing PR: tolgee/billing#305 (required — this branch adds an event that one consumes, and changes a signature it calls).

TLDR: This allows owners of SSO-enabled organizations to enable or disable their users. Because SSO accounts can't leave the organization that manages them, this is the only way to "get rid" of existing users.

The current implementation disables the managed account when the organization owner attempts to remove it from the organization. This behavior wasn't transparent to users, and it didn't allow the organization owner to re-enable accounts disabled this way.

## Design

Remove user (existing option):
Screenshot 2026-06-17 at 16 07 50

Disable user:
Screenshot 2026-06-17 at 18 42 01

Re-enable user:
Screenshot 2026-06-17 at 16 51 54

## Summary

- Add owner-only, super-auth `PUT /v2/organizations/{organizationId}/users/{userId}/disable` and `/enable` endpoints to disable and re-enable accounts of users managed by the organization.
- Revert the #3263 hotfix that overloaded the remove-user `DELETE` endpoint: removing a managed user is rejected again (`USER_IS_MANAGED_BY_ORGANIZATION`); disabling/enabling is now done via the dedicated endpoints.
- Keep disabled **managed** users visible in the organization member list, grayed out and labelled, with enable/disable toggle controls (non-managed members keep the remove action; managed members no longer show a non-functional leave button). The member-list query only surfaces a disabled account when the organization itself disabled it, so both admin-disabled accounts and disabled non-managed accounts stay hidden — an org owner is never shown a member they have no control over.
- Expose `managed` / `disabled` flags on the organization member model.
- Make `UserAccountService.disable`/`enable` idempotent (no duplicate `OnUserCountChanged` event or state change on a no-op, clean `NotFound` for a missing user). This also changes the existing admin disable/enable endpoints from erroring to a no-op on repeat calls.
- Track *who* disabled an account in a new `user_account.disabled_by` column (`ADMIN` / `ORGANIZATION`), so an org owner can no longer re-enable a user a platform admin disabled — that attempt is rejected with `USER_DISABLED_BY_ADMIN`. An admin disable outranks an organization one: disabling an already org-disabled account escalates the origin to `ADMIN`, while the reverse never downgrades it. Neither escalation publishes a seat event, since `disabled_at` does not change. Enabling clears the origin. Existing disabled accounts are backfilled as `ADMIN`, except managed members, who are attributed to their organization.
- Tests: org-owner disable/enable (scope, self-guard, kill-switch for JWT + PAT, re-enable restored access, idempotency, cross-org rejection, managed-user cannot leave/be removed, listing visibility, project-listing divergence), disable-origin coverage (both escalation directions, org owner blocked from undoing an admin disable, admin-disabled member hidden from the listing, enable clears the origin), service-level seat-event assertions, admin idempotency, and updated e2e.

## Since review

- Seat limits are now enforced when an **organization** re-enables a managed member, through a new `OnBeforeOrganizationSeatIncrease` event that billing consumes (tolgee/billing#305). Platform admins are exempt — they must be able to re-enable whatever the customer's plan says.
- `CANNOT_DISABLE_PLATFORM_ADMIN` became `CANNOT_MANAGE_PLATFORM_STAFF_ACCOUNT`: it also covers supporters, and it is now raised on enable as well as disable.
- Disabled **staff** accounts are hidden from the member listing, so an owner is never shown a re-enable button that cannot succeed. The disable side is not symmetric: an *enabled* Tolgee-staff account that a customer's SSO provisioned as managed is listed normally and still shows a Disable button, which 400s with `cannot_manage_platform_staff_account`. The members model carries `managed`/`disabled` but not `role`, so the UI cannot tell; it fails safe, and it needs Tolgee staff to sign in through a customer tenant.
- Set-role and remove-user are scoped to members the organization can actually see, closing a path where an owner who guessed a user id could act on — and promote — an account a platform admin had disabled.
- The last-owner leave rule is a single `isLastEnabledOwner` predicate again.

### Second review pass

The review ran to convergence over twelve rounds. Beyond the items below it caught three defects that would
have shipped:

- **The changelog was not well-formed XML.** An angle-bracketed username pattern inside `<sql>` left changeSet
  `1785252020000-2` unparseable, so Liquibase failed on any fresh database. Local runs did not catch it because
  migrations only re-run when the test container is created from scratch.
- **The seat limit and invoicing usage were skipped for every new sign-up and invitation accept**, because the
  guard that spares disabled accounts read committed state from inside a pre-commit hook and so could not see a
  user created in the publishing transaction.
- **Changing a disabled member's role charged the organization a seat** it does not occupy, and refused the
  write once the plan was full. `main` could not reach this: `setMemberRole` resolved through `findActive`.

- The member-list query's disabled arm treated a legacy `user_account.role IS NULL` row as staff and hid it, while the Kotlin mirror called it visible — an owner could never see such a row to re-enable it. Both sides now agree.
- `isVisibleToOrganization` answered `true` for accounts with no relationship to the organization at all. It is now `isHiddenFromOrganizationByDisable`, which only claims what it checks.
- The self-disable guard sits in `OrganizationController.disableManagedUser`, next to its two siblings (`setUserRole`, `AdministrationController.disableUser`), rather than one layer down in the service.
- Disabled rows are dimmed with a literal `opacity: 0.6` (`DISABLED_ROW_OPACITY` in `MemberItem.tsx`). This is deliberate: the theme defines no emphasis/disabled-opacity token and nothing in the webapp uses MUI's `theme.palette.action.disabledOpacity`, so there is no token to reach for. It matches the existing `WebhookItem.tsx`; both call sites should migrate together if such a token is ever added.

### Known limitations

- An organization cannot remove a managed user, so a disabled managed account stays listed until support intervenes. Keeping the row visible is what makes re-enable reachable at all; proper deprovisioning is a separate product decision.
- **The backfill makes historical admin bans org-reversible.** Pre-column rows carry no discriminator, so the migration attributes every disable of a managed user to the organization. On deploy, an SSO-managed account that a Tolgee admin banned becomes visible again in that organization's member list and re-enableable by its owner; re-securing one requires a Tolgee admin to disable it again. The alternative default (ADMIN, or leaving the column NULL — `canOverrideDisable` treats them identically) is not free either: `removeOrDeactivateUser` on `main` disabled managed users through the ordinary DELETE path, so most existing disabled managed accounts are org-originated and would become permanently unre-enableable by their own organization, which is the population this feature exists for. Staff accounts are protected in either case by `checkOrganizationMayActOnAccount`. The guarantee that an organization can never undo an admin disable therefore holds for disables made from this changeSet onwards, not for historical rows.
- The seat-limit check on re-enable is scoped to the enabling organization, and is a read-then-check-then-store like every other seat path, so it is not serialised against concurrent enables.

## Summary by CodeRabbit

* **New Features**
  * Organization owners can disable and re-enable managed members.
  * Member listings show managed and disabled status.
  * Disabled members display a status indicator and reduced visibility.
  * Disable/enable actions require confirmation and provide success feedback.
* **Bug Fixes**
  * Prevented the final enabled organization owner from leaving.
  * Improved member removal and disabled-account access handling.
  * Recorded whether an account was disabled by an administrator or organization.
  * Prevented unauthorized management of unmanaged and protected accounts.


dkrizan (commented):
@Anty0, General question - I thought the purpose of disabling a user was to restrict it only to Admins. This means we can no longer disable a user from the Admin section because any organization manager or owner can re-enable them. Am I correct? If this is the intended functionality, I’m fine with it.

Anty0:
Yes, for managed users, the organization manages their accounts, including whether they are disabled. We can still disable the organization owner's account - doing so would prevent the organization from re-enabling its users, so there is still a way. I don't expect us to selectively disable individual accounts within an SSO organization.

If we want it separate, I could probably add a `reason` for the account being disabled and only allow the org owner to re-enable if they were the one who disabled it.

dkrizan:
My only concern is the admin moderation path, which this currently regresses. Maybe, keep `disabledAt` as the single login kill-switch and add a `disabledBy` flag (ADMIN / ORGANIZATION). Org `enable` only works when `disabledBy == ORGANIZATION`, and the member list surfaces org-disabled users only — so an admin-disabled account stays out of the org owner's reach. Cheap to add and closes the regression.

What do you think ?

Anty0:
Sure, I can do that. I didn't think it was worth the effort :3
I'll take a look at it during the next cooldown. ^^

Anty0:
@coderabbitai review

webapp/src/views/organizations/members/UserAvailabilityButton.tsx
```diff
@@ -0,0 +1,119 @@
...
+        onClick={confirmAndSubmit}
+        size="small"
+      >
+        {props.icon}
+      </IconButton>
+    </Tooltip>
+  );
+};
+
+export const DisableUserButton = (props: {
```
dkrizan, line 63: If I am not mistaken, we avoid having multiple components in a single .tsx file

dkrizan (dismissed):

Anty0:
Rebased. No code changes. Please re-approve ^^

Anty0:
Another rebase, this time one Test class got partially fixed up by the other PR, so last commit handles incompatibility there.

webapp/src/views/organizations/members/UserAvailabilityButton.tsx
Anty0: splitted into separate files

dkrizan (approved):

Issue #3763, opened 2026-06-21
Community translation - v1.0 (Platform)


ZuzanaOdstrcilova:
### 1. Problem
Tolgee is already used by many open-source projects, but it has no safe way to translate with a community. To involve outsiders today, you must add them to your project. Then they are members you have to manage, they count as paid seats, and you have to set their permissions. And even limited translate rights let them modify all strings in a language. Owners want something simpler: hand the community a link and let them contribute, without administering them, counting them as users, or dealing with permissions. We already have suggestions, which partly cover this, but they are only an intermediate step toward full community translation.

This is one of our top priorities this quarter, with strong user demand. In the age of AI translation, where quality varies, it makes even more sense to let the community and native speakers propose and review translations.

We need a better way: community members can only propose suggestions, while someone with permission approves and merges. They work through a limited view, without exposing the project or paying per volunteer.

---

### 2. Appetite
2 weeks, 1 developer.
_This is tight for the full scope, so the public projects page ships as a simple v1 (list + basic search)._ 

---

### Allow the project to be public
- Project settings: a "Public project" checkbox.
- Project view: a "Public project" / "Private project" icon.
- A public list of projects at "tolgee.io/public-projects". 

> 1. Simple list of Projects with basic search.
> 2. Opening a project or contributing requires login / registration; only the list itself is public.
> 3. Each project in the public view shows its icon, name, organization, and description (possibly some small stats).
> 4. Categories and advanced filters come later.

### Permissions
- Enable Tolgee users to propose translations of strings in public projects
- Enable project owners or people with specific permission to review the proposal and accept it

**"Community permission"**
- A new community permission, granted by a public project page link, not by adding the person as a project member or paid seat.
- Community members can only view translations and project stats, and propose suggestions.
- They cannot edit or overwrite strings, change state, or access anything else in the project.

> Note: Community can comment on a translation. So if the current translation (e.g. done by AI) is not good, they can comment on it there.

### Suggestions
- Community translators see a view with all suggestions and can suggest their own translation.

**Suggestion**
- A contributor only sees a Suggest button. They cannot change the live translation.
- The translation view shows 3 suggestions by default, with a "more" button.
- The accept / decline and hide logic will change:

> 1.  After accepting one suggestion, the others do not disappear.
> 2. A new "accept and hide" action hides them when the user really wants to, but for community purposes all suggestions should stay visible.

- After suggesting, a contributor can still delete their own suggestion (already exists).
[...]

PR #3770, opened 2026-06-27
feat: community permissions for public projects

Anty0:
Permission-resolution slice of Community Translation v1.0 (#3763). Stacked on `jirikuchynka/public-projects` — review/merge that first.

## What this does
Authenticated users on a **public** project now get a **computed** "community" permission (read subset of VIEW + `translations.suggest` + `translation-comments.add`, all languages; origin `COMMUNITY`) **without a stored `Permission` row** — so they contribute translations without taking a paid seat.

- **Floor, not a fallback:** any direct/organization permission layers on top of the community grant, and server admin/supporter elevation also layers on top — so **no authenticated user is ever below a non-member** on a public project, on either the scope or language axis. Anonymous users still get nothing.
- View + suggest are floored to **all languages** (language-restricted members can still only translate/state-change their granted languages).

## Member privacy (gated on `MEMBERS_VIEW`)
- Member emails (the `username` field) and project member count are redacted across project-scoped REST surfaces: suggestions, activity feed, translation history, comments, key-trash, members list, stats. Name + avatar + id are preserved.
- The realtime activity/batch **websocket never carries the author email** — a fan-out broadcast can't evaluate each recipient's `MEMBERS_VIEW`, so the email is stripped unconditionally there (REST applies the per-caller gate).

## `@UseDefaultPermissions` endpoint gating
The community scope set is non-empty, so it passes every project-scoped `@UseDefaultPermissions` endpoint. Audited all of them; out-of-surface ones are now restricted:
- MT credit balance → **`ORGANIZATION_QUOTAS_VIEW`** (new scope, under the VIEW group)
- import-settings store → `KEYS_CREATE`
- branch-merge session endpoints (×11) → `BRANCH_MANAGEMENT`
- `getBlockingTasks` → task-view-or-assigned (matching its siblings)

Intended community-visible reads (languages, namespaces, tags, labels, stats, comments, project metadata, **read-only MT/auto-translation settings**) stay open by design. Writes are self-guarding (`setTranslations` enforces `TRANSLATIONS_EDIT` in the facade, so community can't overwrite strings).

## ⚠️ Breaking changes
- `GET /v2/projects/{id}/machine-translation-credit-balance` now requires **`ORGANIZATION_QUOTAS_VIEW`** (granted by the VIEW/TRANSLATE/REVIEW/EDIT/MANAGE types). Pre-existing **granular/custom permissions** and **API keys** lacking it (or `all.view`) get **403** and must be re-granted. The platform webapp does not call this endpoint, so no UI breaks.
- Branch-merge session endpoints now require `BRANCH_MANAGEMENT`. This includes the read-only `GET` merge/preview/conflicts/changes endpoints, so members without `BRANCH_MANAGEMENT` (e.g. translators) lose read visibility into merge sessions — intentional, since `BRANCH_MANAGEMENT` is currently the only branch scope.
- **Member email masking is global, not public-only.** Author username/email is now redacted for **any** caller lacking `MEMBERS_VIEW` on **every** project, including private ones — previously a private-project member without `MEMBERS_VIEW` could still see author emails on suggestions/activity/history/comments. Treated as intentional privacy hardening; integrators relying on author emails must hold `MEMBERS_VIEW`.

## Tests
New: `CommunityPermissionTest` (HTTP), `CommunityPermissionComputationTest` (floor computation incl. supporter/admin/language cases), `CommunityPermissionScopesTest`, `MemberInfoMaskingTest`, `CommunitySuggestionTest`; plus redaction/gating assertions across suggestion/task/trash/websocket/notification suites. All green.

## Notes
- New i18n keys (`permissions_item_organization_quotas_view`, `project_settings_public_*`) were created in Tolgee; the generated `src/i18n/en.json` is intentionally **not** committed (it syncs from Tolgee).
- `WebsocketWithRedisTest` is flaky locally on Docker-Redis startup (env, not this change); `WebsocketWithoutRedisTest` covers the same code.

### DIFF
--- a/ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/CommunitySuggestionTest.kt
+++ b/ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/CommunitySuggestionTest.kt
@@ -158,16 +158,6 @@ class CommunitySuggestionTest : ProjectAuthControllerTest("/v2/projects/") {
   @Test
   @ProjectJWTAuthTestMethod
   fun `comment author emails are hidden from community`() {
-    userAccount = testData.user
-    performProjectAuthPost(
-      "translations/create-comment",
-      TranslationCommentWithLangKeyDto(
-        keyId = testData.keys[0].self.id,
-        languageId = testData.czechLanguage.id,
-        text = "Member comment",
-      ),
-    ).andIsCreated
-
     val translationId =
       transactionTemplate.execute {
         translationService.find(testData.keys[0].self, testData.czechLanguage).get().id
--- a/ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/CommunitySuggestionTest.kt
+++ b/ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/CommunitySuggestionTest.kt
@@ -1,0 +1,200 @@
+package io.tolgee.ee.api.v2.controllers
+
+import io.tolgee.ProjectAuthControllerTest
+import io.tolgee.development.testDataBuilder.data.SuggestionsTestData
+import io.tolgee.dtos.request.translation.SetTranslationsWithKeyDto
+import io.tolgee.dtos.request.translation.comment.TranslationCommentWithLangKeyDto
+import io.tolgee.ee.data.translationSuggestion.CreateTranslationSuggestionRequest
+import io.tolgee.fixtures.andAssertThatJson
+import io.tolgee.fixtures.andIsCreated
+import io.tolgee.fixtures.andIsForbidden
+import io.tolgee.fixtures.andIsOk
+import io.tolgee.fixtures.node
+import io.tolgee.testing.annotations.ProjectJWTAuthTestMethod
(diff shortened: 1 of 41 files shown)


ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/CommunitySuggestionTest.kt
```diff
@@ -0,0 +1,200 @@
...
+    performProjectAuthGet("activity").andIsOk.andAssertThatJson {
+      node("_embedded.activities[0].author.username").isEqualTo("suggestionsuggestionsTestUser")
+    }
+  }
+
+  @Test
+  @ProjectJWTAuthTestMethod
+  fun `comment author emails are hidden from community`() {
+    userAccount = testData.user
+    performProjectAuthPost(
```
dkrizan: comment probably can be prepared in test data

Anty0: Good catch ^^
I missed it

dkrizan (approved):

PR #3773, opened 2026-06-29
fix: reject misformatted table (XLSX/CSV) imports with a clear error

bdshadow:
## Problem

XLSX and CSV imports share a positional table format: **first column = key**, remaining columns = languages (defined by the header row). When a file is misformatted, the import silently produced **zero keys** — every row was recorded as a per-row "blank key" issue and skipped, with no clear signal that the whole file was structurally wrong.

Two real-world misformats triggered this:

1. **Empty first column** — keys live in column B, so every key (column A) is blank.
2. **Extra title row on top** — a title/filename row above the real header pushes the header into the data and leaves the language columns with blank headers.

A real file (`datahub-tolgee-keys.xlsx`) combined both: a `datahub-tolgee-keys` title row, then a `["", "key", "en"]` header shifted right by an empty first column.

## Change

Validate the parsed entries in the shared `TableProcessor` (covers XLSX **and** CSV in one place) and throw `ImportCannotParseFileException` with a clear message when:

- the key column is blank in **every** row, or
- **no** language headers are detected (e.g. an extra title row above the header).

Partial problems (a few stray blank keys) keep flowing through the existing per-row issue mechanism — only genuinely malformed files fail fast. Files that legitimately rely on the language fallback aren't affected (the fallback is never blank).

This PR also includes a small related change: `ImportFileProcessorFactory` now lists the supported file extensions in the parse error when no matching processor is found.

## Tests

- `import/xlsx/empty_key_column.xlsx` — the actual reported file → rejected.
- `import/csv/empty_key_column.csv` — empty key column → rejected.
- `import/csv/title_row_on_top.csv` — extra title row with valid keys → rejected.
- Existing XLSX/CSV processor tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved import error handling with clearer failure messages, including a list of supported file extensions when the input format can’t be recognized.
  * Added pre-import structural validation for table/spreadsheet data to block invalid files—such as rows with empty key columns or missing/undetected language headers.
* **Tests**
  * Added/extended unit tests for malformed CSV and XLSX inputs to ensure they reliably fail with `ImportCannotParseFileException` and file-specific error details.


backend/data/src/main/kotlin/io/tolgee/formats/genericTable/in/TableProcessor.kt
```diff
@@ -10,9 +11,29 @@ abstract class TableProcessor(
 ) : ImportFileProcessor() {
   override fun process() {
     val (data, format) = parse()
+    data.validateStructure()
```
Anty0, line 14: `validateStructure()` traverses `data` up to 3× (the `none()` checks) and then `importAll()` traverses it again. `data` comes from `parse(): Pair<Iterable<TableEntry>, ImportFormat>`, and the `Iterable` element type doesn't guarantee re-iteration.

It's safe today — both `CsvFileProcessor` and `XlsxFileProcessor` return a materialized `List`. But the contract permits a one-shot source (e.g. a `Sequence.asIterable()`). If a future processor returns one, the first `none()` consumes it, validation passes vacuously, and `importAll()` then imports nothing — a silent data-loss path where a malformed file neither errors nor imports.

Suggest pinning it to a finite, re-iterable collection. Either narrow the abstract return type:

```kotlin
abstract fun parse(): Pair<List<TableEntry>, ImportFormat>
```

backend/data/src/main/kotlin/io/tolgee/formats/importCommon/ImportFileFormat.kt
```diff
@@ -26,6 +26,10 @@ enum class ImportFileFormat(
         }.toMap()
     }
 
+    val supportedExtensions: List<String> by lazy {
```
Anty0, line 29: nit: the`extensionFormatMap` already derives the same thing. Could reuse it to keep one source of truth:

```kotlin
val supportedExtensions: List<String> by lazy { extensionFormatMap.keys.toList() }
```

Anty0 (dismissed):
lgtm
Only two small nits found by agent.

bdshadow:
@Anty0 thank you! both fixed

Anty0 (approved):

PR #3779, opened 2026-06-30
fix: protect MT placeholders from AWS Translate mangling for Traditional Chinese

bdshadow:
## Problem

For some MT targets — notably **Traditional Chinese** (`zh-Hant`, sent to AWS as `zh-TW`) — AWS Translate rewrites Tolgee's ICU-parameter placeholders, so the translated suggestion comes back broken.

Tolgee protects ICU params before MT by swapping `{current} / {max}` for bare sentinel tokens `{xx0xx} / {xx1xx}` (`TextHelper.replaceIcuParams`) and restores them afterward by **exact string match** (`MtBatchTranslator.replaceParams`). The token `xxNxx` is a lowercase, word-like string, and AWS's CJK model *translates* it, so the exact-match restore no longer finds it and a broken placeholder leaks through:

```
{current} / {max}  ->  {七十八次}/{xx1 次}
```

The mangling is **intermittent** (AWS rewrites the token only sometimes), which matches the client report that "the `xx1` pattern comes back often".

## Root cause

The bare `{xxNxx}` sentinel is not opaque to MT engines. It was introduced in the very first MT commit and has never been validated to survive any provider round-trip — it merely happened to work for Latin-script targets. CJK is the first target aggressive enough to translate it.

## Fix

Add a provider-overridable `MtPlaceholderProtector` applied at the single shared choke point (`AbstractMtValueProvider.translate()`):

- **`NoopMtPlaceholderProtector`** (default) — identity; every non-AWS provider is unchanged.
- **`HtmlNoTranslatePlaceholderProtector`** (AWS) — wraps each sentinel in `<span translate="no">…</span>`. AWS Translate honors `translate="no"` and returns the wrapped content verbatim; the wrapping is stripped from the response.

`restore` only unwraps spans whose inner content is a sentinel, so a user-authored `<span translate="no">…</span>` in the source is never touched.

## Verification

Confirmed directly against AWS Translate (`en` → `zh-TW`):

| Input | AWS output | Result |
|-------|-----------|--------|
| `{xx0xx} / {xx1xx}` | `{七十八次}/{xx1 次}` | 🔴 reproduces the bug |
| `<span translate="no">{xx0xx}</span> / <span translate="no">{xx1xx}</span>` | `<span translate="no">{xx0xx}</span>/<span translate="no">{xx1xx}</span>` | 🟢 tokens verbatim |
| `Hello <span translate="no">DoNotTranslateMe</span> world` | `你好 <span translate="no">DoNotTranslateMe</span> 世界` | 🟢 `translate="no"` honored |

The same token AWS mangles when bare comes back untouched when wrapped. 8 unit tests cover the protect/restore round-trip, the exact real-world AWS output, quote-less attributes, and preservation of user-authored spans.

> Local AWS verification requires a separately-issued, translate-only AWS key — prod authenticates via ambient cluster identity, so there is no static key to reuse for a local run.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of placeholder values during machine translation, helping keep variable tokens intact.
  * Added support for protecting placeholders in translations using HTML “do not translate” markers where needed.

* **Bug Fixes**
  * Reduced the risk of placeholders being altered, dropped, or merged by translation engines.
  * Improved restoration of original placeholder text after translation, including tricky HTML formatting cases.

### DIFF
--- a/backend/data/src/main/java/io/tolgee/helpers/TextHelper.kt
+++ b/backend/data/src/main/java/io/tolgee/helpers/TextHelper.kt
@@ -4,6 +4,12 @@ import java.util.regex.MatchResult
 import java.util.regex.Pattern
 
 object TextHelper {
+  private const val PARAM_PLACEHOLDER_PREFIX = "{xx"
+  private const val PARAM_PLACEHOLDER_SUFFIX = "xx}"
+
+  val paramPlaceholderRegex =
+    Regex(Regex.escape(PARAM_PLACEHOLDER_PREFIX) + "\\d+" + Regex.escape(PARAM_PLACEHOLDER_SUFFIX))
+
   @JvmStatic
   fun splitOnNonEscapedDelimiter(
     string: String,
@@ -100,7 +106,7 @@ object TextHelper {
         buffer.append(char)
 
         if (openCount < 1) {
-          val paramPlaceholder = "{xx${params.size}xx}"
+          val paramPlaceholder = "$PARAM_PLACEHOLDER_PREFIX${params.size}$PARAM_PLACEHOLDER_SUFFIX"
           params[paramPlaceholder] = buffer.toString()
           buffer.clear()
           result.append(paramPlaceholder)
--- a/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/AbstractMtValueProvider.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/providers/AbstractMtValueProvider.kt
@@ -4,6 +4,8 @@ import io.tolgee.component.machineTranslation.MtValueProvider
 import io.tolgee.component.machineTranslation.metadata.MtMetadata
 
 abstract class AbstractMtValueProvider : MtValueProvider {
+  protected open val placeholderProtector: MtPlaceholderProtector = NoopMtPlaceholderProtector
+
   private val String.toSuitableTag: String?
     get() = getSuitableTag(this)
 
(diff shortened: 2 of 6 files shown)


dkrizan (commented):
Looks good!

Looks like the same trick works for the other providers too, so might be worth doing them all in this PR:

- **Google** - has the same `translate="no"` exclusion, but it only kicks in when the request is sent as HTML. We currently force format to `text`, so we'd need to switch it to `html`.
https://docs.cloud.google.com/distributed-cloud/hosted/docs/latest/gdcag/application/ao-user/quickstart-translation#exclude-text
- **DeepL** - same idea with ignored tags, you just have to declare the ignore tag in the request (`tag_handling=xml` + `ignore_tags`).
https://developers.deepl.com/docs/xml-and-html-handling/xml#ignored-tags
- **Azure** - has an equivalent too (`textType=html` + `<span class="notranslate">`), if we want to cover everything.
https://learn.microsoft.com/en-us/azure/ai-services/translator/text-translation/how-to/prevent-translation

Btw, once every provider protects params on its own, the `{xxNxx}` placeholder doesn't really need to sit in `MtBatchTranslator` anymore. It could just become the default `MtPlaceholderProtector` in `AbstractMtValueProvider`, providers with native support override it, and only Baidu/LLM fall back to the `{xxNxx}` replacement.

What do you think?

bdshadow:
@dkrizan thank you! I've checked all three docs and you're right, the mechanism exists for every provider:

- Google - `<span translate="no"> / class="notranslate"`, needs `format=html`
- DeepL - `tag_handling=xml` + ignore_tags
- Azure - `textType=html` + `notranslate/translate="no"`

However, I'd like to keep this PR scoped to AWS, for a few reasons:

1. AWS is a special case that made it low-risk. AWS TranslateText already parses tags in every request, so this PR didn't change AWS's mode - it just leaned on behavior that was already on. The other three require flipping the request into HTML/XML mode, which changes how the provider treats the source text of every translation, not only ones with params. In that mode the engine interprets markup/entities, so a user's literal <, >, & can get swallowed or entity-encoded. Doing it safely means adding an escape/unescape layer around the non-placeholder text
2. The others aren't actually broken. So converting a working provider to HTML mode is regression risk for no demonstrated bug.
3. Doc caveat for Google: that link is Google Distributed Cloud (hosted), but we're on the Cloud Translation v2 Basic client. `format=html` preserves tags there, but I couldn't confirm v2 Basic honors translate="no" specifically - would want a real test first.

dkrizan:
@bdshadow 
Ok, let's keep this PR AWS-only.

One thing I'd still like to do eventually, but as a separate follow-up: right now the params get transformed twice, first into `{xxNxx}`, then wrapped in `<span translate="no">` in the provider. It'd be cleaner to wrap the real param directly and drop the `{xxNxx}` step entirely.

dkrizan (approved):

PR #3783, opened 2026-07-02
fix: hide and purge projects when their organization is deleted

bdshadow:
## Problem

The **Generate Project API key** dialog listed projects belonging to already-deleted organizations. The `/v2/projects` response itself returned those projects — a backend issue, not a stale frontend cache.

## Root cause

Soft-deleting an organization (`OrganizationService.delete()`) set the organization's `deletedAt` but left its projects untouched. The project-listing query (`ProjectRepository.findAllPermitted`) filters the project's own `deletedAt` but not the owning organization's, so projects of a soft-deleted organization kept appearing.

## Fix

Two parts, so the existing `deletedAt is null` filters handle everything (no per-query org-deletion filter to remember):

1. **Immediate:** `OrganizationService.delete()` bulk soft-deletes the organization's active projects, so they vanish from listings right away. A bulk `UPDATE` is used rather than `ProjectService.deleteProject`, because `deleteProject` emits `OnProjectSoftDeleted` → an `@Async @TransactionalEventListener` **hard delete** that runs detached from the caller (and, from test teardown, deadlocked against the per-test DB reset).

2. **Cleanup:** `DeletedOrganizationProjectPurgeScheduler` hard-deletes projects whose owning organization is soft-deleted — reclaiming their data and also cleaning up projects orphaned under organizations deleted *before* this cascade existed (so no data migration is needed). It follows the existing `KeyTrashPurgeScheduler` pattern (`SchedulingManager`, distributed lock, batched deletes in separate transactions) and is gated by `tolgee.internal.orphan-project-purge-enabled`, disabled in tests so the hard-delete never races the per-test `TRUNCATE`.

## Tests

- `SoftDeleteTest`: `queries don't return projects of a deleted organization`.
- `DeletedOrganizationProjectPurgeSchedulerTest`: purges projects of a soft-deleted org; keeps projects of an active org.
- `TranslationMemoryServiceTest`: the two project-editing tests now commit the edit in `executeInNewTransaction` so they don't hold a project row lock across `@AfterEach` teardown (teardown's org-delete cascade updates the same row).

All pass locally (incl. `OrganizationServiceTest` and `KeyTrashPurgeSchedulerTest` to confirm scheduling is unaffected).

## Summary by CodeRabbit

* **New Features**
  * Projects are automatically soft-deleted when an organization is deleted, and a background cleanup later permanently removes orphaned projects.
  * Added an internal configuration switch to enable/disable the orphan cleanup per environment.

* **Bug Fixes**
  * Improved API key generation so the dialog remains stable if the previously selected project disappears after refresh.
  * Fixed organization deletion behavior so project listings no longer include projects from deleted organizations.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/project/DeletedOrganizationProjectPurgeScheduler.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/project/DeletedOrganizationProjectPurgeScheduler.kt
@@ -1,0 +1,82 @@
+package io.tolgee.service.project
+
+import io.tolgee.component.LockingProvider
+import io.tolgee.component.SchedulingManager
+import io.tolgee.configuration.tolgee.InternalProperties
+import io.tolgee.util.Logging
+import io.tolgee.util.executeInNewTransaction
+import io.tolgee.util.logger
+import io.tolgee.util.runSentryCatching
+import org.springframework.boot.context.event.ApplicationReadyEvent
+import org.springframework.context.event.EventListener
+import org.springframework.data.domain.PageRequest
+import org.springframework.stereotype.Component
+import org.springframework.transaction.PlatformTransactionManager
+import java.time.Duration
+
+/**
+ * Hard-deletes projects whose owning organization was soft-deleted.
+ *
+ * Organization soft-delete only marks the projects deleted (so they disappear from listings
+ * immediately); this scheduler reclaims their data afterwards. It also cleans up projects that
+ * were orphaned under an organization deleted before that cascade existed.
+ */
+@Component
+class DeletedOrganizationProjectPurgeScheduler(
+  private val projectService: ProjectService,
+  private val projectHardDeletingService: ProjectHardDeletingService,
+  private val lockingProvider: LockingProvider,
+  private val transactionManager: PlatformTransactionManager,
+  private val schedulingManager: SchedulingManager,
+  private val internalProperties: InternalProperties,
+) : Logging {
+  @EventListener(ApplicationReadyEvent::class)
+  fun schedulePurge() {
+    if (!internalProperties.orphanProjectPurgeEnabled) {
+      logger.info("Orphan project purge is disabled, skipping scheduling")
+      return
+    }
+    schedulingManager.scheduleWithFixedDelay(::purge, PURGE_PERIOD)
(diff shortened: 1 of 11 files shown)


dkrizan (dismissed):

backend/data/src/main/kotlin/io/tolgee/service/project/ProjectService.kt
```diff
@@ -315,6 +315,22 @@ class ProjectService(
     return projectRepository.findAllByOrganizationOwnerIdAndDeletedAtIsNull(organizationId)
   }
 
+  @Transactional
+  @CacheEvict(cacheNames = [Caches.PROJECTS], allEntries = true)
+  fun softDeleteAllInOrganization(organizationId: Long) {
```
dkrizan, line 320: `deleteProject` also sets `deletedAt` on all project languages, it might be worth doing so here as well

backend/data/src/main/kotlin/io/tolgee/service/project/DeletedOrganizationProjectPurgeScheduler.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    if (totalPurged > 0) {
+      logger.info("Purged {} projects of deleted organizations", totalPurged)
+    }
+  }
+
+  companion object {
+    private const val PURGE_LOCK_NAME = "orphan_project_purge_lock"
+    private const val BATCH_SIZE = 50
+    private val PURGE_PERIOD = Duration.ofHours(1)
```
dkrizan: I believe running it once every 24 hours is sufficient, but I’ll leave it up to you.

backend/data/src/main/kotlin/io/tolgee/service/project/ProjectService.kt
bdshadow: done

backend/data/src/main/kotlin/io/tolgee/service/project/DeletedOrganizationProjectPurgeScheduler.kt
bdshadow: done

dkrizan (approved):

PR #3786, opened 2026-07-03
fix: use key's configured plural arg name for machine translation

### LINKED ISSUE #3772: Machine translation using "value" for plural variable name
**Describe the bug**
Machine translation using "value" for plural variable name.

**To Reproduce**

Create key with "count" for plural variable name 
Image

Image

Use Machine translation for other lang
Image

After clicking "Save" in key settings, plural variable name became "count" for all lang.
Image

 

**Expected behavior**
Machine translation should use plural variable name from key settings, not "value"

bdshadow:
Fixes #3772

## Problem

Machine translation of a plural key returned whatever plural argument name the model emitted, rather than the key's configured `pluralArgName`. Only the `PROMPT` (LLM) service supports plurals, and it returns raw text (`PromptServiceEeImpl.getTranslationFromPromptResult`); the MT pipeline stored that verbatim (`MtBatchTranslator` line 109) without ever consulting the key's setting. The *save* path normalizes to `key.pluralArgName`, which is why re-saving the key "fixed" it.

With a strong model the emitted name usually matches, so it looks fine. With a weaker model it breaks reliably — verified locally against Ollama `phi3`, where the source `{itemCount, plural, …}` came back as `{Betrag, …}` (translated) or `{count, …}` (copied from the prompt example), never `itemCount`.

Two contributing factors:
1. **The pipeline never enforced `key.pluralArgName`** — it trusted the model's output.
2. **The prompt's plural example hardcoded `count`** (`PromptVariablesHelper`), actively teaching weaker models to emit `count` regardless of the real arg name.

## Fix

- Thread the key's `pluralArgName` into `KeyForMt` (via the MT key query) and, in `MtBatchTranslator.translatePlural`, rewrite the translated plural's arg name using `getPluralArgName`: **key's `pluralArgName` → base translation's arg name → default (`value`)**. This is the same source of truth the save path uses, so MT output no longer depends on the model.
- Render the prompt's `exampleIcuPlural` with the actual arg name instead of the hardcoded `count`.

## Tests

New regression tests in `MtBatchTranslatorTest` covering both branches of the arg-name resolution:
- `falls back to base plural arg name when key has none` — `key.pluralArgName == null`, base carries the name.
- `forces configured key plural arg name even when base differs` — key setting wins over a mismatched base (the reporter's scenario).

Both assert the model's `{value, …}` output is rewritten to the correct name. Existing plural MT tests unchanged and passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of plural argument names in machine-translated plural strings, so translated content now keeps the expected placeholder name more consistently.
  * When available, a key’s configured plural argument name is now used in generated plural examples and translations, with sensible fallback behavior when it isn’t set.
  * Added coverage for cases where the key’s plural argument name differs from the source translation or is missing.

### DIFF
--- a/backend/data/src/test/kotlin/io/tolgee/service/machineTranslation/MtBatchTranslatorTest.kt
+++ b/backend/data/src/test/kotlin/io/tolgee/service/machineTranslation/MtBatchTranslatorTest.kt
@@ -95,6 +95,52 @@ class MtBatchTranslatorTest {
     translated.actualPrice.assert.isEqualTo(100)
   }
 
+  @Test
+  fun `falls back to base plural arg name when key has none (gh-3772)`() {
+    prepareKeyWithBaseArgNameOnly()
+    prepareRawPluralResponse()
+
+    val translated = translateSingle(MtServiceType.PROMPT)
+
+    translated.translatedText.assert.isEqualTo(
+      "{count, plural,\n" +
+        "one {Jeden pes}\n" +
+        "other {'#' psů}\n" +
+        "}",
+    )
+  }
+
+  @Test
+  fun `forces configured key plural arg name even when base differs (gh-3772)`() {
+    prepareKeyWithMismatchedArgName()
+    prepareRawPluralResponse()
+
+    val translated = translateSingle(MtServiceType.PROMPT)
+
+    translated.translatedText.assert.isEqualTo(
+      "{count, plural,\n" +
+        "one {Jeden pes}\n" +
+        "other {'#' psů}\n" +
+        "}",
+    )
+  }
+
+  private fun translateSingle(service: MtServiceType): MtTranslatorResult {
+    val context = getMtTranslatorContext()
+    val translator = MtBatchTranslator(context)
+    return translator
+      .translate(
+        listOf(
+          MtBatchItemParams(
+            keyId = 1,
+            baseTranslationText = preparedKey.baseTranslation,
+            targetLanguageId = 1,
+            service = service,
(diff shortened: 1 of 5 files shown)


backend/data/src/test/kotlin/io/tolgee/service/machineTranslation/MtBatchTranslatorTest.kt
```diff
@@ -95,6 +95,52 @@ class MtBatchTranslatorTest {
     translated.actualPrice.assert.isEqualTo(100)
   }
 
+  @Test
+  fun `falls back to base plural arg name when key has none (gh-3772)`() {
```
dkrizan, line 99: remove `(gh-3772)` from test name please (same in other tests)

bdshadow: why? It's a common approach to have it like this (or as a comment above the test). Hibernate and Spring-framework use it a lot. Allows you to navigate to the issue, which this test checks, right away. Why don't you like it?

dkrizan: I see, but for Tolgee it still feels mostly like noise to me. The issue is already linked through the PR, commit history, while the test name should ideally just describe the behavior. Feel free to close it.

dkrizan (approved):

PR #3789, opened 2026-07-05
fix: keep zero plural form in Apple XLIFF export

### LINKED ISSUE #3074: Apple XLIFF plurals are not functioning
**Describe the bug**

We’re hitting some issues with the Apple XLIFF exports where plurals are not being exported as expected:

Image

In the exported file this is available:

```
<trans-unit id="ATTENDEE_COUNT|==|plural.one">
 <source xml:space="preserve">1 Participant</source>
 <target xml:space="preserve">1 Participant</target>
 </trans-unit>
 <trans-unit id="ATTENDEE_COUNT|==|plural.other">
 <source xml:space="preserve">%%d Participants</source>
 <target xml:space="preserve">%%d Participants</target>
 </trans-unit>

```

But `plural.zero` is not available. 

**To Reproduce**
1. In a project great a translation key that is a plural that contains a 'zero'.
2. Click on export with the following settings:

Image

4. Open the exported file and search for "plural.zero" or "No participants"
5. String not found

**Expected behavior**
"Zero" should be in the exported xliff file. 

**Screenshots**
See above.

**Versions and environment**
Tolgee Cloud 16th May 2025 @ 5pm BST

**Additional context**
N/A

munzzyy:
## Problem

Apple XLIFF export drops any plural form that CLDR doesn't define for the target locale. English only has `one` and `other` in CLDR, so a key with a `zero` form (e.g. "No participants" / "1 Participant" / "%d Participants") gets exported without `plural.zero` at all. The string is just gone from the file.

`AppleXliffExporter.handlePlural` runs the converted forms through `populateForms`, which builds its output set from `getPluralFormsForLocale(languageTag)` only. Any form outside that set gets dropped instead of exported. This is locale-specific behavior, not something Apple's formats actually require: `AppleXcstringsExporter` (the `.xcstrings` exporter) skips this filtering entirely and just writes out whatever forms `formsResult` contains. That's why the reporter's `.xcstrings` workaround kept the zero case while XLIFF/stringsdict export lost it.

## Change

`AppleXliffExporter`'s `populateForms` now keeps every form that's actually present in the message, on top of the locale's CLDR forms (so `many` for Czech still gets filled in from `other` when the message doesn't specify it, same as before). This lines XLIFF export up with what the xcstrings exporter already does, so a `zero` form the user wrote makes it into the export regardless of whether the target locale's CLDR data includes it.

I left `io.tolgee.formats.populateForms` (used by the Android XML resources exporter) alone. Android's plural resources really are constrained to CLDR quantities, so filtering there is correct.

## Tests

Added a test to `AppleXliffFileExporterTest` with an English key carrying `zero`/`one`/`other` forms and checked all three trans-units, including `plural.zero`, show up in the exported XLIFF.

No JDK on this machine, so I couldn't run `./gradlew :data:test` locally. I traced the fix by hand against `PluralsFormUtilTest` (which already asserts `getPluralFormsForLocale("en")` returns only `one, other`) and against how `AppleXcstringsExporter` builds its output. Would appreciate CI confirming the new test passes.

Fixes #3074

## Summary by CodeRabbit

* **Bug Fixes**
  * Apple XLIFF exports now include the plural forms required for each locale.
  * Explicit zero plural translations are preserved, including `=0` translations that are mapped to the zero category when appropriate.
  * Missing plural values fall back to the “other” translation to help prevent incomplete exports.

* **Tests**
  * Added coverage for English exports containing explicit zero and `=0` plural forms.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/formats/apple/out/AppleXliffExporter.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/apple/out/AppleXliffExporter.kt
@@ -5,6 +5,7 @@ import io.tolgee.formats.PossiblePluralConversionResult
 import io.tolgee.formats.apple.APPLE_CORRESPONDING_STRINGS_FILE_ORIGINAL
 import io.tolgee.formats.apple.APPLE_FILE_ORIGINAL_CUSTOM_KEY
 import io.tolgee.formats.apple.APPLE_PLURAL_PROPERTY_CUSTOM_KEY
+import io.tolgee.formats.getPluralFormsForLocale
 import io.tolgee.formats.xliff.model.XliffFile
 import io.tolgee.formats.xliff.model.XliffModel
 import io.tolgee.formats.xliff.model.XliffTransUnit
@@ -231,10 +232,13 @@ class AppleXliffExporter(
     languageTag: String,
     conversionResult: PossiblePluralConversionResult?,
   ): Map<String, String> {
-    if (conversionResult?.formsResult == null) {
-      return emptyMap()
-    }
-    return io.tolgee.formats.populateForms(languageTag, conversionResult.formsResult)
+    val forms = conversionResult?.formsResult ?: return emptyMap()
+    // Apple's plural formats (stringsdict/xcstrings) accept any form the user actually wrote
+    // (e.g. "zero"), not just the forms CLDR defines for the target locale, so we keep every
+    // authored form in addition to filling in the ones CLDR expects (e.g. "many" for cs).
+    val otherForm = forms["other"] ?: ""
+    val allForms = getPluralFormsForLocale(languageTag) + forms.keys
+    return allForms.associateWith { forms[it] ?: otherForm }
   }
 
   private fun getResultXliffFile(
--- a/backend/data/src/main/kotlin/io/tolgee/formats/apple/out/AppleXliffExporter.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/formats/apple/out/AppleXliffExporter.kt
@@ -5,6 +5,7 @@ import io.tolgee.formats.PossiblePluralConversionResult
 import io.tolgee.formats.apple.APPLE_CORRESPONDING_STRINGS_FILE_ORIGINAL
 import io.tolgee.formats.apple.APPLE_FILE_ORIGINAL_CUSTOM_KEY
 import io.tolgee.formats.apple.APPLE_PLURAL_PROPERTY_CUSTOM_KEY
+import io.tolgee.formats.formKeywords
 import io.tolgee.formats.getPluralFormsForLocale
 import io.tolgee.formats.xliff.model.XliffFile
 import io.tolgee.formats.xliff.model.XliffModel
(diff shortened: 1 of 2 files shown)


munzzyy:
Deliberately not delegating to `io.tolgee.formats.populateForms` here. That helper builds its result set from `getPluralFormsForLocale(languageTag)` alone, which is exactly the bug this PR fixes — it's why a `zero` form written by the user gets dropped for locales like English where CLDR doesn't define `zero`. The local version adds `+ forms.keys` so any form the user actually wrote survives, in addition to the CLDR-required ones.

The shared helper stays correct as-is for the Android XML resources exporter, since Android plural resources really are constrained to CLDR quantities (noted in the PR description). Keeping this one local avoids changing behavior for that other caller. Happy to rename it to something more specific like `populateAppleForms` if the duplication is confusing, but I'd rather not merge the two implementations.

munzzyy:
Checked the shared helper before switching to it: `pluralFormsUtil.kt`'s `populateForms` (lines 33-40) builds its result set from `getPluralFormsForLocale(languageTag)` alone, so for `en` (CLDR forms `one, other`) it drops any `zero` form the user actually wrote — which is the exact bug this PR fixes, and what `AppleXliffFileExporterTest`'s new "keeps a zero plural form" test asserts against. My local version unions in `forms.keys` on top of the CLDR set specifically to keep that. Leaving the inline version as-is since delegating here would reintroduce the drop.

munzzyy:
On the suggestion to reuse `populateForms`: that helper builds its form list from `getPluralFormsForLocale(languageTag)` on its own, so it keeps only the CLDR forms for the locale and drops any authored form CLDR doesn't list. For `en` that's the `zero` form this PR is trying to keep, so delegating back to it would undo the fix.

I kept the change inside the Apple exporter on purpose instead of widening the shared helper. Apple's stringsdict/xcstrings accept whatever forms the author wrote, but the other formats that call `populateForms` don't necessarily want CLDR-extra forms, and broadening the shared helper would change their output too. Happy to move this into an Apple-specific helper if you'd rather not inline it, but it shouldn't route back through `populateForms` as it stands.

Anty0:
Hi! Thanks for the PR ^^
Technically, no other plural forms are supported for the language, and per the ICU spec, it's undefined behavior. That's probably why we strip them.

For ICU, the proper way to handle the `zero` case is `=0` rather than `zero`, since `zero` isn't guaranteed to mean only `0`. I'm not sure whether Apple supports these explicit `=N` forms. I'll check with the rest of the team to see what they think about this change.

munzzyy:
Checked the Apple side of this, since that's what the exporter feeds. Apple's plural docs use exactly this case as their worked example: English with an optional `zero` variant — "the English language only requires the `one` and `other` categories to represent plural forms, and `zero` is optional" — and the sample stringsdict carries `zero` -> "No homes found". That example only does anything if the `zero` variant renders for a count of 0 in English, so on Apple's side it's a documented optional category for every language rather than undefined behavior. And there's no `=N` syntax in stringsdict or string catalogs; `zero` is how that format spells the 0 case.

https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals

I think that's the reason to keep this scoped to the Apple XLIFF exporter: in ICU terms an authored `zero` for English is nonstandard, agreed — but the Apple target doesn't consume ICU, and stripping the form on export drops content that Xcode itself round-trips fine. Curious what the team comes back with.

Anty0:
Hmm, in that case, we could probably keep all the standard categories (including `zero`) and filter out `=N`. On top of that, if the plural doesn't have `zero` but does have `=0`, we could fold it into `zero`. What do you think? I'll still check with colleagues, but this sounds like the best way to go. Also, there are two Apple formats. I wonder whether the other behaves differently or whether we should fold the changes into that format too.

munzzyy:
That works, and it's a better shape than what I have.

Filtering `=N` is the part my patch currently gets wrong. I merge `forms.keys` wholesale, so an authored `=0` goes straight into the XLIFF as a `=0` quantity key, and that isn't one of the six Apple accepts. Restricting the merge to `formKeywords` fixes it.

Folding `=0` into `zero` when there's no `zero` makes sense for the Apple targets specifically, since stringsdict and xcstrings have no way to express an exact match. `zero` is the closest thing the format has.

On the second format: `populateForms` is only called from `AppleXliffExporter`, so xcstrings and strings/stringsdict never hit the locale filter and don't strip an authored `zero` today. The `=N` half may still apply to them though. `StringsdictWriter` writes the form key straight into the plist as the quantity key, so a `=0` would land there verbatim. I'd rather add a test and see what those two actually emit than assume it.

Happy to implement it the way you describe, with the `=N` filter and the `=0` fold, and to put it somewhere shared so all the Apple formats get the same treatment instead of just XLIFF. Want me to push that onto this PR?

munzzyy:
Pushed the shape we talked through. `populateForms` now keeps the locale's CLDR forms plus any named category the user actually wrote, and drops the `=N` keys instead of merging them in — so an authored `=0` no longer lands in the XLIFF as a `=0` quantity. When a plural has `=0` but no explicit `zero`, it comes out as `zero`, since that's the nearest thing stringsdict/xcstrings can say. Added a test for that fold next to the existing zero-form one.

No rush on your end — if the team lands somewhere different after you've talked it over, I'm happy to reshape it.

PR #3798, opened 2026-07-09
fix: raise default LLM max tokens, drop temperature from Anthropic requests

JanCizmar:
Fixes for the Anthropic LLM provider, following the production incident where AI batch translation failed with `LLM_PROVIDER_NOT_RETURNED_JSON`:

- **`MAX_TOKENS_DEFAULT` 2000 → 8192.** Longer translations can hit a 2000-token output cap, and a response truncated by `max_tokens` is unparseable JSON even when structured outputs are enabled. 8192 is within every configured provider's output limit; it's a cap, not a cost — billing is by actual tokens.
- **Anthropic requests no longer send `temperature: 0`.** It was hardcoded in the request body. Claude Sonnet 5 / Opus 4.7+ reject non-default sampling parameters, so any model bump would start returning 400s. The OpenAI service already omits temperature.
- **Explicit handling for responses truncated by the output limit.** `AnthropicApiService` now checks `stop_reason` and throws `LlmProviderMaxTokensExceededException` (`LLM_PROVIDER_MAX_TOKENS_EXCEEDED`) instead of letting the truncated text fail later as generic not-returned-JSON. Batch translation classifies it like the other consistent-per-input LLM failures (builds on #3797): remaining chunk items still get translated, the failed item is not retried.

Companion config change enabling `format: json_schema` (structured outputs) for the Anthropic providers: tolgee/deployment#755.

## Summary by CodeRabbit

* **New Features**
  * Increased the default LLM token limit (up to 8192), enabling longer responses by default.
  * Added clearer handling for token-limit truncation, including a dedicated user-facing error code/message.

* **Bug Fixes**
  * Improved batch processing so a token-limit failure no longer stops other items in the same batch from being processed.
  * Updated web app messaging to reflect truncation due to output token limits consistently.


backend/data/src/main/kotlin/io/tolgee/batch/MtProviderCatching.kt
```diff
@@ -47,6 +48,10 @@ class MtProviderCatching(
         exceptions.add(
           RequeueWithDelayException(Message.LLM_PROVIDER_NOT_RETURNED_JSON, successfulTargets, e, maxRetries = 0),
         )
+      } catch (e: LlmProviderMaxTokensExceededException) {
+        exceptions.add(
+          RequeueWithDelayException(Message.LLM_PROVIDER_MAX_TOKENS_EXCEEDED, successfulTargets, e, maxRetries = 0),
```
JanCizmar: Why it's requeue with delay. We don't want to requeue it...

JanCizmar: It doesn't actually requeue — with `maxRetries = 0`, `ChunkProcessingUtil.retryFailedExecution` hits `errorKeyRetries (0) >= maxRetries (0)` on the first failure and returns without scheduling a retry execution, so the chunk just fails.

The reason for using `RequeueWithDelayException` instead of `FailedDontRequeueException` is the aggregation path: `iterateCatching` collects per-item failures into the `exceptions` list (typed `List<RequeueWithDelayException>` via `MultipleItemsFailedException`), which is what lets the loop keep translating the remaining items of the chunk. `FailedDontRequeueException` is thrown immediately, so one truncated key would again abort the rest of the chunk — the thing #3797 fixed. Same pattern as the not-returned-JSON and empty-response branches above.

So the semantics here are 'collectable per-item failure with zero retries'; the class name just reads badly. I can rename it (e.g. `ChunkItemFailedException` with a `maxRetries` field) or add a clarifying comment at this catch if you prefer.

JanCizmar: Renamed to `ChunkItemFailedException` in fe0ee1b12.

PR #3799, opened 2026-07-09
feat: add escapeJson helper for custom AI prompts

### LINKED ISSUE #3683: Provide escaped key descriptions and text as template variables in custom LLM prompts
When defining a custom prompt, you may want to inject key and translation metadata into a structured JSON or YAML or other structure to be parsed by your custom model.  I tried using this for a custom prompt:
```json
{
  "context": "{{key.name}} {{key.description}}",
  "text": "{{source.translation}}",
  "locale": "{{target.languageTag}}"
}
```

It worked great, except the text of the key description was unescaped, so my downstream model couldn't parse the JSON. 

It would be great if Tolgee provided additional variables like:
`key.description.escaped`
`source.translation.escaped`
that escapes double quotes, newlines, etc. for use in structured prompt templates.

Another simpler solution would be to just have a pre-baked default structural prompt (vs. the default "conversational prompt") and we just adapt our code to use whatever structure you decide. Less flexible, but more "easy button" and you all decided what structure and escaping are needed.

**Additional context**
Current workaround. Use YAML (which requires less escaping in text blocks) and use $$$ in the prompt and "re-indent" any newlines using those markers.

Image

Image

JanCizmar:
Closes #3683

## Problem

Custom AI prompt templates that build a structured payload (JSON or YAML) break as soon as a variable contains a quote or a newline, because values like `{{key.description}}` are substituted raw. The reporter is mapping Tolgee's prompt into an internal translation API that expects structured input, and had to work around it with YAML block scalars plus `$$$` markers and manual re-indentation.

## Solution

Register an `escapeJson` Handlebars helper, so template authors can write:

```json
{
  "context": "{{escapeJson key.name}} {{escapeJson key.description}}",
  "text": "{{escapeJson source.translation}}",
  "locale": "{{target.languageTag}}"
}
```

The helper emits the JSON-string-escaped value **without** surrounding quotes, so it embeds inside a larger quoted string. That output is also valid inside a double-quoted YAML scalar.

### Idempotence, and why there is no behavior change

Some variables are already JSON-escaped when collected — `translation` backs `source.translation`, `target.translation` and `other.<tag>.translation` — while `key.description`, `key.name` and `project.description` are raw. Introducing a general-purpose helper would have made `{{escapeJson source.translation}}` escape an already-escaped value and emit `\\"` where the author expects `\"`.

Rather than removing the pre-escaping (which would change what existing saved prompts render), pre-escaped values are now tagged with a `JsonEscapedString : Handlebars.SafeString` marker, and `escapeJson` returns such a value unchanged. The result:

- `{{source.translation}}` renders byte-for-byte as before — it was already emitted as a `SafeString`, so nothing about existing templates changes.
- `{{escapeJson <anything>}}` is always correct, on every variable. There is no subset of variables the author must remember not to wrap.

**No breaking change**, and no migration needed.

## Playground

`escapeJson` is offered in the editor autocomplete in the name slot only (never as an argument to itself, and never under a nested path like `{{key.`), and accepting it immediately opens the variable list. Hovering it shows a description instead of the previous "unknown variable" tooltip.

## Tests

- `PromptHandlebarsHelpersTest` (8) — escaping of quotes/backslashes/newlines/tabs, null and missing values, `SafeString` input, no HTML escaping, non-string values, and the two idempotence cases.
- `PromptControllerTest` (14) — renders a key description containing `"quotes"`, a newline and a backslash through `/prompts/run`; asserts `{{escapeJson source.translation}}` does not double-escape; asserts the default `translationInfo` fragment still escapes.
- `customPrompt.cy.ts` (16) — autocomplete offers the helper and chains into variables, the negative guards (argument position, nested path), the hover tooltip, and end-to-end escaping through the playground preview.

## Follow-up

Documentation for the AI playground needs a section on `escapeJson` (separate docs repo).

## Summary by CodeRabbit

* **New Features**
  * Added Handlebars helper support in the prompt editor, including helper autocomplete and dedicated tooltips for helper usage.
  * Introduced a JSON-escaping helper for prompt rendering to correctly escape quotes, newlines, tabs, and backslashes, and to handle empty/missing values.

* **Bug Fixes**
  * Prevented double-escaping by preserving already escaped content during prompt rendering.
  * Improved escaping/variable rendering behavior in JSON embedding scenarios.

* **Tests**
  * Expanded end-to-end and unit test coverage for helper suggestions, tooltips, and JSON escaping correctness.


jdimeo:
This is a fantastic approach! Even better than what I was suggesting. I think it will benefit many users in the future for many integration use cases.

PR #3800, opened 2026-07-10
fix: purge trashed keys that have translation suggestions

bdshadow:
Trashed keys were never permanently deleted. This PR fixes two independent causes and hardens the purge against a recurrence.

## 1. Missing suggestion cleanup (the trigger)

`KeyService.hardDeleteMultiple` / `hardDelete` deleted every child entity of a key **except** `translation_suggestion`. `TranslationSuggestion` has a non-nullable `@ManyToOne key: Key` (FK `translation_suggestion.key_id → key.id`, no cascade), so `DELETE FROM key` fails with a foreign-key violation once a key has any suggestion.

Confirmed in production Sentry (`delete from key ... violates foreign key constraint ... on table "translation_suggestion"`), 3,000+ occurrences, firing hourly: [TOLGEE-BACKEND-3TV](https://tolgee.sentry.io/issues/TOLGEE-BACKEND-3TV) / [TOLGEE-BACKEND-3TW](https://tolgee.sentry.io/issues/TOLGEE-BACKEND-3TW).

**Fix:** delete a key's translation suggestions in the hard-delete path (single + multiple), following the existing `deleteAllByProject` / `deleteAllByLanguage` pattern (OSS no-op, EE does the real delete). The EE impl chunks the id list (`IN_CLAUSE_BATCH_SIZE = 30_000`) so the `in :keyIds` bind list stays under PostgreSQL's 65535 parameter limit.

## 2. Purge wedged by one bad key (the amplifier)

`KeyTrashPurgeScheduler` deleted a batch of 100 keys in a **single transaction** and always re-fetched **page 0** (`PageRequest.of(0, BATCH_SIZE)`, ordered by `k.id`). So one key that fails to delete aborted the entire batch transaction (swallowed by `runSentryCatching`), the batch never advanced, and the purge stayed permanently stuck on the same lowest-id keys — freezing even healthy keys that shared the batch.

**Fix:**
- Page forward with an **id cursor** (`k.id > :afterId`) so every pass makes progress and terminates.
- Fall back to **per-key deletion** when a batch delete fails, so one un-deletable key can no longer block the rest. Individual failures are reported to Sentry and retried on the next run.

## Tests

`KeyTrashPurgeSchedulerTest`:
- `purges keys that have translation suggestions` — fails before fix with the exact FK `DataIntegrityViolationException`, passes after.
- `keeps purging when one key cannot be hard-deleted` — spies `KeyService` to make one key throw and asserts the healthy key in the same batch is still purged while the bad key survives (proves fix).
- The 5 existing purge tests remain green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Permanently deleting trashed keys now also deletes any linked translation suggestion records.
  - Trash cleanup is more resilient: if a batch hard-delete partially fails, healthy keys are still purged and failed ones are left for a later retry.

- **Improved Reliability**
  - Updated trash purge to use cursor-based processing to ensure all eligible items are handled consistently.

- **Monitoring**
  - Trash purge deletion errors are now reported to Sentry for easier troubleshooting.

- **Tests**
  - Added coverage for keys with suggestions and retry behavior after batch purge failures.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/service/key/KeyTrashPurgeScheduler.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/service/key/KeyTrashPurgeScheduler.kt
@@ -1,5 +1,6 @@
 package io.tolgee.service.key
 
+import io.sentry.Sentry
 import io.tolgee.component.CurrentDateProvider
 import io.tolgee.component.LockingProvider
 import io.tolgee.component.SchedulingManager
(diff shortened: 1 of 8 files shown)


backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationSuggestionServiceOssImpl.kt
```diff
@@ -20,4 +20,8 @@ class TranslationSuggestionServiceOssImpl : TranslationSuggestionService {
   override fun deleteAllByProject(id: Long) {
     // doing nothing
   }
+
+  override fun deleteAllByKeyIds(keyIds: Collection<Long>) {
+    // doing nothing
```
dkrizan: I understand this follow the existing pattern (`deleteAllByProject`, `deleteAllByLanguage`), but since `hardDelete` runs regardless of oss/ee, this should too. DB that had suggestions under an EE build and later runs an OSS build would hit exactly the FK violation this PR fixes.
Deleting leftover rows shouldn't be an EE feature.

backend/data/src/main/kotlin/io/tolgee/service/key/KeyTrashPurgeScheduler.kt
```diff
@@ -39,28 +40,56 @@ class KeyTrashPurgeScheduler(
...
+  }
+
+  private fun tryPurge(ids: List<Long>): Boolean {
+    try {
+      executeInNewTransaction(transactionManager) {
+        keyService.hardDeleteMultiple(ids)
+      }
+      return true
+    } catch (e: Exception) {
+      if (ids.size == 1) {
```
dkrizan: I see the intent (only report on the final per-key attempt), but the size-based condition is non-obvious and double-reports single-key batches.
I'd move the error handling into `purgeBatch`, so each catch block states its policy directly and the 1-key batch captures only once.

backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationSuggestionServiceOssImpl.kt
bdshadow: i guess, what you described is also valid for the existing `deleteAllByProject` and `deleteAllByLanguage`. Do you want to refactor it all? WDYT?

backend/data/src/main/kotlin/io/tolgee/service/key/KeyTrashPurgeScheduler.kt
bdshadow: done

backend/data/src/main/kotlin/io/tolgee/service/translation/TranslationSuggestionServiceOssImpl.kt
dkrizan: Exactly. I wouldn’t typically call for a fix in this PR, but since the title mentions “purge trashed keys that have translation suggestions,” the case I described is a real gap. 
If possible, I would refactoring it in this PR.
What is your opinion on this ? We can do it in follow-up PR too, if you prefer it more.

bdshadow: done

dkrizan (approved):

PR #3804, opened 2026-07-15
chore: migrate to Spring Boot 4

bdshadow:
## Summary

Migrates the Tolgee platform backend from **Spring Boot 3 to Spring Boot 4** (Spring Framework 7). This is a large, cross-cutting framework upgrade; opening as a **draft** while the test suite is brought fully green.

Core version bumps:

| Area | From → To |
|------|-----------|
| Spring Boot | 3.x → **4.1.0** |
| Spring Framework | 6 → **7** |
| Hibernate ORM | 6 → **7.4.1** |
| Spring Data | 3 → **4** |
| Spring Batch | 5 → **6** |
| Spring Security | 6 → **7** |
| Jackson | 2 (`com.fasterxml`) → **3 (`tools.jackson`)** |
| Redisson | 3.27 → **4.6.1** |
| Spring Cloud Azure | 5.x → **7.3.0** |
| Kotlin | → **2.3.21** |
| Gradle | → **9.6.0** |

## Notable migration work

- **Autoconfigure package splits** — Boot 4 splits the autoconfigure monolith into per-module packages (data-redis, hibernate, persistence, security, servlet, webmvc-test, restclient, liquibase, …); imports and excludes updated accordingly.
- **Jackson 2 → 3** — moved to `tools.jackson` coordinates and APIs; annotations remain on `com.fasterxml.jackson.annotation`.
- **Liquibase** — Boot 4 moved Liquibase auto-configuration to a separate `spring-boot-liquibase` module; added it so the custom `SpringLiquibase` beans run before the `EntityManagerFactory`.
- **Redisson 4.x** — extend `RedissonAutoConfigurationV4`, exclude the new unconditional V4 autoconfig, add `redisson-spring-cache` (split out of core), and declare `micrometer-core` explicitly (Redisson 4 marks it optional).
- **Hibernate 7 HQL** — added explicit `select` clauses to join queries that no longer infer a result type.
- **JSONB attributes** — restored JSON-based deep-copy for non-`Serializable` values (hypersistence-utils 3.15 dropped that fallback).
- **JPA auditing / startup ordering** — wire the date-time provider via `dateTimeProviderRef` and defer DB access out of bean construction, since Boot 4 creates servlet-filter beans before Liquibase migrates.

### Upstream bug found

Hibernate 7's `hibernate-models` crashes with `IndexOutOfBoundsException` on a JSONB entity attribute whose type is a `Map` subclass reaching `Map<K,V>` through a one-type-parameter generic intermediate. Reported upstream: https://github.com/hibernate/hibernate-models/issues/263 (worked around in `CompactSharedMap`).

## Status

- ✅ Full backend compiles (all modules)
- ✅ Application boots end-to-end (context loads, Liquibase migrates, web server starts, batch jobs execute)
- ✅ Core integration tests green (e.g. `BatchJobsGeneralWithoutRedisTest`)
- ⏳ Widening test-suite coverage; the Redis-backed concurrent batch tests (`BatchJobsGeneralWithRedisTest`, `BatchJobRecoveryTest`, …) are pre-existing flaky/timing-sensitive tests still being verified

## Related

The matching `billing` branch (`bdshadow/spring-boot-4`) carries the corresponding billing changes.


backend/data/src/main/kotlin/io/tolgee/service/project/ProjectHardDeletingService.kt
```diff
@@ -209,6 +213,29 @@ class ProjectHardDeletingService(
     }
   }
 
+  private fun deleteTasks(projectId: Long) {
```
dkrizan: It’s strange that this fundamental thing was missing here
I guess since the hard delete feature is running in the background, we haven’t discovered that projects with tasks are not being deleted.

Anyway, it would be great to move it to TaskService.

(nitpick, this goes against the title of the PR, which is “chore,” but we can probably overlook it.)

bdshadow: i also was thinking about it, however, decided to leave it as it is, and stay consistent. Check other delete statements - they all live here

dkrizan: Well, some are, but those were all added recently in 89996591a, and for a specific reason - that commit made the project delete a bulk JPQL delete, so orphanRemoval children had to be deleted by hand. The task isn't one of those, it feels closer to languages/keys/labels, which go through their services.

I'll leave it up to you. I just wanted to share my opinion.

bdshadow: done.
But it required much bigger refactoring and a fix of the issue: https://github.com/tolgee/tolgee-platform/pull/3804/commits/02b65d476f9ad0736d364fe626faeb1fd698cdfd

Anty0: > we haven’t discovered that projects with tasks are not being deleted

@dkrizan Yeah, it's a recurring theme. It triggers an error in Sentry (usually a constraint violation). Unfortunately, the backlog of these is so large that I probably won't recognize it the first time it fires, and it waits until someone gets to triage the bugs there. Unfortunately, we would need another such long cooldown to get to things like cleaning up Sentry (and it kinda needs it a lot). Maybe we can make it a shape-up thing - pitch to go through Sentry and make sure it's clean (or as clean as possible). Or maybe I can try to make a skill to identify and fix some clear ones automatically (if the agent can both reproduce it and validate that the fix prevents the crash, it could be pretty much solved case without much additional work).

Anty0 (dismissed):
I went through the PR and nothing jumps out to me. I had a look at the migration guide for Spring Boot and other libraries, but going through it properly and verifying all was done would take a while - let me know if I should do that, but otherwise the changeset looks fine.

JanCizmar:
I completely missed this. Such a big updates should not be released as chore, but as fix, so we can revert it. Please next time, make it "fix: " 🙏

PR #3765, opened 2026-06-24
feat: Community Translation v1

Anty0:
Integration branch for **Community Translation v1.0** (#3763). Individual lanes are reviewed as sub-PRs that target this branch and squash-merge into it; this PR merges to `main` once the v1.0 must-haves have landed.

## Landed lanes
- ✅ **Public project toggle** — `public` flag on `Project` (`is_public` column + Liquibase migration), exposed on `ProjectModel`/`ProjectDto`; owner-only `PUT /v2/projects/{projectId}/publishing` (gated by `PROJECT_EDIT` + `checkUserIsOwnerOrServerAdmin`, activity-logged, cache-evicting); confirmation-guarded "Public project" switch in advanced settings.
- ✅ **Community permission** (#3770) — authenticated users on a public project get a computed community floor (read subset of VIEW + `translations.suggest` + `translation-comments.add`, all languages; origin `COMMUNITY`) without a stored `Permission` row, so they contribute without taking a paid seat. Direct/org permissions and admin/supporter elevation layer on top, so no authenticated user is below a non-member; anonymous users get nothing. Member emails + member count are redacted behind `MEMBERS_VIEW` (REST + websocket); out-of-surface endpoints are gated (MT credits → `ORGANIZATION_QUOTAS_VIEW`, import-settings → `TRANSLATIONS_EDIT` (mirrors the import menu gate `translations.edit` + `keys.view`), branch-merge → `BRANCH_MANAGEMENT`, blocking-tasks → task-view-or-assigned).

- ✅ **List-row public indicator** (#3771) — public projects show a second line under the project name in the shared list row (`DashboardProjectListItem`): an outlined `public` badge + organization name. Adds `public` to `ProjectWithStatsModel` + assembler. Sole indicator (no project-detail icon).
- ✅ **Suggestion resolution split** (#3776) — splits the single suggestion Accept into "Accept only" (`declineOther=false`, leaves the competing suggestions active) and "Accept and decline others" (`declineOther=true`, prior behavior). Inline row actions are ✓ Accept-only and ✕ Decline; an `⋮` overflow menu holds Accept-and-decline-others and Delete (own suggestions only). Frontend-only — the backend already supported `declineOther`. Optimistic cache keeps the surviving sibling visible until refetch (vitest-covered). Two new i18n keys created in "Tolgee itself".
[...]

PR #3808, opened 2026-07-17
fix: community access to public-project organizations

Anty0:
Re-lands the organization-permissions lane of Community Translation v1, which is missing from `main`.

## The problem

Opening a public project partially fails. The frontend sets the preferred organization to the project's owner org on every project open (`ProjectContext.tsx` → `PUT /v2/user-preferences/set-preferred-organization/{id}`), and that call is rejected for community users.

`main` still has the strict check in `UserPreferencesController`:

```kotlin
organizationRoleService.checkUserCanView(organization.id)
```

which resolves to `OrganizationRepository.canUserView`, whose predicate is `(perm is not null or mr is not null)` — it requires a **stored** `Permission` or `OrganizationRole` row. The community floor from #3770 is *computed* by design, so contributors don't consume a paid seat. No row → `PermissionException(USER_CANNOT_VIEW_THIS_ORGANIZATION)` → 403.

## How it went missing

#3792 was merged into `jirikuchynka/public-projects` on 2026-07-13 (merge commit `93e6fd9cc`). That branch was subsequently rebased and force-pushed; the rebase dropped the merge commit along with its two payload commits. #3765 then squash-merged the rebased branch into `main` as `4a4ea3f55`, shipping v1 without this lane. GitHub still reports #3792 as MERGED, which is why it went unnoticed.

## This PR

`git cherry-pick -m 1 93e6fd9cc` onto `main` — applied with no conflicts, and the resulting diff against `main` is identical to the original payload (55 files, 2197+/129-).

Contents (unchanged from #3792):

- `OrganizationRoleService.canUserViewStrictOrPublic` / `checkUserCanViewOrPublic` — the org view floor: strict check OR `projectService.hasPublicProjects(organizationId)`
- Applied at `UserPreferencesController:61` and `OrganizationAuthorizationInterceptor:75`
- `PreferredOrganizationFacade` heals stale/missing preferences in place and emits `PrivateOrganizationModel.limitedView` for below-member viewers
- Scoping of org language reads to accessible projects for below-member viewers
- Glossary guest access, member-listing and usage exclusions for floor viewers

## Verification

- `:data:`/`:api:`/`:security:`/`:development:compileKotlin` — BUILD SUCCESSFUL
- `UserPreferencesSetPreferredOrganizationTest` — 7/7 passed
- `OrganizationFloorAccessTest` — 29/29 passed
- webapp + e2e `eslint --fix` and `tsc` — clean, no files changed

## Summary by CodeRabbit

* **New Features**
  * Community and below-member viewers now receive organization models tailored to public-project access, including a new `limitedView` flag and safer subscription-related visibility behavior.
* **Bug Fixes**
  * Preferred-organization and “managed-by” responses now respect updated viewability rules and permission filtering, with stronger stale-preference healing when the stored organization becomes unavailable.
  * Organization language/project listing queries are now restricted to only the projects accessible to the current caller.
  * Authorization “view floor” enforcement was tightened to allow the correct default endpoints while keeping role-required routes protected.

### DIFF
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationLanguageController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationLanguageController.kt
@@ -5,11 +5,9 @@ import io.swagger.v3.oas.annotations.tags.Tag
 import io.tolgee.dtos.cacheable.OrganizationLanguageDto
 import io.tolgee.hateoas.language.OrganizationLanguageModel
 import io.tolgee.hateoas.language.OrganizationLanguageModelAssembler
-import io.tolgee.security.authentication.AuthenticationFacade
 import io.tolgee.security.authorization.UseDefaultPermissions
 import io.tolgee.service.language.LanguageService
-import io.tolgee.service.organization.OrganizationRoleService
-import io.tolgee.service.project.ProjectService
+import io.tolgee.service.security.SecurityService
 import org.springdoc.core.annotations.ParameterObject
 import org.springframework.data.domain.Pageable
 import org.springframework.data.domain.Sort
@@ -32,22 +30,8 @@ class OrganizationLanguageController(
   private val languageService: LanguageService,
   private val organizationLanguageModelAssembler: OrganizationLanguageModelAssembler,
   private val pagedOrganizationLanguageAssembler: PagedResourcesAssembler<OrganizationLanguageDto>,
-  private val authenticationFacade: AuthenticationFacade,
-  private val organizationRoleService: OrganizationRoleService,
-  private val projectService: ProjectService,
+  private val securityService: SecurityService,
 ) {
-  private fun accessibleProjectIds(
-    organizationId: Long,
-    requested: List<Long>?,
-  ): List<Long>? {
-    val user = authenticationFacade.authenticatedUser
-    if (organizationRoleService.canUserViewAtLeastMember(user, organizationId)) {
-      return requested
-    }
-    val accessible = projectService.getBelowMemberAccessibleProjectIds(organizationId, user.id).toSet()
-    return requested?.filter { it in accessible } ?: accessible.toList()
-  }
-
   @Operation(
     summary = "Get all languages in use by projects owned by specified organization",
     description = "Returns all languages in use by projects owned by specified organization",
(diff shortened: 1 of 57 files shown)


webapp/src/component/layout/BaseSettingsView/SettingsMenuItem.tsx
```diff
@@ -34,6 +34,7 @@ type Props = {
   selected?: boolean;
   matchAsPrefix?: boolean;
   hidden?: boolean;
+  dataCyItem?: string;
```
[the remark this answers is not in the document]
Anty0: data-cy needs to be static string, it cannot be dynamically selected - when selector needs to be passed through components or it needs to depend on dynamic propery, then this is a proper way to do this - use static data cy and pass dynamic selector like data cy item.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationLanguageController.kt
```diff
@@ -29,7 +32,22 @@ class OrganizationLanguageController(
   private val languageService: LanguageService,
   private val organizationLanguageModelAssembler: OrganizationLanguageModelAssembler,
   private val pagedOrganizationLanguageAssembler: PagedResourcesAssembler<OrganizationLanguageDto>,
+  private val authenticationFacade: AuthenticationFacade,
+  private val organizationRoleService: OrganizationRoleService,
+  private val projectService: ProjectService,
 ) {
+  private fun accessibleProjectIds(
```
dkrizan: I would put it in a service (maybe SecurityService?).

backend/data/src/main/kotlin/io/tolgee/model/enums/OrganizationRoleType.kt
```diff
@@ -1,5 +1,9 @@
 package io.tolgee.model.enums
 
+/**
```
dkrizan, line 3: I believe this is outdated, as it will be fixed in this https://github.com/tolgee/tolgee-platform/pull/3807. If so, there’s no need for this comment.

Anty0: Not the case unfortunately. The #3807 only changes cache behavior. The `OrganizationRoleType` is stored as ordinal in DB. We can change that, but that would require migration and some other things changed.

dkrizan (approved):

PR #3810, opened 2026-07-21
fix: hard-delete of a project with a default namespace fails on FK constraint

bdshadow:
## Root cause

`ProjectHardDeletingService.hardDeleteProject` nulls the project's `baseLanguage` before deleting the project's languages, but never nulls `defaultNamespace` before deleting the project's namespaces. So for any project that has a default namespace, deleting its namespaces fails:

```
ERROR: update or delete on table "namespace" violates foreign key constraint
"FK4pgl5uneh3it2odrc36nyubbx" on table "project"
  Detail: Key (id)=(...) is still referenced from table "project".
  at io.tolgee.service.key.NamespaceService.deleteAllByProject
```

## Impact

`hardDeleteProject` is used by both:
- the async `OnProjectSoftDeleted` listener (normal project deletion), and
- `DeletedOrganizationProjectPurgeScheduler` (purging projects of soft-deleted organizations, added in #3783).

The exception is caught and rethrown into the async/scheduled error handler, so it fails silently. In production the purge scheduler was running daily (visible via the "Running orphan project purge" INFO log) but always threw here before deleting anything — so projects of deleted organizations kept their `deleted_at` set and were never purged.

## Fix

Null `project.defaultNamespace` alongside `project.baseLanguage` before the flush that precedes namespace deletion.

## Test

`ProjectHardDeletingServiceTest.deletes project that has a default namespace` — reproduces the FK violation without the fix (verified: fails with `DataIntegrityViolationException`), passes with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hard deletion for projects that have a default namespace configured.
  * Ensures projects are fully removed without keeping stale references to related language or namespace data.
* **Tests**
  * Added a new test case covering hard-deleting a project with a default namespace.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/service/project/ProjectHardDeletingServiceTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/service/project/ProjectHardDeletingServiceTest.kt
@@ -19,6 +19,7 @@ import io.tolgee.dtos.BigMetaDto
 import io.tolgee.dtos.RelatedKeyDto
 import io.tolgee.fixtures.waitFor
 import io.tolgee.model.Project
+import io.tolgee.model.key.Namespace
 import io.tolgee.service.bigMeta.BigMetaService
 import io.tolgee.testing.assert
 import io.tolgee.util.executeInNewRepeatableTransaction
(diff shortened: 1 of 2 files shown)


bdshadow:
### Production evidence

This throws on every scheduled run in prod (`tolgee-prod`, 2026-07-20 14:33:55, pod `tolgee-prod-c4c9c4d9-rmbxp`), right after the scheduler's `Running orphan project purge` log:

```
ERROR: update or delete on table "namespace" violates foreign key constraint
"FK4pgl5uneh3it2odrc36nyubbx" on table "project"
  Detail: Key (id)=(1009600001) is still referenced from table "project".
org.springframework.dao.DataIntegrityViolationException: JDBC exception executing SQL [
    delete from namespace where project_id = ?
] ... constraint [FK4pgl5uneh3it2odrc36nyubbx]
    at io.tolgee.service.key.NamespaceService.deleteAllByProject(NamespaceService.kt:162)
    ...
ERROR o.s.s.s.TaskUtils$LoggingErrorHandler : Unexpected error occurred in scheduled task
```

`runSentryCatching` captures this before rethrowing, so it is also in Sentry (surfaced as a `DataIntegrityViolationException` from the scheduled task, not under the scheduler class name). Because it's swallowed by the scheduled-task error handler, the purge silently deletes nothing — projects of soft-deleted organizations keep `deleted_at` set indefinitely.

bdshadow:
issue in sentry: TOLGEE-BACKEND-3TW

dkrizan (approved):

PR #3813, opened 2026-07-21
feat: auto-invalidate caches by value shape instead of wiping on startup

JanCizmar:
## Why

Cleaning the whole cache on every startup made frequent deploys slow: each restart flushed the Redis cache and hammered the DB with a cold-cache thundering herd. That wipe only ever existed to drop cached entries whose serialized shape changed between versions.

## What

Namespace every cache by a **structural fingerprint** of the type it stores. When a cached type's shape changes, the fingerprint changes, the physical cache name changes, and stale entries are simply never read again — they expire by TTL. Only the caches whose shape actually changed go cold, and rolling deploys coexist because each running version reads its own namespace.

- **`CacheValueFingerprint`** — deterministic structural hash of a type graph (expands `io.tolgee.*` types into sorted `property → type` signatures, treats JDK/stdlib types as opaque leaves, includes enum constants and generic arguments, cycle-guarded).
- **`CacheFingerprintRegistry`** — discovers `@Cacheable`/`@CachePut` caches and their value types at startup; resolves physical names for direct-access caches from the type supplied at the call site.
- **`FingerprintingCacheManager`** — decorator that rewrites `getCache(name)` → `name--<fp>`, idempotent on already-physical names.
- **`TolgeeRedissonSpringCacheManager`** — applies the default TTL to every cache via `createDefaultConfig()`. This removes the need to enumerate cache names and fixes caches (e.g. `batchJobs`, `rateLimits`) that were previously created with **no TTL** because they were missing from the hand-maintained list.

Per-key eviction on deserialization failure (`TolgeeCacheErrorHandler` / `ResilientCacheAccessor`) stays as the backstop for entries that fail to decode outright.

## Removed

The hand-maintained cache registry that had drifted from reality: `AllCachesProvider`, `AdditionalCachesProvider`, `Caches.caches`, the startup `CacheCleaner`, and the `clean-on-startup` property.

## Tests

- Unit: fingerprint core (structural sensitivity, determinism, enums/generics/nullability), registry idempotency + unknown-name passthrough, manager routing, Redisson default-TTL.
- Redis integration: a shape change routes to a fresh physical cache with **no cross-version reads** (auto-invalidation + rolling-deploy coexistence), and a genuinely undecodable entry heals via per-key eviction.

## Note

Requires the matching branch `jancizmar/no-cache-clean-on-startup` in `tolgee/billing` (direct-access billing caches thread their value type through the changed `CacheWithExpirationManager` signature).

## Summary by CodeRabbit

* **New Features**
  * Cache entries are now isolated by the detected structure of their values, preventing cross-version reads.
  * Added automatic healing when cached data is corrupted or can’t be deserialized.
  * Cache managers now apply the configured default TTL to on-demand caches.
* **Bug Fixes**
  * Improved resilience for cache reads, reducing impact from stale/invalid cached values.
* **Chores / Documentation**
  * Removed the startup cache-cleaning configuration option.


backend/data/src/main/kotlin/io/tolgee/component/cache/FingerprintingCacheManager.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.cache.Cache
+import org.springframework.cache.CacheManager
+
+class FingerprintingCacheManager(
+  private val delegate: CacheManager,
+  private val registry: CacheFingerprintRegistry,
+) : CacheManager {
+  override fun getCache(name: String): Cache? = delegate.getCache(registry.physicalName(name))
+
+  override fun getCacheNames(): Collection<String> = delegate.cacheNames
```
bdshadow: i think, here it would be more correct to "repllicate" what parent `CacheManager` does and to return the logical names. It will stop leaking the fingerprints into everything that enumerates caches - Micrometer metrics, the actuator caches endpoint, logs - which otherwise showed `userAccounts--a1b2c3`

backend/data/src/main/kotlin/io/tolgee/component/cache/CacheFingerprintRegistry.kt
```diff
@@ -0,0 +1,146 @@
...
+  private var fingerprintByCacheName: Map<String, String>? = null
+
+  override fun afterSingletonsInstantiated() {
+    val built = buildFingerprints()
+    fingerprintByCacheName = built
+    logger.info("Fingerprinted caches: {}", built.keys.sorted())
+  }
+
+  fun physicalName(cacheName: String): String {
+    if (cacheName.contains(SEPARATOR)) return cacheName
```
bdshadow: nitpick, i'll leave it on you. Just to totally avoid that it can break on "--" in the cache name, we can add a regex:
```kt
val PHYSICAL_NAME_SUFFIX = Regex("$SEPARATOR[0-9a-f]+$")
```

and use it here as 
```kt
 if (PHYSICAL_NAME_SUFFIX.containsMatchIn(cacheName)) return cacheName
```

backend/data/src/test/kotlin/io/tolgee/unit/cache/CacheFingerprintRegistryTest.kt
```diff
@@ -0,0 +1,175 @@
...
+    fun b(): ShapeB = ShapeB(1, 2)
+  }
+
+  class CachePutBean {
+    @CachePut("annotatedCache")
+    fun get(): ShapeA = ShapeA(1)
+  }
+
+  class CachingBean {
+    @Caching(cacheable = [Cacheable("annotatedCache")])
```
bdshadow, line 51: also makes sense to add a test for the CachePut variant of Caching
```kt
  class CachingPutBean {
    @Caching(put = [CachePut("annotatedCache")])
    fun get(): ShapeA = ShapeA(1)
  }
```

bdshadow (changes requested):
really nice feature.

Some nitpciking comments. Requesting changes mainly because of the comment on `getCacheNames` method

backend/data/src/main/kotlin/io/tolgee/component/cache/FingerprintingCacheManager.kt
JanCizmar: Good catch — done. `getCacheNames()` now maps back to logical names, so metrics/actuator/logs no longer see the `--<fp>` suffix.

backend/data/src/main/kotlin/io/tolgee/component/cache/CacheFingerprintRegistry.kt
JanCizmar: Adopted — added `PHYSICAL_NAME_SUFFIX = Regex("$SEPARATOR[0-9a-f]+$")` and use it both for the already-physical guard and to strip the suffix in `logicalName`.

backend/data/src/test/kotlin/io/tolgee/unit/cache/CacheFingerprintRegistryTest.kt
JanCizmar: Added `CachingPutBean` with `@Caching(put = [CachePut("annotatedCache")])` and a test asserting it gets fingerprinted.

bdshadow (approved):

PR #3816, opened 2026-07-22
fix: let users without an organization open public projects

Anty0:
Follow-up to #3808 (merged). That PR fixed organization permissions for public projects on the backend; this one fixes the frontend bootstrap that still left the target user stuck.

## Problem

A signed-in user with **no organization** — organization creation disabled on the server, or an SSO-provisioned account — was stuck on the "No permissions" page and could not open a public project, even from a direct link.

## Root cause

`RequirePreferredOrganization` wrapped the whole authenticated route tree, so `ProjectsRouter` never mounted. `ProjectContext` — the only thing that adopts the project's owning organization as the preferred one — therefore never ran.

## Fix

Move the `/projects` routes above the terminal gate and re-apply the gate to the `/projects/add` leaf inside `ProjectsRouter`: viewing a project is reachable without a preferred organization, creating one still requires it.

Supporting changes that fall out of the route move:

- `DashboardRouter` extracted from `RootRouter` so the gated and ungated arms are declared in one place, and `HelpMenu` has a single structural mount.
- `HelpMenu` guard changed from `!preferredOrganization` to `!user` — it must render for a signed-in user who has no organization, and must not render for anonymous visitors.
- `OrganizationSwitch` / `BaseProjectView` handle the no-preferred-organization state instead of rendering an empty switch or a dangling breadcrumb chevron. On `/community-projects` the surface label still renders; the dropdown (which would open onto an empty list) does not.
- Project creation permission consolidated behind one `useCanCreateProject`, mirroring the server rule in `OrganizationRoleService.checkUserCanCreateProject` (owner or maintainer of the preferred organization, or a server `ADMIN`). It is shared by the add button on the project list, that view's empty-state shortcut, and the submit button on `/projects/add`. A user parked on an organization they cannot create in reaches the form only by typing the URL — it now renders with the submit disabled and the organization switcher still usable, instead of accepting input that would 403 on submit.

  One behaviour change on the project list follows, outside the organization-less persona this PR is about: a `SUPPORTER` viewing an organization they are not a member of **loses** the add button and empty-state shortcut, because the server refuses them — `isAdmin()` is `role == ADMIN`, so their create always 403'd. Server admins are unaffected. `isAdminAccess` is left exactly as it was in `ProjectListView`: it drives the admin-access banner, which is a different question from who may create.
- The disabled submit is accompanied by an explanation, so the state is not silent. Two messages, picked by whether an organization is selected at all: `project_create_no_permission_message` names the organization, and `project_create_no_organization_message` covers the state where there is none to name (reachable because `RequirePreferredOrganization` passes an email-unverified user straight through). Without it this traded a late-but-explicit 403 for an early-and-silent dead button — worst for the organization-less user, who on a server with `userCanCreateOrganizations=false` also sees an empty switcher popover with its "add organization" entry hidden. Both come from one `submitDisabledReason` prop on `BaseFormView`, which renders the reason and disables the button, so the two cannot fall out of step.
- The organization switcher on `/projects/add` is no longer restricted to organizations the user owns. `filterCurrentUserOwner` maps to `roleType = OWNER` server-side, so a maintainer — who may create — could not reach their own organization from the page that had just told them to switch.
- `StandardForm` blocks submission itself when the submit is refused, rather than relying on the button's `disabled` attribute. Callers using `formId` with their own submit button bypass that attribute entirely, so the gate belongs on the Formik submit path.
- `ProjectCreateView` no longer keeps `organizationId` in Formik state, taking it from the preferred organization at submit time instead. `StandardForm` sets `enableReinitialize`, so an id in `initialValues` meant that using the organization switcher — now the prescribed way out of a disabled submit — reset every field the user had typed.
- Support-chat entitlement consolidated behind one `useHasSupportChat`, shared by `useChatwoot` and `useIntercom`, and made viewer-scoped: a community contributor parked on someone else's organization no longer inherits that organization's paid support widget. Previously `setChatwootAttributes` would have pushed the host organization's plan, subscription status and enabled features to Chatwoot on behalf of a non-member.
- `useIntercom` now shuts the messenger down when a booted session becomes de-entitled mid-navigation — reachable now that the preferred organization changes as a side effect of ordinary navigation.

Preferred organization stays synchronized with the selected project; that is deliberate, since organization settings and glossaries need a selected organization.

## Backend fix folded in

Reviewing the client rule surfaced a defect in `ProjectsController.createProject`. It looked the role up with `organizationRoleService.getType(dto.organizationId)` — the *throwing* accessor — to decide whether to grant a maintainer full access, and that call happens after `ProjectCreationService.createProject` has already committed its own transaction. `getType` raises `USER_IS_NOT_MEMBER_OF_ORGANIZATION` for a non-member, so a server `ADMIN` creating a project in an organization they do not belong to passed the permission check, had the project persisted, and *then* received a 403 — leaving it behind in a stranger's organization with nothing in the UI indicating it existed.

The lookup only needs the role, so it wants the nullable `findType`. Swapped, with a regression test asserting the create succeeds and that no full-access permission is granted. Unrelated to organization-less public-project access, but it is the rule the client mirrors, and mirroring a broken rule was the alternative.

## Tests

- Backend: two new cases in `PreferredOrganizationCommunityTest` covering an organization-less user adopting an organization via a public project, and failing to adopt one without public projects (7/7 pass). `ProjectsControllerCreateTest` gains a case pinning the `getType` → `findType` fix.
- Unit: `organizationEntitlement` pins the support-tier and limited-view rules (mutation-checked — dropping either support tier or the limited-view conjunct turns the suite red). `organizationRole` pins the role predicates the create rule is built from.
- E2E: `communityPreferredOrganization.cy.ts` extended with an organization-less persona (server configured with `userCanCreateOrganizations=false`) covering cold deep-link adoption, reload persistence, switch-failure resilience, community project listing, and the create-project/create-organization walls; plus a case asserting the create-project submit is disabled for an organization-less user after they adopt a foreign organization. A second case covers the member persona end to end: submit disabled on the foreign organization, typed input surviving an organization switch, submit re-enabled, and the created project landing in the switched-to organization. A third adds a `SUPPORTER` persona (`supporterCommunityUser`) who owns no organization and is a plain `MEMBER` of one. It is checked in both positions — parked on an organization it only supports, and switched to the one it belongs to — so the add button and the create submit stay refused in each. That combination is what makes the `ADMIN`-vs-`ADMIN`-or-`SUPPORTER` distinction fail loudly if it regresses, and because the user owns nothing, the switch itself only succeeds while the switcher is not restricted to owned organizations. A fourth persona (`dualOrgCommunityUser`) owns one organization and is a plain `MEMBER` of another — the only shape that can cross from may-create to may-not — and its case delays the preference PUT to assert the submit is blocked while the switch is still in flight, which is the window in which a stale answer would create the project in the organization being left.

## Notes for review

Support chat is now scoped to the viewer's own relationship with the organization, via the `limitedView` flag from `PrivateOrganizationModel`. Who keeps it:

| Viewer | `limitedView` | `activeCloudSubscription` | Chatwoot | Intercom |
|---|---|---|---|---|
| Organization member (any role) | `false` | present | yes | yes |
| Project-level member, no organization role | `false` | null | yes | no (also no on `main`) |
| Admin / supporter | `false` | present | yes | yes |
| Public-project viewer only | `true` | null | **no** (changed) | no |

`limitedView` is `!isSupporterOrAdmin && !canUserViewStrict`, and `canUserViewStrict` counts direct project permissions as well as organization roles — so the only group that loses support chat is a viewer whose *sole* access route is public-project visibility. Everyone who has the widget on `main` keeps it.

The Intercom column is not a change from this PR: `activeCloudSubscription` is nulled unless `isAtLeastMember` (organization role only), and it is populated only by the cloud billing module, so Intercom could never boot for below-member or self-hosted users. This PR makes `useIntercom`'s menu-entry condition match the boot condition it already had, which removes a menu entry that rendered but did nothing on self-hosted instances.

`limitedView` is applied to support chat only. QA checks, tasks and translation memory still follow the content owner's entitlement, which is the intended behaviour — those are properties of the project being translated, whereas support chat is about the viewer's own relationship with Tolgee.

## Summary by CodeRabbit

* **New Features**
  * Community “preferred organization” adoption now works reliably for org-less users through public projects, with preferences preserved across sessions.
  * Help/support UI now appears conditionally based on sign-in and support eligibility.
  * Project creation UI now reflects organization preference and permission rules more directly.
* **Bug Fixes**
  * Improved access control around viewing/adopting preferred organizations and preventing unauthorized adoption or repeated updates.
  * Organization switch and routing behavior were tightened when no preferred organization is available.
* **Tests**
  * Expanded Cypress coverage for community preferred-organization and org-less scenarios.
  * Added backend and UI entitlement/helper tests to validate edge cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ e2e/cypress/integration/examples/
 /webapp/.DS_Store
 /webapp/.env.local
 /webapp/.env.development.local
+/webapp/.env.e2e
 /webapp/.env.test.local
 /webapp/.env.production.local
 /webapp/src/eeSetup/eeModule.current.tsx
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/BusinessEventController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/BusinessEventController.kt
@@ -6,9 +6,10 @@ import io.swagger.v3.oas.annotations.tags.Tag
 import io.tolgee.component.reporting.BusinessEventPublisher
 import io.tolgee.dtos.request.BusinessEventReportRequest
 import io.tolgee.dtos.request.IdentifyRequest
-import io.tolgee.exceptions.AuthenticationException
 import io.tolgee.openApiDocs.OpenApiHideFromPublicDocs
+import io.tolgee.security.authentication.AuthenticationFacade
 import io.tolgee.service.organization.OrganizationRoleService
+import io.tolgee.service.project.ProjectService
 import io.tolgee.service.security.SecurityService
 import io.tolgee.util.Logging
 import io.tolgee.util.logger
@@ -27,6 +28,8 @@ class BusinessEventController(
   private val businessEventPublisher: BusinessEventPublisher,
   private val securityService: SecurityService,
   private val organizationRoleService: OrganizationRoleService,
+  private val projectService: ProjectService,
+  private val authenticationFacade: AuthenticationFacade,
 ) : Logging {
   @PostMapping("/report")
   @Operation(summary = "Reports business event")
@@ -34,13 +37,16 @@ class BusinessEventController(
     @RequestBody eventData: BusinessEventReportRequest,
   ) {
     try {
-      eventData.projectId?.let { securityService.checkAnyProjectPermission(it) }
-      eventData.organizationId?.let { organizationRoleService.checkUserCanView(it) }
-      businessEventPublisher.publish(eventData)
+      val projectId = eventData.projectId?.takeIf { securityService.hasAnyProjectPermission(it) }
+      val owningOrganizationId = projectId?.let { owningOrganizationId(it) }
+      val organizationId =
+        eventData.organizationId
+          ?.takeIf { canReportOrganization(it) }
+          ?.takeIf { agreesWithReportedProject(it, owningOrganizationId) }
+      val backedClaims = eventData.copy(projectId = projectId, organizationId = organizationId)
+      logDroppedClaims(eventData, projectId, organizationId)
+      businessEventPublisher.publish(backedClaims)
     } catch (e: Throwable) {
-      if (e is AuthenticationException) {
-        return
-      }
       logger.error("Error storing event", e)
       Sentry.captureException(e)
     }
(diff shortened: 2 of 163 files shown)


Anty0:
Superseded by #3883

PR #3829, opened 2026-07-29
fix: audit existing exception handling done by matching error message

Anty0:
## Initial problem

`GET /v2/projects/{id}/keys?sort=<unknown>` answers **500** in production.

`ExceptionHandlers` already meant to turn this into `400 unknown_sort_property`, but it recognised the failure by matching the ORM's message text:

```kotlin
ex.message?.contains("could not resolve property", true)
```

That is the Hibernate 5 wording. Hibernate 6 emits `Could not resolve attribute 'keyName' of 'io.tolgee.model.key.Key'`, so the guard stopped matching at the Hibernate 6 upgrade and quietly stopped working — no compile error, no failing test, just 500s.

Because that failure mode is invisible by construction, this branch also audits every other place in `tolgee-platform` and `billing` that decides what to do by looking at an error message.

## The fix

The wording is gone from the guard; Hibernate 6 throws typed exceptions for this:

```kotlin
private fun Throwable.isUnresolvablePath() = this is PathElementException || this is PathException
```

Both terms are load-bearing, and the tests prove it rather than a comment asserting it: `PathElementException` is an `IllegalArgumentException` and `PathException` a `SemanticException`, so neither implies the other. Deleting `PathException` and re-running `KeyControllerTest` reddens `returns bad request when sorting dereferences a basic property` (`?sort=name.nope`), which is how that second term earned its place.

The handler also requires the client to have actually asked for a sort:

```kotlin
val sortedByClient = request.getParameter("sort") != null
if (sortedByClient && ExceptionUtils.getThrowableList(ex).any { it.isUnresolvablePath() }) {
  logger.debug("Unresolvable property in a query", ex)
  return ResponseEntity(ErrorResponseBody(Message.UNKNOWN_SORT_PROPERTY.code, null), HttpStatus.BAD_REQUEST)
}
Sentry.captureException(ex)
throw ex
```

Without that condition the handler classifies by exception type alone, and only the `?sort=` parameter is user-controlled: an unresolvable path in one of our own repository queries is a **server** bug wearing the same exception, and it would be answered as a client `400` and reported nowhere. With the condition, Sentry still receives every case the client did not cause.

## A second dead guard, in the same file

`handleQueryException` sat directly below the one this PR started with and had rotted further:

```kotlin
if (ex.message!!.contains("could not resolve property")) {
```

No `ignoreCase`, and Hibernate 6 capitalises it — `Could not resolve property: x of: y` — so it could never match. That string is produced only by `AbstractPropertyMapping`, which nothing in hibernate-core references any more, so the guard is unreachable even case-insensitively. Every `QueryException` fell through to `throw ex` and answered 500, exactly as with no handler at all. `ex.message!!` would also NPE on a null-message exception.

Removed, along with `Message.COULD_NOT_RESOLVE_PROPERTY`, which had no other references. The generated API schema is regenerated accordingly (exactly two deletions).

## Guards this audit converted

Each literal was checked against the version actually on the classpath (hibernate-core 7.4.1, spring-web 7.0, azure-storage-common 12.30.1, JDK 21) rather than from memory.

| Site | Was | Now |
|---|---|---|
| `handleInvalidDataAccessApiUsage` | `contains("could not resolve property")` | `PathElementException` / `PathException` in the cause chain, and only for client-supplied sorts |
| `FrontendUrlProvider` | `contains("No current ServletRequestAttributes")` | `RequestContextHolder.getRequestAttributes() !is ServletRequestAttributes` — the precondition Spring itself asserts |
| `AzureFileStorageFactory` | `message == "Invalid connection string."` | `try` scoped to the `connectionString()` call, where any `RuntimeException` means exactly that (the SDK reports an unparsable `BlobEndpoint` as a `RuntimeException` wrapping a `URISyntaxException`). A missing connection string and an unparsable endpoint now answer `invalid_connection_string` instead of `cannot_create_azure_storage_client`; both are 400 |
| `PostgresAutoStartConfiguration` | `contains("the database system is starting up")` | `SQLException.sqlState == "57P03"`, which the server never localizes |
| `SetKeysNamespaceChunkProcessor` | `getRootCauseMessage(e).contains("key_project_id_name_namespace_id_idx")` | `PSQLException.serverErrorMessage.constraint` against the names `schema.xml` creates |
| EE licensing (`catchingOutOfCredits` and `releaseSubscription`) | `HttpClientErrorException.message` | the parsed body, via a shared `parseTolgeeErrorBody` |
| billing `createOrReuseSchedule` | regex `sub_sched_[\w-]+` scraped out of the error text | `Subscription.schedule` (tolgee/billing#285) |

### Licensing errors identified by response code instead of message

`catchingOutOfCredits` and `releaseSubscription` matched `HttpClientErrorException.message` — Spring's own rendering of the status and the body. A substring match on the code can hit anywhere in that string, including inside `params`, and the format is Spring's to change. Sibling handlers in the same class already read the parsed body; these now do too.

`parseTolgeeErrorBody` uses a plain Jackson 3 mapper, which ignores unknown properties by default; `ParseTolgeeErrorBodyTest` pins that, so a stricter mapper cannot quietly reintroduce the silent-stop this branch removes (one unrecognised field in the licensing server's error body would otherwise fail the parse and drop every code check).

### A guard that had already gone dead

`SetKeysNamespaceChunkProcessor.catchingKeyAlreadyInNamespace` decided "this key name is taken" from an index name inside a driver message. changeSet `1758202102054-2` dropped `key_project_id_name_namespace_id_idx` and `key_project_id_name_idx` (`schema.xml:4761,4764`) and created `key_project_branch_name_no_ns` / `key_project_branch_name_ns` instead (`schema.xml:5115,5120`). On a migrated database the guard **could not match**: the batch job had been rethrowing the raw constraint violation rather than reporting `KEY_EXISTS_IN_NAMESPACE`.

It now reads the constraint name from the driver's own `ErrorResponse` field — not `ConstraintViolationException.constraintName`, which Hibernate fills in by searching the server message for the English `violates unique constraint "` and so leaves empty under a non-English `lc_messages`. `KeyUniquenessConstraintNameTest` provokes a real duplicate-key violation and asserts the reported index is one the guard recognises, so a further rename fails a test rather than silently disabling the guard.

### A test that never ran

`CreditLimitTest.correctly propagates credit spending limit exceeded` carried only `@ProjectJWTAuthTestMethod`, which — unlike its sibling `@ProjectApiKeyAuthTestMethod` — is not meta-annotated with `@Test`. JUnit never discovered it, so the `CREDIT_SPENDING_LIMIT_EXCEEDED` guard had no coverage while its twin did. `@Test` added; the test passes.

Meta-annotating `@ProjectJWTAuthTestMethod` itself was tried and reverted: `V2ExportAllFormatsTest.it exports to all formats` carries both `@ParameterizedTest` and that annotation, and JUnit cannot resolve a method that is both. A sweep of all six test source roots found this was the only one of 886 usages missing `@Test`, so the annotation is left as it is.

## Guards deliberately left matching on messages

| Site | Why |
|---|---|
| `handleOtherExceptions` `"Broken pipe"` | The text comes from the OS via the JDK and Tomcat, so there is no local producer to drive and no typed signal on the `IOException`. |
| `SentryBeforeSendCallback` | Decides only whether an event reaches Sentry; no caller-observable control flow. |
| `getGroupOrNull` `"No group with name"` | `Pattern.namedGroups()` (JDK 20+) would work, but adopting it changes the helper's signature and all seven call sites across two regexes in the placeholder-parsing path. Its drift is loud rather than silent — a reword rethrows instead of returning null, reddening the po/PHP/C/i18next format tests. |
| `ApplicationStopper` `"SQLGrammarException"` | Left exactly as it is on `main`. Scanning the cause chain would loosen the `stopRightAfterStart` exit-0 build gate in one direction only, with no test to pin it; that belongs in its own change, not in this audit. |
| billing `"already attached"` | The field alternative is destructive — see tolgee/billing#285 — and the test that could detect a reword belongs in the credential-gated real-Stripe suite. |
| `EmailNotVerifiedView.tsx` | Not message matching. `RateLimitResponseBody.message` is a `Message` enum with `@JsonValue`; the field name is just misleading. |

## What still answers 500

Probed against this branch on the endpoint from the bug report:

| `?sort=` | answer |
|---|---|
| `nope` | `400 unknown_sort_property` |
| `name.nope` | `400 unknown_sort_property` |
| `name,nonsensedirection` | `400 unknown_sort_property` |
| `translations` | **500** `UnsupportedOperationException` |
| `keyMeta.tags` | **500** `UnsupportedOperationException` |

Sorting by a **to-many association** raises `UnsupportedOperationException` before any path resolution, so it never reaches this handler. It is deliberately not folded in: that exception carries nothing identifying it as sort-related, so mapping it to 400 would answer client errors for genuine server bugs.

Endpoints using **Spring Data derived queries** raise `PropertyReferenceException`, which no handler covers, so those still answer 500 too.

Both gaps want the same fix, out of scope here: validate the sort parameter against the entity metamodel before the query is built, instead of classifying failures afterwards. Two consequences worth recording, since they are the argument for scheduling it:

- `?sort=translations` is a **client-triggerable Sentry event source**. It reaches `handleOtherExceptions`, which captures unconditionally, so any authenticated caller can emit unlimited events from a query parameter. Pre-validation closes the wrong status code and this together.
- The new `400` answers `unknown_sort_property`, and `webapp/src/translationTools/useErrorTranslation.ts` has no case for it (its `default` returns the raw code). The webapp does not send bad sort values, so this surfaces only if the webapp itself has a sort bug or a user drives the API from the UI. Adding the case needs the `t()` key created on Tolgee project 1 first, so it is called out here rather than half-done.

## Tests

Deliberately kept to the guards this branch changes — tests that characterised untouched production code were removed rather than carried.

- `KeyControllerTest` — the two disjoint Hibernate hierarchies the fix recognises: an unknown attribute (`?sort=keyName`) and dereferencing a basic property (`?sort=name.nope`).
- `ExceptionHandlersTest` — the rewritten handler: a direct `PathElementException`, one buried below an unrelated cause, an unrelated `InvalidDataAccessApiUsageException` that must be rethrown, and an unresolvable path on a request with no `sort` parameter, which must also be rethrown.
- `KeyUniquenessConstraintNameTest` — a real duplicate-key violation, inside and outside a namespace.
- `PostgresAutoStartConfigurationTest` — SQLState `57P03` directly and wrapped, and other SQLStates that must not match.
- `FrontendUrlProviderTest`, `AzureFileStorageFactoryTest`, `ParseTolgeeErrorBodyTest`, `EeLicenseControllerTest` — the remaining converted guards.

## History

After review the branch was rebuilt as nine self-contained commits, one per converted guard, each compiling and passing ktlint on its own; read it commit by commit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid sort properties now return a clear 400 error instead of causing an unexpected server failure.
  - License release succeeds when the remote service no longer recognizes the license key, while other failures remain visible.
  - Improved handling of Azure storage connection-string errors.
  - PostgreSQL startup detection is more reliable across wrapped database errors.
  - Duplicate key processing now consistently identifies supported uniqueness conflicts.


backend/app/src/main/kotlin/io/tolgee/ExceptionHandlers.kt
```diff
@@ -222,29 +223,25 @@ class ExceptionHandlers : Logging {
...
   @ExceptionHandler(InvalidDataAccessApiUsageException::class)
-  fun handleFileSizeLimitExceeded(ex: InvalidDataAccessApiUsageException): ResponseEntity<ErrorResponseBody> {
-    Sentry.captureException(ex)
-    val contains = ex.message?.contains("could not resolve property", true) ?: false
-    if (contains) {
+  fun handleInvalidDataAccessApiUsage(
+    ex: InvalidDataAccessApiUsageException,
+    request: HttpServletRequest,
+  ): ResponseEntity<ErrorResponseBody> {
+    val sortedByClient = request.getParameter("sort") != null
```
dkrizan, line 236: based on claude code-review:

> ... on a sorted request a server bug now gets the client's 400 and loses Sentry (the old handler called Sentry.captureException(ex) unconditionally, first statement).

> Scenario: a Specification or @Query references an attribute renamed by a refactor; the user opens the keys list, which sends ?sort=name; the endpoint answers 400 unknown_sort_property blaming a perfectly valid sort, and nothing reaches Sentry — the exact silent failure this branch exists to remove, just relocated. Cheapest fix: capture to Sentry (or logger.warn) before returning 400 when the unresolvable path isn't the one the client asked for, or compare it against the sort parameter value.

seems to me it's valid to handle, even though the chance this scenario happens might be low

Anty0: I'll have a look. Worth noting: I had issue with this PR before - it kinda blew up. AI kept finding more and more edge cases where we could return wrong code or something, until the patchset had over 5k lines added and all of that just to handle few more edge cases like this, so I threw it away and make this PR. This PR is kinda inverse - smallest patch that fixes the problem. But I'll have a closer look at what you sent - might be worth addressing.

Anty0: Had a look, and yeah, this is exactly one of the things I threw away. I don't want Sentry spammed every time someone sends a bad sort param, and there's no sane way to get out of Hibernate what actually failed. Telling "our query is broken" apart from "user sent a wrong sort" basically means string matching again plus reflection magic, with special handling for native queries.

That's pretty much what the first version of this PR was. It validated sort params against the JPA metamodel before running the query, parsed `@Query` strings, had its own annotation for native queries... just that part of the PR ended up at ~2.6k lines. Happy to push it if you want to see how bad it got :D

So I went with ignoring it and relying on tests. If one of our queries is wrong, it fails every time that code runs, so tests should catch it. Dynamic queries can technically break only for some filter combinations, but that's a whole other can of worms.

So I think it's either:
- report it to Sentry and live with the spam from bad sort params
- go back to the big contraption
- ignore it like now
- ignore it, but at least log it as a warning

What do you think?

dkrizan: ok, ignore :)

dkrizan (dismissed):

dkrizan (dismissed):

backend/data/src/test/kotlin/io/tolgee/unit/component/fileStorage/AzureFileStorageFactoryTest.kt
```diff
@@ -1,47 +1,62 @@
...
+  }
+
+  private fun create(connectionString: String?) {
+    val config =
+      AzureContentStorageConfigDto().apply {
+        this.connectionString = connectionString
+        this.containerName = "container"
       }
-    exception.cause.assert.isNotNull
+    AzureFileStorageFactory().create(config)
```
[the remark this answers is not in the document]
Anty0: This is actually fine. TestData are only required for data that has to be stored in database. If the test doesn't store data, just uses it to call functions, it doesn't need to be in TestData. Usually the difference is integration vs unit tests - integration tests test whole thing and need the data in db, while unit tests just pass the data as params and doesn't touch the db.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/EeLicenseControllerTest.kt
```diff
@@ -200,6 +205,46 @@ class EeLicenseControllerTest : AuthorizedControllerTest() {
     }
   }
 
+  @Test
+  fun `releases a license key the remote no longer knows`() {
+    prepareSubscription()
```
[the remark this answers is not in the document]
Anty0: Preexisting pattern in this test, but otherwise valid. Not worth fixing right now on already reviewed PR.

dkrizan (approved):

PR #3831, opened 2026-07-30
feat: word-based pricing

dkrizan:
Platform side of word-based pricing. Pairs with the corresponding server-side billing changes on the same branch name — both land together.

## What's here

**Word metric and limits**
- `HOSTED_WORDS` metric type and a hosted-words count in `OrganizationStatsService`
- `WordsLimitChecker` + a words usage limit, and a per-translation word-delta helper so edits are checked against the delta rather than a recount
- Self-hosted translation edits are blocked over the words limit, unless auto-upgrade is on — then going over is allowed and reported

**Self-hosted subscription model**
- `words` on the self-hosted usage limits and plan models, carried onto the instance subscription
- Self-hosted word usage is reported to the billing server
- `autoUpgradeEnabled` (from the licence), `boostEndsAt`, and the scheduled downgrade are exposed on the subscription model
- `includedWords` must be positive for `HOSTED_WORDS` self-hosted plans

**Admin plan forms**
- Cloud plan validation reworked for atomic tiers, plus hosted-words and multi-tier self-hosted validation
- `cloud-plan` renamed to `cloud-plan-tier` in the billing frontend to match the new model

**Billing UI**
- Billing plans sub-route, with the organization billing menu items merged into one. The `/subscriptions` and `/invoices` routes still resolve; only the menu entries are consolidated
- Word allowances rendered like the existing string/seat allowances (`IncludedWords`), and word usage shown in the usage detail
- Subscription dates extracted into a shared `ActivePlanDates` so the cloud and self-hosted active cards stay identical
- The plan-limit dialog handles an exceeded word limit and offers enabling auto-upgrade
- `useMoneyFormatter` no longer throws when a caller asks for fewer digits than its default minimum

**Plan card redesign**

The plan cards were reworked to match the pricing design, which is where most of `PlanStyles.tsx` comes from. The layout values it introduces have no theme token behind them, so listing them here rather than in code comments:

- **Header band** — plan name and price move into a coloured band (`PlanHeader`). `min-height: 140px` gives a contact-us plan, which has one line where the others have price + period, the same band height so the row's header reads as continuous. `padding-top: 44px` reserves room for `PlanHeaderBadge`, which is out of flow.
- **Joined row** — `PlanRow` is one surface with pseudo-element dividers; individual cards drop their own chrome via `.inRow`. `PLAN_ROW_MAX_COLUMNS = 4` is the single source for both the `minmax()` floor and the `--plan-columns` fallback; `350px` per column caps a sparse tab so two self-hosted plans do not stretch to twice the width the other tabs use.
- **Break-out** — the active/popular card is pulled `-16px 0` out of the row. `PlanHeader` absorbs those 16px (`min-height: 155px`, `padding-top: 39px`, less the 1px border only a standout card draws) and `PlanContent` absorbs them at the bottom (`padding-bottom: 47px`), so the band's bottom edge, the plan name, the price and the bottom-aligned CTA stay level with the plain cards.
- **Standout chrome** — the pink card leans on its border plus a `primary._states.hover` wash layered over the card's own surface (kept opaque, so the row's border does not show through the 16px it breaks out past the edge) rather than a coloured glow: `elevation.pricingActive` is teal and reads as a halo of the wrong hue around a pink card.
- **Badge pill** — `top: 12px`, `padding: 3px 12px`, `border-radius: 999px`, on the theme's `caption` size and 500 weight. `elevation.*` carries colour only, so the `0 0 20px 0` shadow geometry has no token; it lives in one local `pricingShadow()` helper.

**API models and messages**
- New message codes: `CANNOT_DOWNGRADE_WORD_TIER`, plan tiers disagreeing on billing period, USD price outside a hosted-words plan, organization currency already fixed by Stripe
- `Currency` enum, per-currency MT credit price on the plan prices model, word usage on the current-usage model
- Batch job types for the billing auto-upgrade notice and the pre-renewal check

**Fixes picked up along the way**
- Word-count reporting is skipped on organization deletion when billing is enabled
- Settings menu prefixes match at path-segment boundaries only (needed because `/billing` must not match `/billing-test-clock-helper`)
- Vitest picks up the billing frontend tests when the billing checkout is present, and says so when it does — the billing frontend has no test tooling of its own, so those tests otherwise run nowhere (the `eslint` and `tsc` scripts in the same package already resolve that sibling the same way)
- The key-count reporting listener now shares the word listener's soft-deletion predicate, so creating an entity is no longer read as a deletion
- The auto-upgrade notice email no longer injects an English noun (`per {periodNoun}`) into a translated sentence; the period is an ICU `select` inside the string
- `auto_upgrade_cannot_be_disabled_over_limit`, `cannot_downgrade_word_tier` and `organization_currency_already_set` reach the user as sentences rather than raw codes
- The plan-limit dialog offers auto-upgrade based on the error code that was raised, not on a usage ratio — the rejected write is never persisted, so the ratio only crossed 1 in the rare case where the previous accepted write landed exactly on the ceiling
- `BILLING_AUTO_UPGRADE_RENEWAL` keeps its 3 x 30s retry window; a renewal that exhausts it is handled on the billing side

**Cost of counting words**

`countAllWordsOnInstance()` is a full-instance aggregation, unlike the index-friendly `select count(k)` behind the key metric, so both self-hosted paths gate it:

- **Reporting** — a change only raises `UsageToReport.wordsDirty`; `UsageReportingService.refreshWordsToReport` is the single place that pays for the count, and it skips instances whose licence does not meter words (reporting `null` rather than a fabricated zero). Whether a licence meters words is read from the plan's `metricType`, now persisted on `EeSubscription`, not inferred from the allowance: the licence server sends `Limit(-1, -1)` both for a keys-and-seats plan and for a word plan whose allowance is unlimited, and `Limit(-2, -2)` for a negotiated one, so the numbers cannot tell them apart. Enforcement stays on the numbers (`limit >= 0`) — a negotiated allowance has nothing to block on, and comparing writes against -2 would refuse all of them. The cost is that unlimited and negotiated word instances now pay the 15-minute aggregation they previously skipped; that is the only way the figure reaches the licence server at all. No aggregation ever runs in the request that wrote the translation, and at most one runs per reporting window. The flag is read and cleared by one statement (`takeWordsDirty`) in its own short transaction, so the answer comes from the row rather than from the cached DTO — an eviction that lands before the raising writer commits would otherwise let a concurrent read repopulate the cache with a stale `false` that nothing evicts again, freezing the reported figure indefinitely. The row is not locked across the aggregation or across the licence-server call that follows, and the count itself runs in its own transaction so a failed aggregation cannot take the keys and seats figures down with it. The word listener is ordered last on the activity event so its lock never encloses the key listener's synchronous licence-server POST, which would otherwise stall every translation write on the instance for the duration of that call.

One residual window is worth naming: the flag is cleared before the count it authorises runs, so a process death in that window (up to one aggregation, ~1.7 s at 1M keys) loses the recount until the next content change or licence-limit change. Failures are handled — the catch is on `Throwable`, restores the flag and reports the last stored figure — but a kill is not. Enforcement is unaffected, since it counts live. The fix is to replace the boolean with a change sequence the reporter compares after counting; it needs one more thing solved first, because storing `words_counted_at` only when the sequence is unchanged means a busy instance never closes its recount window and aggregates on every 60-second tick. A failure is logged, restores the dirty flag and reports the last stored figure rather than propagating — otherwise an aggregation that keeps failing on a large instance would halt reporting for two metrics that have nothing to do with words.

Words also carry their own minimum interval (`tolgee.ee.reportWordsMinIntervalInMs`, 15 minutes) rather than the keys-and-seats cadence: the figure drives monthly billing and a renewal-time decision, and nothing reads it sub-minute, so a full-instance aggregation every 60 seconds buys nothing. The clock (`words_counted_at`) is stamped when the count is taken rather than when a report is sent, so a word-count-preserving edit — a typo fix, a retranslation of the same length — does not leave the window open and re-pay for the aggregation every tick. A figure already counted but not yet accepted by the licence server is retried on every tick without recounting.

The visible consequence is that enforcement is live while reporting is throttled: a write refused with `plan_word_limit_exceeded` is checked against an exact count taken in that transaction, but the self-hosted limit dialog reads the cloud's last accepted report and can sit up to a recount window behind. The dialog says so rather than showing a figure that contradicts the refusal.

Raising the flag is an unconditional row update, which is deliberate: the row lock is what makes the reporter's read-and-clear wait for a writer that has raised the flag and not yet committed. The price, worth naming, is that the lock is taken before the activity revision is written and held to commit, so concurrent translation commits on a word-metered instance serialise behind the largest in-flight write's activity storage. The two alternatives were rejected: `ActivityHolder.afterActivityFlushed` is a single-slot callback already used by `BatchJobActivityFinalizer`, and a conditional update plus a periodic safety recount trades bounded contention for an unbounded window of a stale billing figure.

Paths that bypass the activity interceptor publish `OnProjectContentReplaced` instead — admin project import clears content with bulk JPQL and writes entities with activity logging disabled, so the cloud would otherwise keep billing the pre-import figure. Both the word and the key reporting listeners handle it; keys had the same staleness and are the metric self-hosted licences bill on today.

The same asymmetry would have blocked the import outright: the wipe emits no entity events while the re-insert emits one per row, so the enforcement listener would add a project's whole word count to a baseline that still included it and refuse a restore of content the instance already held. `ContentReplacementScope` (transaction-scoped, so it cannot leak between concurrent requests) marks that transaction and **both** limit listeners stand down for it. The marked block encloses the importer's explicit `flush()`, because a dirty update's entity event is emitted at flush rather than at the mutating call, and a flag reset in a `finally` can never cover the commit-time flush — the key listener had the identical shape and refused the same import, so exempting only the word one would have left the import still blocked. `ProjectExportImportImporterTest` pins this through the real importer; without the guard it fails with `plan_key_limit_exceeded`.

The instance declares which metrics it can report (`reportedMetrics`) on every licence call — activation and refresh alike, not activation only. A subscription can be moved onto a word plan long after its key was set, and the refresh is the only call that can see it happen; an instance too old to understand the metric would otherwise enforce nothing and report nothing, forever, with neither side able to notice. The refresh deliberately still returns a usable subscription whatever the declaration says — refusing there would strip the licence and every EE feature with it. Acting on the declaration, by refusing to move a subscription onto a metric its instance cannot report, belongs on the billing side.

Two costs worth stating rather than discovering: `prepareSetLicenceKey` and `setLicenceKey` both take the full word count before their remote call, so an admin activating a licence on a 1M-key instance waits ~1.7 s twice — the preview needs the figure for the cloud to say whether the plan fits. And an unrecognised `metricType` from a newer licence server falls back to `KEYS_SEATS`, so such a plan's card renders keys-and-seats allowances; enforcement is unaffected, since it reads the limits rather than the metric.

It also carries its own deferral clock (`words_reported_at`): the shared `reported_at` is advanced by every key and seat report, so on an instance with steady key activity it never reopens and word usage would never reach the cloud.
- **Enforcement** — `EeWordCountLimitListener` only runs on an instance with an enforced word limit and auto-upgrade off, and skips the count entirely when the transaction's net word delta is `<= 0`. The count itself stays exact: the check blocks writes, so it is not run against a stale figure. Only translation writes are checked; restoring a key, language or branch and turning branching on raise the count without being blocked, which is deliberate (refusing an undo of a deletion is a worse answer than blocking the next edit) — pinned by `WordCountLimitTest`. The cloud is told, and now actually records it: a periodic report is a measurement of content that already exists, so `/report-usage` stores the figure before it judges it, and it no longer refuses a word count from an instance whose subscription has auto-upgrade in effect — that instance was told to stop enforcing, so growing past its allowance is the designed behaviour and the report is what triggers the upgrade. Previously the licence server refused such a report outright and the platform sent keys, seats and words in one POST, so the refusal left all three unstored, the identical payload was retried every minute, key and seat billing froze alongside words, and the cloud never learned the figure auto-upgrade is driven by — so the instance could not leave the state. The platform now retries without the refused figure so the other metrics still land, and does not re-offer a word count the server already refused until it changes.

Measured, warm cache, `word_count` populated on every row, keys on a project's default branch as in production:

| Instance size | `countAllWordsOnInstance()` | `select count(k)` equivalent |
|---|---|---|
| 200k keys / 400k translations | ~390 ms | — |
| 1M keys / 2M translations | ~1.7 s | ~40 ms |

Five ways to make the query cheaper were tried and measured, and none of them beat it: splitting the branch collapse out at translation level, at key level, swapping `is not distinct from` for a coalesced join key, dropping the `materialized` hint, and raising `work_mem` to 256 MB. The aggregate looks like the cost on a fixture whose keys have no branch, but once every key resolves a branch row — which is production, since every project gets a default branch and every key is assigned one — the join dominates and skipping the grouping saves nothing. **All five were reverted**, so the query carries no branch-collapse fast path: the `max(word_count)` collapse over `(project, name, namespace, language)` is unconditional.

The two branch predicates in `counted_keys` are correctness filters, not an optimisation, and are worth reading as such:

- `k.branch_id is null or b.deleted_at is null` — drop keys whose branch was soft-deleted, while keeping keys that belong to no branch at all.
- `p.use_branching = true or k.branch_id is null or b.is_default = true` — when a project has branching switched off, only its default-branch (and unbranched) keys are billable; its dormant feature-branch content is not.

Neither gates a cheaper path, and neither adds a subquery. The one `exists` in the query is the soft-deleted-language filter, also a correctness filter.

Three known enforcement gaps, all recorded rather than closed. Restoring content is not limit-checked (see above). A transaction that *replaces* content is checked on its intermediate peak, because Hibernate flushes inserts before deletes — so replacing 50 words with 50 words at the ceiling is refused (both pre-existing in the key-count listener). And on a branching project a transaction can skip the check entirely: the running total is a raw sum of per-translation deltas while the billed figure is `MAX(word_count)` per key name across branches, so a decrease on a branch sibling is subtracted from a total it never contributed to and can cancel a real increase made in the same transaction. An instance at its ceiling that empties a branch twin and writes 50 words elsewhere lands 50 over, unchecked. The damage is bounded and self-correcting — the next word-increasing transaction re-reads the true baseline and blocks, and reporting still sends the true figure, so billing stays correct. Pinned by `WordCountLimitTest`.

**Unlimited and negotiated word allowances**

`CloudPlanRequestValidator.validateWordTier` accepts the `-1` (unlimited) and `-2` (negotiable) sentinels, and the plan card renders both, but the admin plan form requires `includedWords > 0` so neither can be created through it. The form is deliberately left as the conservative setting: ascending `includedWords` is assumed as tier order in a dozen places on the billing side — `PlanDowngradeDetector.rank`, three model assemblers, four repository queries, `CloudBillingService`, `SelfHostedEeBillingService` and `AutoUpgradeService` — so a `-1` tier would sort as the *smallest* and silently invert plan-page order, auto-upgrade targets and downgrade verdicts. Settling the disagreement means either special-casing the sentinels in that ranking or tightening the cloud validator to match the form, and that is a billing-side product decision. The instance side does not depend on the answer: metering is keyed off `metricType`, so a plan holding either sentinel is handled correctly however it was created.

**Seats on a word plan**

Only the free word tier caps seats; paid word plans do not meter them, so their cards read "Unlimited seats" rather than quoting a number nothing enforces. The free tier's cap is not yet enforced anywhere — `BillingSubscriptionLimitsProvider.getSeatsLimit` returns `-1` for every `HOSTED_WORDS` plan — which is a billing-side follow-up. Keys-and-seats plans are untouched — they do meter seats. The usage bar follows the server: `getProgressData` shows a metric's bar only when the server reports a ceiling for it (`…Limit !== -1`), which now applies to keys, strings, seats and words alike rather than to seats only. Making the free word tier's seat cap actually enforced is a billing-side change (`BillingSubscriptionLimitsProvider.getSeatsLimit` returns `-1` for any plan whose metric does not use seats) and is not in this PR.

**Auto-upgrade: two flags, two names**

The licence response carries the customer's `autoUpgradeEnabled` toggle and, separately, `limits.autoUpgradeEffective` — whether auto-upgrade will actually cover an overrun, which is narrower (a subscription with no bigger tier to move to has to be blocked by the instance anyway). Only the effective one is persisted and enforced against, and it keeps that name the whole way down: `EeSubscription.autoUpgradeEffective`, `ee_subscription.auto_upgrade_effective`, `EeSubscriptionDto`, and `UsageLimits.Limit.autoUpgradeEffective` — on the words limit rather than at the top of `UsageLimits`, since it governs that one metric.

**Forward compatibility of the licence response**

A self-hosted instance can be older than the licensing server it talks to. `HttpClient`'s response mapper ignores unknown properties and falls back for unknown enum values — `MetricType` to `KEYS_SEATS` (the metric an older instance can actually enforce), `SubscriptionStatus` to `UNKNOWN`, `BillingPeriod` to `MONTHLY`, and nullable enums such as `Currency` to `null`. `LicenceResponseMapperTest` pins each case; without it a single unknown value fails the whole licence check rather than the field it appeared in.

## Open decisions

Two things are deliberately left for the reviewer to settle rather than decided here.

**1. Should word-limit enforcement stay exact?** Enforcement runs `countAllWordsOnInstance()` inside the blocking transaction, so the figure it refuses on is never stale — at the measured cost above (~390 ms at 400k translations, ~1.6–1.7 s at 2M, against ~40 ms for the key metric). Reporting has been moved entirely off the request path and is unaffected; this is only about the write that trips the limit. The alternatives are to approximate (cache the baseline for a short window and accept letting some writes past the ceiling) or to enforce only above a threshold. Both trade exactness for latency on the largest instances, which is a product call.

**2. Is the unconditional branch collapse the right shape?** See the paragraph above: five cheaper variants were measured and none won on production-shaped data, so the collapse is paid by every count. If the benchmark's assumption is wrong for real Tolgee data — specifically that every key resolves a branch row — the conclusion changes and the split-out variants deserve another look.

## Known and not addressed

Surfaced by review and deliberately left, rather than fixed late:

- `getProgressData` treats `wordsLimit` as a hard ceiling, so the critical-usage warning pins on for auto-upgrade word customers, who are the ones *meant* to exceed their tier. Needs `autoUpgradeEffective` on the cloud usage model to fix properly.
- `setLicenceKey` stores an exact fresh word count but leaves `wordsDirty` raised, so a freshly activated instance pays one redundant full-instance aggregation 15 minutes later.
- `PlanLimitPopoverCloud` renders the generic dialog first and pops in the auto-upgrade paragraph and button once the subscription query resolves.
- The `plan_word_limit_exceeded` string still reads "Upgrade your plan or enable auto-upgrade", which names an action a customer who already has auto-upgrade on cannot take. The dialog now says the right thing in that case; the shared string is live in 17 locales, so rewording it is a Tolgee-side re-translation rather than a code change.

## Testing

Backend unit tests added for the new behaviour (`WordsLimitCheckerTest`, `SelfHostedLimitsProviderTest`, `SubscriptionFromModelAssignerTest`, `WordUsageIncreaseAmountTest`, `SelfHostedUsageLimitsModelDeserializationTest`), plus `WordCountLimitTest`, `KeyCountLimitTest`, `WordUsageReportingTest`, `OrganizationStatsWordCountTest` and a `ProjectExportImportImporterTest` case covering an import that momentarily appears to exceed both limits, on the integration side, and `getProgressData` / `GlobalValidationSchema` word-tier cases in the webapp. End-to-end coverage for the feature runs against this branch from the billing test suite.

## Summary by CodeRabbit

* **New Features**
  * Added hosted-word billing plans with allowances, pricing, usage tracking, and progress indicators.
  * Added self-hosted word limits, enforcement, and usage reporting.
  * Added automatic upgrade options and renewal notices for word-based plans.
  * Added subscription dates, currencies, invoicing details, and onboarding boosts.
  * Consolidated organization billing navigation under a single Billing section.

* **Bug Fixes**
  * Improved handling of newer billing responses and unknown values.
  * Improved plan-limit and word-limit error messages and dialogs.
  * Fixed usage calculations across translations, branches, and content replacement.
  * Improved currency-aware price formatting and billing plan validation.


ee/backend/app/src/main/kotlin/io/tolgee/ee/service/eeSubscription/EeSubscriptionErrorCatchingService.kt
```diff
@@ -27,8 +27,10 @@ class EeSubscriptionErrorCatchingService(
       when (body.code) {
         Message.SEATS_SPENDING_LIMIT_EXCEEDED.code,
         Message.KEYS_SPENDING_LIMIT_EXCEEDED.code,
+        Message.WORDS_SPENDING_LIMIT_EXCEEDED.code,
         Message.PLAN_KEY_LIMIT_EXCEEDED.code,
         Message.PLAN_SEAT_LIMIT_EXCEEDED.code,
+        Message.PLAN_WORD_LIMIT_EXCEEDED.code,
         ->
           throw BadRequestException(body.code, body.params)
```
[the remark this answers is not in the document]
dkrizan: Not applied. `Message.code` is `name.lowercase(Locale.ROOT)` (`Message.kt:383-384`), and the licence server builds its error bodies from that same `Message` enum via `ErrorResponseBody`, so the uppercase input this guards against cannot be produced. `catchingLicenseUsedByAnotherInstance` three lines below compares the same way and has shipped unchanged; normalising only here would make the two inconsistent without covering a real case.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/selfHostedLimitsAndReporting/KeyCountLimitTest.kt
```diff
@@ -155,6 +159,23 @@ class KeyCountLimitTest : AbstractSpringTest() {
...
+    }
+    val testData = saveTestData(2)
+
+    // The wipe is bulk JPQL and produces no negative delta, so without the marker the re-insert
+    // would be added to a baseline that still counts the content it replaces.
+    executeInNewTransaction {
+      contentReplacementScope.replacingContent {
+        createKey(testData)
+      }
+    }
```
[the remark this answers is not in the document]
dkrizan: Not applied. `AbstractSpringTest` extends `AbstractTransactionalTest`, which registers `CleanDbTestListener` in its `@TestExecutionListeners` — the database is already wiped between tests, so a persisted root cannot leak into a later one. That is also why these tests can assert absolute instance-wide figures such as `countAllWordsOnInstance() == 100` at all; without it they would fail on the first leftover fixture.

The existing `@AfterEach` in `WordCountLimitTest` and `WordLimitsAssignedTest` deletes the `ee_subscription` row, which is a different concern: it is a singleton the licence lookup reads, not a TestData root.

webapp/src/ee/billing/component/Plan/IncludedUsage.tsx
```diff
@@ -1,27 +1,53 @@
...
+import { T } from '@tolgee/react';
+import { SecondaryChip } from 'tg.component/common/chips/SecondaryChip';
 import { PlanType } from './types';
 import {
   IncludedCredits,
   IncludedKeys,
   IncludedSeats,
   IncludedStrings,
+  IncludedWords,
 } from '../IncludedItem';
```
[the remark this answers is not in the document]
dkrizan: Not applied — the alias would break the Vite build.

`tg.ee.module/billing/*` is mapped to two roots in `webapp/tsconfig.extend.json:16`:

```
"tg.ee.module/billing/*": ["../../../billing/frontend/src/*", "ee/billing/*"]
```

`tsc` walks that fallback list, Vite resolves the first entry only. We hit exactly this earlier on this branch: an aliased import of a module that lives solely under `webapp/src/ee/billing` type-checked clean and failed at runtime.

The convention in this subtree reflects that — `webapp/src/ee/billing` currently has 43 relative sibling imports against 2 aliased ones. `./types` and `../IncludedItem` follow the local majority.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/BranchedWordCountLimitTestData.kt
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * The same key on two branches of a branching project, so the instance word count is the larger of
+ * the two translations rather than their sum.
+ */
+class BranchedWordCountLimitTestData(
+  defaultBranchWordCount: Int,
+  featureBranchWordCount: Int,
+) {
+  val root = TestDataBuilder()
```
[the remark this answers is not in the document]
dkrizan: Applied in 573db9c8. Now extends `BaseTestData("branched-word-count-limit-user", "Branched Word Count Limit Project")`; only the branch-specific setup stays in `projectBuilder`. 86 lines down to 40, and `root`/`project` now come from the base class so no test changed.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/projectExportImport/ProjectExportImportImporterTest.kt
```diff
@@ -169,6 +179,37 @@ class ProjectExportImportImporterTest : AbstractSpringTest() {
...
+        enabledFeatures = Feature.entries.toTypedArray()
+        lastValidCheck = Date()
+        includedKeys = 1
+        keysLimit = 1
+        includedSeats = -1
+        seatsLimit = -1
+        includedWords = 1
+        wordsLimit = 1
+      },
+    )
```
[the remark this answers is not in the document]
dkrizan: Applied in 573db9c8 — the fixture now sets `metricType = MetricType.HOSTED_WORDS`.

Confirmed this does not cost the key half: `EeKeyCountLimitListener` never reads `metricType`, and `SelfHostedLimitsProvider` passes `keys` straight through from the subscription regardless of it, so both listeners are live.

Verified the word path is genuinely exercised now rather than assuming it: with the listener's `contentReplacementScope.isReplacingContent` guard temporarily removed, the test fails. Before this change it passed either way, which is what made the word half inert.

Left out the extra assertion that the source word count exceeds the limit — `includedWords = 1` against a fixture importing multi-word translations, and the guard-removal check above already pins the premise.

webapp/src/ee/billing/component/Plan/IncludedUsage.tsx
```diff
@@ -1,27 +1,53 @@
...
   metricType,
+  free,
   highlightColor,
   sx,
   className,
+  onboardingBoostMonths,
+  onboardingBoostCredits,
 }: Props) => {
+  const hasBoost = Boolean(onboardingBoostMonths && onboardingBoostCredits);
+  const wordPlanSeats = free ? includedUsage?.seats ?? -1 : -1;
```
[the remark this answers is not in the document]
dkrizan: Applied in 573db9c8. `TrialChipTooltip` now passes `free={plan.free}`, and the prop is required so a future caller cannot omit it.

Worth noting it was not reachable: the tooltip only renders when `activeCloudSubscription.status == 'TRIALING' && trialEnd` (`useTrialInfo.ts:27-28`), and a free plan is not trialed — so `wordPlanSeats` happened to land on the right answer through `undefined` being falsy. Applied anyway, since relying on that coincidence is the actual defect.

webapp/src/globalContext/useOrganizationUsageService.ts
```diff
@@ -58,7 +59,8 @@ export const useOrganizationUsageService = ({
         : val
     );
 
-  const incrementPlanLimitErrors = () => {
+  const incrementPlanLimitErrors = (code?: string) => {
+    setPlanLimitErrorCode(code);
     setPlanLimitErrors((v) => v + 1);
```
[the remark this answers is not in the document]
dkrizan: Applied in 573db9c8 — `increaseCreditPlanLimitErrors` now clears the code.

Also not reachable today, for a reason worth recording: `planLimitErrorCode` is only ever set by `incrementPlanLimitErrors`, which raises the counter in the same call, so a set code implies `planLimitErrors >= 1`. This path only increments when the counter is `0`, and the counter is never reset — so a stale word code and a credit-triggered increment cannot coexist, and `GlobalLimitPopover` reopens on counter changes only.

Applied because it is one line and the invariant is implicit: adding any reset of `planLimitErrors` would make it live immediately.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountLimitListener.kt
```diff
@@ -0,0 +1,76 @@
...
+    if (limits.words.autoUpgradeEffective == true) {
+      return
+    }
+
+    addedWords += event.getWordUsageIncreaseAmount()
+    if (addedWords <= 0) {
+      return
+    }
+
+    WordsLimitChecker(limits = limits).check(instanceWordCountBeforeTransaction + addedWords)
```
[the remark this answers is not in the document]
dkrizan: Confirmed and accepted as a real gap — moved to its own PR rather than fixed here.

The behaviour is already documented in the listener and pinned by two tests (`a mixed transaction on a branching project can take the instance over the ceiling unchecked`, `over-counts a branching project by this transaction's own writes`), so it is a known bound rather than an oversight.

It is not a review-sized change. The suggested final branch-aware validation is the exact-enforcement design, and the cost is already measured: `countAllWordsOnInstance()` is ~1.6 s at 2M translations, ~0.4 s at 400k. Today that aggregation already runs once per word-metered writing transaction via `instanceWordCountBeforeTransaction`, so recounting after the flush instead of before is roughly CPU-neutral — but it has to read the transaction's own uncommitted state, which moves a ~1.6 s aggregation inside the writing transaction and holds its locks for the duration. That is not hypothetical here: an earlier `REQUIRES_NEW` recount in this file held a row lock across the aggregation and produced two self-deadlocks.

So it gets its own branch (`dkrizan/word-limit-exact-enforcement`), with the two tests above rewritten to assert enforcement, concurrency coverage for the lock-duration change, and a before/after benchmark. Leaving this thread open until that lands.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/eeSubscription/usageReporting/UsageToReportService.kt
```diff
@@ -128,18 +135,97 @@ class UsageToReportService(
...
+   *
+   * The eviction condition keeps an idle word-metered instance from evicting the cache on every
+   * tick: the recount window stays open until something is actually counted, so this otherwise runs
+   * once a minute forever and clears nothing. `#result` requires beforeInvocation to stay false.
+   *
+   * @return whether the flag was raised, and therefore whether a recount is owed
+   */
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  @CacheEvict(Caches.EE_LAST_REPORTED_USAGE, key = "1", condition = "#result")
+  fun takeWordsDirty(): Boolean =
```
Anty0, line 204: Four paragraphs on `takeWordsDirty` covering the eviction race, the transaction boundary, a caller restriction and the idle-tick condition. The genuinely non-obvious bit is the Spring cache rule (`#result` needs `beforeInvocation` to stay false) — that one earns its place.

"Only the reporting path may call it" is the part that worries me: a self-deadlock warning that only lives in a KDoc is exactly the one someone burns themselves on, because prose doesn't stop them. Can we make that visibility, or an assert? The remaining two paragraphs describe shapes that are visible in the annotations right below.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountReportingListener.kt
```diff
@@ -0,0 +1,118 @@
...
+   * the same event and makes a blocking licence-server POST whenever its one-minute window is open,
+   * while [onWordCountChanged] takes the exclusive lock on the single usage_to_report row and holds
+   * it to commit. Running last keeps that lock from enclosing the network call, which would stall
+   * every other translation-writing transaction on the instance for its duration. Atomicity is
+   * unchanged: the flag is still raised inside the writer's transaction, before commit.
+   */
+  @Order(Ordered.LOWEST_PRECEDENCE)
+  @EventListener
+  fun onActivity(event: OnProjectActivityEvent) {
+    if (billingConfProvider().enabled) {
```
Anty0, line 43: `if (billingConfProvider().enabled) { return }` is repeated verbatim in all four listeners, and all four funnel into `onWordCountChanged()`, which already has its own `metersWords` guard. Moving it to the top of `onWordCountChanged()` next to the guard it belongs beside would help.

Only difference is that `changesWordCount()` (an in-memory map inspection) would then be evaluated on cloud instances too, which I'd call negligible.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountReportingListener.kt
```diff
@@ -0,0 +1,118 @@
...
+  fun onWordCountChanged() {
+    if (!selfHostedLimitsProvider.getLimits().metersWords) {
+      return
+    }
+    logger.debug("Word count change detected. Marking it for the next report.")
+    usageReportingService.markWordUsageChanged()
+  }
+
+  private fun OnProjectActivityEvent.changesWordCount(): Boolean {
+    if (modifiedEntities.keys.any { it == Translation::class }) {
```
Anty0, line 98: `modifiedEntities.keys.any { it == Translation::class }` is a linear scan spelling `containsKey` — `modifiedEntities.containsKey(Translation::class)`.

webapp/src/ee/billing/getSelfHostedProgressData.ts
```diff
@@ -8,22 +8,33 @@ export const getSelfHostedProgressData = ({
...
+        usage.words.current,
+        isLimitEnforced(usage.words.limit)
+      )
+    : undefined;
+
   return {
     keysProgress,
     seatsProgress,
     creditProgress: creditsProgress,
+    wordsProgress,
```
Anty0, line 38: Local is `creditsProgress`, returned as `creditProgress: creditsProgress`. The rename only exists to match `getProgressData`'s spelling, and this aliasing line is the only place the two meet — naming the local `creditProgress` and returning it shorthand is one less thing to notice.

ee/backend/app/src/main/resources/db/changelog/ee-schema.xml
```diff
@@ -141,4 +141,56 @@
         <addDefaultValue columnDataType="bigint" columnName="seats_limit" defaultValueNumeric="0" tableName="ee_subscription"/>
         <addDefaultValue columnDataType="boolean" columnName="is_pay_as_you_go" defaultValueBoolean="false" tableName="ee_subscription"/>
     </changeSet>
+    <changeSet author="dkrizan" id="1751900000000-1">
```
Anty0: Same as on the billing side: these use `author="dkrizan"` while the existing changesets by the same author are `danielkrizan`. `author` is part of the identity liquibase records, so worth unifying while they're unapplied.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountLimitListener.kt
Anty0: This puts a full-instance aggregate in front of essentially every translation-writing transaction on a word-metered instance: `instanceWordCountBeforeTransaction` runs `countAllWordsOnInstance()`, which materialises a CTE over every non-deleted key, joins every non-empty translation and groups by (project, name, namespace, language) for `max(word_count)`.

I know this mirrors the existing `EeKeyCountLimitListener.initialKeyCount`, so the pattern isn't new — but it's a materially heavier query on a materially bigger table (group-by aggregate over the whole translation join vs a distinct count over `key`), and translation writes are far more frequent than key writes. The PR measures ~1.7s at 1M keys / 2M translations; that's in front of every save.

Worth noting the reporting path already recognised the cost and gated the same call behind `reportWordsMinIntervalInMs`, while enforcement has no throttle at all.

Flagging as a scaling risk rather than a defect — but I'd like an explicit answer rather than carrying it in on the key listener's precedent. Could enforcement use the figure `refreshWordsToReport` already maintains plus this transaction's delta, or the same min-interval with a conservative margin?

webapp/src/ee/billing/limitPopover/useWordsAutoUpgrade.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * The subscription's own plan model is assembled from the active tier alone, so it carries neither
+ * `tiers` nor `currentTierId`. The ladder only comes back from the plans listing.
+ */
+type Offering = {
+  tiers?: { id: number; includedWords: number }[];
+  currentTierId?: number | null;
+};
+
+const isWordPlan = (subscription: Subscription) =>
```
Anty0: `useWordsAutoUpgrade` — the thing the module is named for — is defined last, below the three exported predicates and the private `isWordPlan`. Milder than elsewhere since the predicates are separately tested and legitimately public, but the hook is what a reader opens the file for, and `isWordPlan` sits above both of its callers.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountLimitListener.kt
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  // `addedWords` is a raw sum of per-translation deltas, while the billed figure is MAX(word_count)
```
Anty0: The billed figure is defined in `OrganizationStatsService` as a sum of `MAX(word_count)` per (project, key name, namespace, language) — a deduplicating aggregate. Enforcement here compares `instanceWordCountBeforeTransaction + addedWords`, where `addedWords` is a raw sum of per-translation deltas. A sum of deltas can't track a max-per-group aggregate, so the ceiling can't be exactly right on a branching project.

The PR knows this — the comment says a mixed transaction on a branching project can go over unchecked, and `WordCountLimitTest` has a case named for it. My worry is that this pins the escape as expected behaviour, so it stops being something anyone will trip over and fix.

The accumulator also buys less than it looks like it does: the listener already runs the full `countAllWordsOnInstance()` aggregate once per translation-writing transaction to get the baseline. The delta only exists because that baseline was taken in a separate transaction and can't see the pending writes.

Two shapes worth considering, both cheaper than living with a knowingly-wrong ceiling:
- keep `addedWords` as a cheap trigger, and when `before + addedWords` comes within a margin of the ceiling re-run the exact aggregate and decide on that — one extra aggregate per near-limit transaction, and we're paying for one regardless;
- or run the exact aggregate in the writer's own transaction at pre-commit after a flush — `EntityPreCommitEvent` fires inside that transaction, so the same EntityManager sees the pending writes and the accumulator goes away entirely.

Either way I'd swap the test that pins the escape for one asserting the ceiling holds on a branching project.

backend/api/src/main/kotlin/io/tolgee/hateoas/limits/SelfHostedUsageLimitsModel.kt
```diff
@@ -12,4 +12,9 @@ class SelfHostedUsageLimitsModel(
   val keys: LimitModel,
   val seats: LimitModel,
   val mtCreditsInCents: LimitModel,
+  // A licence server older than this instance omits the field; unlimited is the safe absence
+  // value, since the cloud still enforces at renewal.
+  val words: LimitModel = LimitModel(-1, -1),
+  /** What the instance enforces on: the customer's setting AND eligible AND a higher tier exists. */
+  val autoUpgradeEffective: Boolean? = null,
```
Anty0, line 19: Pure AI (I didn't manage to confirm if this is a real problem, better if you have a look yourself): `autoUpgradeEffective` now travels at two levels of the same payload: top-level on `SelfHostedUsageLimitsModel`, and again on `UsageLimits.Limit`, where the KDoc says it's null for every metric but words. `SelfHostedLimitsProvider` sets it on the words limit only, so three of the four `Limit` instances carry a permanently null field.

Auto-upgrade is a property of the subscription, not of a metric ceiling — pushing it into the generic `Limit` gives every metric a field that's meaningless for it, and makes the same fact travel twice with nothing enforcing that the two agree. `metersWords` was correctly kept at `UsageLimits` level and this is the same kind of fact.

Keeping it on `UsageLimits` next to `metersWords` and having `EeWordCountLimitListener` read `limits.autoUpgradeEffective` would give one field at one level and no per-metric nulls.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/eeSubscription/usageReporting/UsageReportingService.kt
```diff
@@ -89,39 +113,102 @@ class UsageReportingService(
...
+      return null
+    }
+    // A figure already counted but not yet accepted by the licence server is retried every tick;
+    // only the count itself is throttled.
+    val recountAllowedAt = usageToReport.wordsCountedAt.time + eeProperties.reportWordsMinIntervalInMs
+    if (recountAllowedAt > currentDateProvider.date.time) {
+      return usageToReport.wordsToReport
+    }
+    return try {
+      recountWords(usageToReport)
```
Anty0, line 182: First line of `recountWords` is `if (!usageToReportService.takeWordsDirty()) { return usageToReport.wordsToReport }` — so it consumes a dirty flag (a committing mutation, behind a name that promises only a count) and frequently returns the stored figure without counting anything.

`recountWordsIfDirty` would make the call site honest and stop the flag consumption being a surprise.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/selfHostedLimitsAndReporting/WordUsageReportingTest.kt
```diff
@@ -0,0 +1,750 @@
...
+
+      usageToReportService
+        .findOrCreateUsageToReport()
+        .wordsDirty.assert
+        .isTrue()
+    }
+  }
+
+  @Test
+  fun `creating a key does not, even though creation lists deletedAt among its modifications`() {
```
Anty0, line 393: This name only completes against the test above it — on its own, "creating a key does not" doesn't say what's being claimed, and in a CI failure report the name arrives alone. Every other case in this file states its promise in full, so this is the odd one out: `creating a key does not mark the word count dirty, even though creation lists deletedAt among its modifications`.

webapp/src/ee/billing/component/Plan/PlanStyles.tsx
```diff
@@ -9,24 +11,143 @@ export const PlanContainer = styled('div')`
...
+    }
+  }
+`;
+
+export const PLAN_ROW_MAX_COLUMNS = 4;
+
+const PLAN_COLUMN_WIDTH_PX = 350;
+
+const PLAN_CONTENT_PADDING_Y_PX = 32;
+const PLAN_HEADER_MIN_HEIGHT_PX = 140;
```
Anty0: Raw px is this file's existing convention and the commit series does document most of this family of values, so this isn't blocking. What isn't covered anywhere is the 155/39/47 trio and the 44 — the numbers that only exist to compensate for the break-out, and the ones a designer would most want folded into tokens. One line in the standout-variant commit body (or the PR description) naming 140/155/44/39/47 as the header geometry, and noting there's no header-height token, would close it.

Small thing while you're here: `PlanHeaderBadge` uses `border-radius: 999px` while the billing-side pills (`PillToggle.tsx`, `ActiveTypeTag.tsx`) use `9999px` for the same effect.

webapp/src/ee/billing/component/ActiveSubscription/ActivePlanTitle.tsx
```diff
@@ -1,9 +1,9 @@
-import { T, useTranslate } from '@tolgee/react';
-import { Box, Chip, Tooltip, Typography } from '@mui/material';
+import { useTranslate } from '@tolgee/react';
+import { Box, Chip, Tooltip } from '@mui/material';
 
 import { components } from 'tg.service/billingApiSchema.generated';
-import { useDateFormatter } from 'tg.hooks/useLocale';
 import { PlanTitleText } from '../Plan/PlanTitle';
+import { ActivePlanDates } from './ActivePlanDates';
```
Anty0, line 6: `'./ActivePlanDates'` — imports inside `webapp/src` should go through a `tg.*` alias: `'tg.ee.module/billing/component/ActiveSubscription/ActivePlanDates'`.

Anty0 (changes requested):

backend/api/src/main/kotlin/io/tolgee/hateoas/limits/SelfHostedUsageLimitsModel.kt
dkrizan: Moved in db0b6c7c6, one field on UsageLimits. Wire model unchanged.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/eeSubscription/usageReporting/UsageToReportService.kt
dkrizan: Cut to two lines in 399ccc61c. The caller restriction stays prose: `internal` is defeated by the test module, and a transaction check trips tests that call it inside executeInNewTransaction on purpose. Open to a third way.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/eeSubscription/usageReporting/UsageReportingService.kt
dkrizan: Done in 5b19e774a.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountReportingListener.kt
dkrizan: Done in 5b19e774a. Checked first - the four handlers are the only callers.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountReportingListener.kt
dkrizan: Done in 5b19e774a.

webapp/src/ee/billing/getSelfHostedProgressData.ts
dkrizan: Done in 4e6350435.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/selfHostedLimitsAndReporting/WordUsageReportingTest.kt
dkrizan: Done in 977725d1f, shortened to fit 120 columns.

webapp/src/ee/billing/component/ActiveSubscription/ActivePlanTitle.tsx
dkrizan: Done in 4e6350435. The PlanTitle one predates this branch, left it.

ee/backend/app/src/main/kotlin/io/tolgee/ee/service/eeSubscription/usageReporting/UsageToReportService.kt
dkrizan: Made it an error rather than prose (3d288c895): takeWordsDirty sets a 30 s lock_timeout in its own transaction, so a caller that already marked the row dirty fails with a message naming the rule instead of hanging. Pinned by a test.

ee/backend/app/src/main/kotlin/io/tolgee/ee/component/limitsAndReporting/EeWordCountLimitListener.kt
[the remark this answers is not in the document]
dkrizan: Tracked as #3919 rather than a branch: the local branch never got a commit or a push, so there is nothing for this thread to wait on. The issue carries the one-transaction bound, the measured cost, and three options; it needs a benchmark before a fix, so it stays out of this PR.

Anty0 (approved):

PR #3832, opened 2026-07-31
fix: hard-delete project fails on FK constraints for imports, tasks and branch merges

bdshadow:
Follow-up to #3810. After that fixed the `default_namespace` FK, the orphan-organization purge (`DeletedOrganizationProjectPurgeScheduler`) still fails one FK further along, and different orphaned projects hit different constraints. Confirmed in prod (Sentry `TOLGEE-BACKEND-3WB`, and current logs `DELETE FROM language ... violates FK on import_language`).

## Root cause — `hardDeleteProject` leaves child rows referencing the project's languages/keys

I audited every FK into the tables `hardDeleteProject` bulk-deletes (`language`, `key`, `namespace`). Most children are already handled (e.g. `languageService.deleteAllByProject` deletes translations, auto_translation_config, translation_suggestion, language_stats, language_qa_config). Three gaps remained:

1. **Soft-deleted imports** — `getAllByProject` filters `deletedAt is null`, so soft-deleted imports' `import_language` (`existing_language_id → language`) and `import_key` rows survive and block language/key deletion. Now uses `getAllByProjectIncludingDeleted`.
2. **Tasks** — only `task_key` was deleted, never the `task` entity itself (`task.language_id → language`, `task.project_id → project`). Now deletes `task_key`, `task_assignees`, unlinks `notification.linked_task_id`, then the task.
3. **Branch merge changes** — `branch_merge_change.{source,target}_key_id → key` were only cleared via `branchService` *after* key deletion. Now cleared before it.

Native SQL is used for the EE-owned `task`/`branch_merge_change` tables (they exist in the shared schema; matching the existing native deletes in `hardDeleteProject`/`LanguageService`).

## Impact

`hardDeleteProject` is used by the purge scheduler and by the async `OnProjectSoftDeleted` listener, so this also silently broke the normal hard-delete of any project with imports/tasks/branch-merges.

## Tests

`ProjectHardDeletingServiceTest`:
- `deletes project with a soft-deleted import`
- `deletes project with tasks`

Both reproduce the FK violation without the fix (verified) and pass with it; all existing hard-delete tests still pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved permanent project deletion to fully remove projects that contain previously deleted imports.
  - Project deletion now also cleans up associated tasks and branch merge data.
  - Prevented leftover project-related records from remaining after permanent deletion.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/service/project/ProjectHardDeletingServiceTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/service/project/ProjectHardDeletingServiceTest.kt
@@ -14,7 +14,9 @@ import io.tolgee.development.testDataBuilder.data.ContentDeliveryConfigTestData
 import io.tolgee.development.testDataBuilder.data.MtSettingsTestData
 import io.tolgee.development.testDataBuilder.data.ProjectWithQaEntitiesTestData
 import io.tolgee.development.testDataBuilder.data.SuggestionsTestData
+import io.tolgee.development.testDataBuilder.data.TaskTestData
 import io.tolgee.development.testDataBuilder.data.WebhooksTestData
+import io.tolgee.development.testDataBuilder.data.dataImport.ImportTestData
 import io.tolgee.dtos.BigMetaDto
 import io.tolgee.dtos.RelatedKeyDto
 import io.tolgee.fixtures.waitFor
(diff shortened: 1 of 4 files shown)


bdshadow:
Superseded by #3853.

Rebasing this onto current `main` showed two of its three fixes already landed via #3804:

- soft-deleted imports — `ImportService.deleteAllByProject` now uses `findAllByProjectIdIncludingDeleted`
- tasks — `AbstractTaskService.deleteAllByProjectId` deletes notifications, `task_assignees`, `task_key` and the task, in the owning service instead of native SQL in `ProjectHardDeletingService`

`main` also carries tests for both cases.

The remaining piece — clearing `branch_merge_change` before the keys — moved into #3853, along with three further FK gaps the audit turned up (`branch_merge`, the `branch_key_snapshot` family, and the `branch.origin_branch_id` self-reference).

PR #3836, opened 2026-08-01
fix: set correct MIME type on content delivery published files

### LINKED ISSUE #2504: Publishing to Azure storage doesn't set correct MIME type

Anty0:
Fixes #2504.

## Problem

Files published by Content Delivery to S3-compatible and Azure Blob storage were stored without a `Content-Type`, so both services fall back to `application/octet-stream`. CDNs in front of those buckets (Front Door, CloudFront, Cloud CDN) then skip their default caching and compression rules, and browsers download the files instead of rendering them. Reported for Azure, then confirmed on S3 and GCS in the issue thread.

## Fix

The content type is resolved per published file and passed into the upload request itself (`putObject` / `uploadWithResponse`), so no extra round trip.

- Source of truth is `ExportFormat.mediaType` — the same value the export download endpoint already sends — so the two cannot drift.
- Two cases the format alone cannot express: a zipped publish is `application/zip`, and `APPLE_STRINGS_STRINGSDICT` emits both `.strings` and `.stringsdict` from one format whose enum `mediaType` is blank.
- `; charset=UTF-8` is appended to `text/*` only. Those files are now *rendered* rather than downloaded, so without it a browser falls back to a locale-derived encoding and accented/CJK translations render as mojibake — which the CDN would then cache. JSON and XML define their own encoding detection; binary types must not carry a charset.

### Apple path resolution

`fileStructureTemplate` only has to *contain* `{extension}`, not place it anywhere particular, so the placeholder can render into a directory (`strings/en`), carry a suffix (`Localizable.stringsdict.bak`) or sit against another character (`en_strings`). Candidates are ranked: file name before directories, exact dot-segment before a fuzzy word match. Two equally exact directory segments resolve to *nothing* rather than a guess, since a wrong Content-Type is worse than none.

The resolver is given the export-relative path, never the user-controlled slug. A publish that cannot recover the extension keeps today's behaviour and logs a warning naming the config, format and paths (scrubbed of non-printables, since templates are user-authored).

GCS users are covered by the S3 path, which is what its interoperability endpoint speaks.

## Tests

- `ResolveExportContentTypeTest` — exhaustive `ExportFormat` → expected table, asserted complete against `ExportFormat.entries` so a new format forces an explicit decision.
- `GetAppleExportPathContentTypeTest` — 17 cases covering the template shapes above, the ranking, and the refusal to guess.
- `ContentDeliveryContentTypeTest` — end-to-end publish of an Apple config, asserting `.strings` and `.stringsdict` reach storage with different content types, plus the degradation case.
- `FileStorageS3Test` / `FileStorageAzureTest` — content type is really sent, and is absent when none is given; the Azure upload keeps overwrite semantics (`ifNoneMatch` unset), so republishing still works.

Also adds the `cleanTestData` teardown the project's test-data rule requires to the content-delivery test classes; without it those suites were order-dependent and failed a different subset of cases per run.

## Deliberately out of scope

- **`V2ExportController` still derives its own header.** Routing the download endpoint through this resolver would add `; charset=UTF-8` to public response headers and start emitting a type for single-file Apple downloads — a user-visible API change that belongs in its own PR.
- **`AzureBlobFileStorage.fileExists` returns `true` unconditionally** (it discards `exists()` and catches an AWS exception type Azure never throws). Real bug, unrelated to MIME types; worth a separate `fix:` with its own test.

## Summary by CodeRabbit

* **New Features**
  * Content-delivery exports now apply appropriate content types, including UTF-8 text, ZIP archives, and Apple `.strings`/`.stringsdict` files.
  * Apple export paths are analyzed to determine the correct content type, including nested directories and filename extensions.
  * Cloud storage uploads now preserve content-type metadata for Azure Blob Storage and Amazon S3.

* **Bug Fixes**
  * Improved handling of generated storage paths and custom content-delivery slugs.
  * Added safer, clearer logging for files whose content type cannot be determined.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/contentDelivery/ContentDeliveryConfigControllerTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/contentDelivery/ContentDeliveryConfigControllerTest.kt
@@ -2,14 +2,16 @@ package io.tolgee.api.v2.controllers.contentDelivery
 
 import io.tolgee.ProjectAuthControllerTest
 import io.tolgee.batch.BatchJobConcurrentLauncher
-import io.tolgee.component.fileStorage.AzureBlobFileStorage
 import io.tolgee.component.fileStorage.AzureFileStorageFactory
 import io.tolgee.component.fileStorage.FileStorage
-import io.tolgee.component.fileStorage.S3FileStorage
 import io.tolgee.component.fileStorage.S3FileStorageFactory
 import io.tolgee.development.testDataBuilder.data.ContentDeliveryConfigTestData
+import io.tolgee.fixtures.CONTENT_DELIVERY_GENERATED_SLUG_PATTERN
 import io.tolgee.fixtures.andIsBadRequest
 import io.tolgee.fixtures.andIsOk
+import io.tolgee.fixtures.assertPrunedSingleDirectory
+import io.tolgee.fixtures.assertStoredSingleFile
+import io.tolgee.fixtures.mockCreatedStorage
 import io.tolgee.service.contentDelivery.ContentDeliveryConfigService
 import io.tolgee.testing.ContextRecreatingTest
 import io.tolgee.testing.annotations.ProjectJWTAuthTestMethod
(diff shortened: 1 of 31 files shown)


dkrizan (dismissed):

Diggsey:
Would love to see this merged!

dkrizan (approved):

Anty0:
The PR was reverted. It caused regressions (and also wasn't really ready - I was still working on addressing some issues). I'll reopen it later when it's in a better shape. ^^

PR #3843, opened 2026-08-05
fix: size the async and streaming thread pools instead of leaving both single-threaded

JanCizmar:
## The problem

On any Tolgee instance, only **one** of these could happen at a time — for the whole server, across every user and every project:

- exporting a project
- exporting a glossary
- importing (the progress stream runs for the entire import)
- asking for machine-translation suggestions in the editor

They all queued behind each other on a single thread. So while one person ran a large import, **everyone else's exports and MT suggestions just waited** — sometimes for minutes. If the wait went past 3 minutes the request didn't even fail cleanly: it returned a 500 and raised a Sentry alert.

The same was true of background work. E-mail sending, analytics reporting, translation-statistics updates, project deletions and live websocket updates all shared one thread, so a slow e-mail could hold up the "someone edited this key" notification another translator was waiting for.

Nobody noticed because nothing measured it. Neither pool appears in Grafana today.

## Why it happened

Both thread pools were created like this:

```kotlin
val executor = ThreadPoolTaskExecutor()   // no sizing
```

That constructor gives you `corePoolSize = 1` and an **unbounded queue**. A Java thread pool only starts a second thread when the queue refuses a task — and an unbounded queue never refuses. So the pool never grew past one thread, and the `maxPoolSize` it also sets was dead config that could never take effect.

Production evidence: `hikaricp_connections_usage_seconds_max` peaks at **~20 minutes** on production pods (vs 3–7 s on testing), because a streaming response holds its database connection for as long as the stream lasts.

## What this changes

**Both pools are now sized properly**, and the size is worked out from the database connection pool — because every streaming response holds one database connection until it finishes, so the database is the real limit, not the CPU.

| | streaming | background |
|---|---|---|
| formula | `db pool ÷ 3` (min 2) | `db pool ÷ 6` (min 2) |
| self-hosted (db pool 10) | 3 | 2 |
| Tolgee Cloud (db pool 100) | 32 | 16 |

Both are overridable with `tolgee.async.streaming.*` and `tolgee.async.background.*`. Tolgee Cloud sets them explicitly in tolgee/deployment#801 so the numbers are reviewable rather than implied.

**When the server really is at capacity, it now says so.** Instead of making the client wait three minutes and then returning a 500, an overloaded server answers straight away with `503` and `Retry-After: 5`, and the webapp shows "The server is busy right now. Please try again in a moment." A queue still absorbs normal bursts — it holds at least 50 requests regardless of how small the instance is — so this only happens under genuine overload, not during an ordinary busy moment.

**Two kinds of work stay strictly one-at-a-time**, because making them concurrent would break them:

- **Live activity updates.** The webapp applies these as ordered edits, so if two edits to the same key arrived out of order, another translator would see a stale value.
- **Automation triggers.** These de-duplicate using a "have I already queued this job?" check. Running them in parallel lets several checks all see "no" and each queue a job, so one burst of edits could fire several content-delivery publishes where the 30-second debounce exists to fire exactly one.

Both now have their own single-thread pool, keeping today's behaviour exactly.

**You can now see all of it.** Both pools are Spring beans, so Prometheus exports thread counts and queue depth for each one, plus new counters for rejected and timed-out streams and a timer for how long each kind of stream runs. As a bonus, Spring now shuts the pools down on exit — previously they were never shut down at all. The server also logs its resolved capacity at startup and warns if the pools would crowd the database connection pool.

## What users will notice

- Exports, imports and MT suggestions **no longer block each other**. On Tolgee Cloud that's up to 96 at once across the three pods, instead of 1.
- Under real overload, a fast, clear "server is busy, try again" instead of a three-minute wait ending in an error.
- No change to live updates or automations — those keep their current ordering guarantees on purpose.

## Notes for review

- **Self-hosted defaults are deliberately conservative** (3 streaming threads on the stock 10-connection pool), because raising the database pool for everyone on a minor upgrade is riskier than the bug being fixed. Operators can raise both.
- **`spring-security#9175` is not fixed here.** Higher streaming concurrency makes this pre-existing race more likely. An earlier revision fixed it by making Spring Security write headers eagerly, but that broke image and screenshot caching (`ImageStorageController` uses `addHeader`, and Security only skips when the header is already set — so responses got a duplicate `Cache-Control` plus `Pragma`/`Expires`). Reverted; it needs a fix that doesn't reorder header writing globally.
- `Message.SERVER_BUSY` is intentionally **not** added to `webapp/src/i18n/en.json` — per `webapp/CLAUDE.md` that file is generated from Tolgee at release. The key exists in the `Tolgee itself` project, tagged `draft: async-pool-sizing`.

This was reviewed over 13 rounds by an automated reviewer panel; ~175 findings, ~140 applied.

## Follow-ups (not in this PR)

1. **Stop streaming responses holding a database connection for their whole duration.** `MtResultStreamer` keeps one while it waits on DeepL/OpenAI, doing no database work at all. This is the biggest capacity win available and would remove the connection-pool coupling entirely — it needs its own connection-leak test, because the wrapper it lives in exists to fix one.
2. Accept both `spring.datasource.maximum-pool-size` spellings, or warn about the ignored one.
3. Separate pools per workload once (1) lands.
4. Revisit `spring.mvc.async.request-timeout` using the new per-stream timings.

Requires tolgee/deployment#801 for the production and testing values.
Companion: tolgee/billing#291 — makes billing's Stripe usage-invoicing idempotent (Stripe idempotency keys) and its expected-usage cache update atomic (Redisson lock); both relied on the old pool being accidentally single-threaded, and only per pod.

## Summary by CodeRabbit

- **New Features**
  - Added configurable limits for streaming and background processing.
  - Added monitoring for streaming duration, rejected requests, and queue timeouts.
  - Added startup capacity reporting to identify potential configuration issues.

- **Bug Fixes**
  - Streaming requests now return a clear “Server busy” response with a five-second retry instruction when capacity is unavailable.
  - Improved temporary export-file cleanup and request-timeout handling.
  - Preserved security and response headers for rejected streaming requests.
  - Improved reliability and ordering for background and real-time processing.


backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/administration/ProjectExportImportController.kt
```diff
@@ -56,9 +57,12 @@ class ProjectExportImportController(
   ): ResponseEntity<StreamingResponseBody> {
     val export = projectExportImportExporter.exportToTempFile(projectId, versionProvider.version)
     val tempFile = export.path
+    // A saturated streaming pool rejects the body after this method returns, so the delete below
```
JanCizmar: Unnecessary comment. Move it to separate well named method without comment.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncCapacityReporter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+import io.tolgee.configuration.tolgee.TolgeeProperties
+import io.tolgee.util.Logging
+import io.tolgee.util.logger
+import org.springframework.boot.context.event.ApplicationReadyEvent
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+@Component
+class AsyncCapacityReporter(
```
JanCizmar, line 20: Dcomuentetaion comment about what this class does would be awesome.

JanCizmar: Do we even need this class for production?

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncMethodConfiguration.kt
```diff
@@ -1,23 +1,70 @@
...
+  }
+
+  /**
+   * Single-threaded, because the webapp applies activity events as ordered deltas: two edits of the
+   * same key delivered out of order leave another user's editor showing a stale value. The queue is
+   * unbounded for the same reason — any rejection policy would either drop an event or run it on the
+   * caller ahead of everything already queued.
+   */
+  @Bean(WEBSOCKET_EXECUTOR_BEAN_NAME)
+  fun websocketAsyncExecutor(): ThreadPoolTaskExecutor =
```
JanCizmar, line 42: Well there can be a lot of websocket consumers. I am not sure whether is good idea to give it maxThreads 1, since the websocket delivery is highly dependant on the time client consumes it. We might rather not guarantee the order and protect our frontend somehow or come up with better solution to the ordering then keeping it single thread. Correct me if I am wrong.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncMethodConfiguration.kt
```diff
@@ -1,23 +1,70 @@
...
-    executor.initialize()
-    return executor
+
+  /**
+   * Also single-threaded. BatchJobService.tryDebounceJob is a find-then-insert with no lock, and the
+   * automation triggers share a debouncing key on purpose, so running two revisions at once lets
+   * both see no PENDING job and enqueue a publish the debounce exists to collapse into one.
+   */
+  @Bean(AUTOMATION_EXECUTOR_BEAN_NAME)
+  fun automationAsyncExecutor(): ThreadPoolTaskExecutor =
```
JanCizmar, line 56: I am confused. AFAIK, automations should be backed by batch jobs, not async. In the same time I cannot see any real usage of this bean except for tests...

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/administration/ProjectExportImportController.kt
JanCizmar: Done in d668fedb7 — it is now `deleteAtJvmExitIfTheStreamNeverRuns(tempFile)` and the comment is gone.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncCapacityReporter.kt
JanCizmar: Added in d668fedb7.

JanCizmar: Honestly: for production, no. Cloud pins its numbers in `values.yaml`, so the warning can never fire and the INFO line only restates what is already in Helm.

Its value is self-hosted. Those instances never set `tolgee.async.*`, so their pool sizes come from a formula against whatever their Hikari pool happens to be — invisible unless something says so. One boot line turns a support question from "let us work out what your connection pool is" into "paste your startup log". The warning then catches the one misconfiguration that actually hurts: pools sized so streaming starves ordinary requests of connections.

So it earns its keep outside Cloud, not inside it. If you would rather not carry ~50 lines for that, say so and I will drop it — nothing else depends on it.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncMethodConfiguration.kt
JanCizmar: You are right about the delivery layer, but that layer is not this pool — I checked both implementations before answering.

`onActivity` ends at `websocketEventPublisher(...)`, and neither implementation waits for a client:

- **With Redis (production):** `RedisWebsocketEventPublisher` does `redisTemplate.convertAndSend("websocket", json)` — a Redis PUBLISH. It returns as soon as Redis accepts it, regardless of what any subscriber does.
- **Without Redis:** `SimpleWebsocketEventPublisher` calls `SimpMessagingTemplate.convertAndSend`, which hands the message to Spring's `clientOutboundChannel`. That channel has its own 8-thread executor (`clientOutboundChannelExecutor`, visible in Grafana today), and per-client delivery and slow-consumer handling happen there.

So a slow client cannot block this thread. What the single thread serializes is only the assembly before the publish — fetching the author, building the model, extracting relation descriptions — which is bounded DB and CPU work, not client-paced.

On dropping the ordering guarantee instead: that is a bigger change than it looks, because the frontend applies these as ordered deltas rather than snapshots (`useWebsocketService.ts` flushes an array of them every ≤100ms). Making delivery unordered means the client needs a way to detect and recover from a gap — a sequence number and a resync — which is a real feature, not a tweak. Worth doing if this thread ever becomes a bottleneck; the new `executor_queued_tasks{name="websocketAsyncExecutor"}` metric will show that before users do.

Happy to be wrong if you know of a path where the publisher blocks on consumption — I only traced these two.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncMethodConfiguration.kt
JanCizmar: Both halves are fair, and the second one is a real readability problem.

**Usage:** it is referenced by name, not by type, which is why "find usages" finds nothing: `AutomationActivityListener` (backend/data) has `@Async("automationAsyncExecutor")` on both its `listen` methods. `AsyncMethodConfiguration` is in backend/app, so the constant cannot be shared across the module boundary and the annotation carries a string literal. `AsyncExecutorConfigurationTest.automation triggers are pinned to a serial executor` reflects on those annotations and asserts the value equals `AUTOMATION_EXECUTOR_BEAN_NAME`, so the link is pinned even though the IDE cannot see it.

**Batch jobs vs async:** you are right that the automation *work* is a batch job — `AutomationsBatchJobCreator` creates one. What runs on `@Async` is only the trigger path: react to an activity revision, work out which automations match, and create that batch job.

That path is why the pool exists. `BatchJobService.tryDebounceJob` is a find-then-insert with no lock — it looks for a PENDING job with the same debouncing key and inserts only if none exists. `CONTENT_DELIVERY_PUBLISH` deliberately shares its key across revisions and carries a 30s debounce. Until this PR every `@Async` task shared one thread, which serialized that check by accident. Widening the background pool to 16 would let several invocations each find no PENDING job and each insert one, so a burst of edits fires several content-delivery publishes where the debounce exists to fire exactly one.

Pinning the listener restores exactly the previous behaviour. The alternative is a lock around the debounce in `BatchJobService`, which is shared by every batch-job caller — a wider change than this PR should make. Happy to go that way instead if you prefer the fix at the source.

backend/app/src/test/kotlin/io/tolgee/configuration/StreamingBackpressureHttpTest.kt
```diff
@@ -0,0 +1,173 @@
...
+      .isEqualTo(rejectedBefore + 1)
+
+    response.statusCode().assert.isEqualTo(HttpStatus.SERVICE_UNAVAILABLE.value())
+    response.body().assert.contains("server_busy")
+    response
+      .headers()
+      .firstValue("Retry-After")
+      .orElse(null)
+      .assert
+      .isEqualTo("5")
```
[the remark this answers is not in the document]
JanCizmar: Fixed. The MockMvc `andAssertThatJson` helper is not usable here — this test drives a real `HttpClient` against a `RANDOM_PORT` container, because MockMvc never performs the ASYNC dispatch that Spring triggers when the task executor rejects, so it cannot observe the 503 at all.

Structural assertion instead: the body is parsed and its `code` field compared to `Message.SERVER_BUSY.code`, so malformed JSON or a wrong code now fails.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncExecutorFactory.kt
```diff
@@ -0,0 +1,95 @@
...
+      // ThreadPoolExecutor rejects allowCoreThreadTimeOut with a non-positive keep-alive.
+      setAllowCoreThreadTimeOut(keepAliveSeconds > 0)
+      setThreadNamePrefix(threadNamePrefix)
+      // Making these beans gave them a shutdown they never had. Without this, a task submitted by a
+      // request still in flight when the context starts closing is rejected into its caller.
+      setAcceptTasksAfterContextClose(true)
+      // Queued streams have a client that is already gone once the connector stops; background work
+      // does not, so only that is worth draining.
+      setWaitForTasksToCompleteOnShutdown(drainOnShutdown)
+      setAwaitTerminationSeconds(SHUTDOWN_DRAIN_SECONDS)
```
dkrizan, line 58: this will probably slowdown the shutdown, is it ok ?

having 4 pools it can last up to 80 seconds, Kubernetes default `terminationGracePeriodSeconds` is 30 seconds, so it might terminate some running work

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/AsyncProperties.kt
```diff
@@ -0,0 +1,86 @@
...
+      "Its queue is unbounded — background work is queued, never dropped and never run on the " +
+      "thread that submitted it.",
+)
+class BackgroundAsyncProperties {
+  @DocProperty(
+    description = "How many background tasks Tolgee runs in parallel on this instance.",
+    defaultValue = "-1",
+    defaultExplanation = "A sixth of the database connection pool size, never less than 2",
+  )
+  var maxThreads: Int = -1
```
dkrizan, line 84: Based on the code, setting it to 0 means ‘auto,’ I believe this should be mentioned in the description to prevent self-hosted users from accidentally setting it to auto while thinking they disabled it.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncExecutorFactory.kt
JanCizmar: Good catch — you are right, and I checked the chart: `terminationGracePeriodSeconds` is not set for the Tolgee deployment (only inside the bundled redis chart), so pods get Kubernetes' default of 30s.

There are four pools and Spring destroys them one at a time, so the 20s I had was a worst case of 80s. Past 30s the kubelet SIGKILLs mid-drain, which is the worst of both: the pod spends its entire grace period waiting and still loses the work the wait existed to protect.

Dropped to 5s per pool, so the worst case is 20s and leaves room for the rest of the shutdown (web server, JPA, Redis). Streaming already has `waitForTasksToCompleteOnShutdown = false` — a stream's client is gone once the connector stops — so in practice only the background and automation pools use much of it.

The alternative is raising `terminationGracePeriodSeconds` in the chart, but that slows every rollout for a drain that is rarely needed. Say the word if you would rather go that way.

backend/data/src/main/kotlin/io/tolgee/configuration/tolgee/AsyncProperties.kt
JanCizmar: Agreed, and this is the better framing — an earlier reviewer raised the same asymmetry and I argued it away on the grounds that 0 threads is not a valid configuration. That is true but beside the point: the risk is someone *trying* it to switch the pool off and silently getting the derived size instead.

Documented on both `streaming.max-threads` and `background.max-threads`: "There is no way to switch streaming off: `0` and any negative value mean *derive it*, not *disable it*."

Kept the behaviour as-is rather than failing on 0, because a pool of 0 threads cannot run anything — an instance that accepted it would deadlock every export and every background task instead of just logging a surprise. Note `queue-capacity: 0` is different and genuinely honoured: it means reject rather than buffer, and there is a test pinning that.

backend/app/src/main/kotlin/io/tolgee/configuration/AsyncExecutorFactory.kt
```diff
@@ -0,0 +1,101 @@
...
+import javax.sql.DataSource
+
+@Component
+class AsyncExecutorFactory(
+  private val tolgeeProperties: TolgeeProperties,
+  private val dataSourceProvider: ObjectProvider<DataSource>,
+) {
+  /** Read off the live bean, not `spring.datasource.*` — see PostgresAutoStartConfiguration. */
+  val connectionPoolSize: Int? by lazy {
+    (dataSourceProvider.ifAvailable as? HikariDataSource)?.maximumPoolSize
```
bdshadow, line 20: `getIfAvailable()` doesn't return null for "not created yet".  It forces the DataSource into existence earlier than it would otherwise be created.

backend/app/src/main/kotlin/io/tolgee/ExceptionHandlers.kt
```diff
@@ -263,6 +272,63 @@ class ExceptionHandlers : Logging {
...
+    if (request.getAttribute(StreamingResponseBodyProvider.STREAM_STARTED_ATTRIBUTE) != null) {
+      return handleOtherExceptions(ex)
+    }
+    metrics.streamingQueueTimeoutCounter.increment()
+    logger.debug("Request timed out waiting for a streaming thread", ex)
+    return serverBusy(response)
+  }
+
+  private fun isStreamingUnavailable(ex: Throwable): Boolean =
+    generateSequence(ex) { it.cause }.any { it is StreamingUnavailableException }
```
bdshadow, line 304: Check for instance is always 'false'. This will become an error in language version 2.4. See https://youtrack.jetbrains.com/issue/KTLC-365.

PR #3850, opened 2026-08-07
feat: show non-breaking and zero-width characters in the web UI

### LINKED ISSUE #3041: Show Non-breaking space differently in web ui
Hello!
I'm always frustrated when I try to understand whether I used a non-breaking space.
Current web UI doesn't differentiate ordinary space with non-breaking making it impossible to maintain.

Image

Here is a screenshot of what it looks like in IDE (Intellij IDEA)

Image

This is what it looks like in Sublime

Image

Show some special symbol like in IDE, instead of the transparent space, or highlight the space with some color.

I didn't find a solution for my pain.

dkrizan:
Closes #3041

Non-breaking and zero-width characters were indistinguishable from ordinary text everywhere in the web UI, so a translator could neither spot nor audit them. A non-breaking space looked exactly like a space; zero-width characters pasted in from Word or a CAT tool were completely undetectable.

They are now marked in both places a value appears:

- **Read-only** — grid cells (table and list), single-key view, tools panel TM/MT items, import conflict dialogs, branch-merge diffs, translation-memory entries, and key names.
- **Editor** — the same treatment as CodeMirror decorations, active in every editor mode.

The NBSP family gets a faint tint behind the space; zero-width characters get a thin bar, since they have no box to paint. Hovering a marker names the character, e.g. `No-break space (U+00A0)`.

Marked: U+00A0, U+202F, U+2007 (tint) and U+200B, U+FEFF, U+00AD (bar). Regular spaces are left alone, as are other Unicode spaces that carry no non-breaking or zero-width semantics.

### Notes

- Characters are **wrapped, never substituted**, so copying text out of a cell still yields exactly what is stored. In the editor, decorations are view-level only (`Decoration.mark` / `Decoration.widget`, never `replace`), so the document is untouched.
- Colours reuse existing theme tokens (`palette.label.lightBlue`, `palette.label.orange`). No theme file was modified — `figmaTheme.ts` is a generated Figma export.
- A shared registry (`webapp/src/fixtures/invisibleCharacters.ts`) owns which characters matter; the React and CodeMirror paths are thin adapters over it. Adding a character later is one array entry.
- The pre-existing key-name whitespace indicator (`codemirrorVisibleWhitespace.ts`) is untouched and resolves to the same colour token, so the two never disagree where they overlap.

### Known limitations

- An NBSP **inside ICU placeholder syntax** is not marked in read-only cells, because placeholders render as chips and their raw text is never emitted. The editor does show it.
- PO `msgctxt` is not marked **inside the editor** — `KeyNamePlugin` in `@tginternal/editor` renders it via `Decoration.replace`, so decorations in that range are never drawn. Read-only cells do mark it.
- The U+FEFF marker introduces an atomic inline box, creating a soft-wrap opportunity where that character exists to suppress one.

### Testing

- 17 unit tests over the registry, covering each character, boundary positions, adjacency, regular-space rejection, and code-unit offsets in surrogate-pair text.
- 6 Cypress tests covering both read-only and editor rendering plus the hover tooltip.

## Summary by CodeRabbit

* **New Features**
  * Invisible characters, including non-breaking and zero-width spaces, are now visibly marked in translation cells, editors, QA messages, glossary highlights, and translation memory.
  * Hovering over a marker displays a tooltip identifying the character.
  * Detection works alongside placeholders, modifiers, contexts, and message IDs without disrupting existing formatting.
  * Regular spaces remain unmarked.

* **Tests**
  * Added coverage for detection, rendering, editor behavior, tooltips, character boundaries, and repeated or adjacent occurrences.


e2e/cypress/e2e/translations/invisibleCharacters.cy.ts
```diff
@@ -0,0 +1,87 @@
...
+import { waitForGlobalLoading } from '../../common/loading';
+import { gcy, gcyAdvanced } from '../../common/shared';
+import { editCell, visitTranslations } from '../../common/translations';
+
+describe('Invisible characters', () => {
+  let projectId: number;
+
+  beforeEach(() => {
+    translationSingleTestData.clean();
+    translationSingleTestData.generate().then((data) => {
```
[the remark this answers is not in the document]
dkrizan: Not applicable here. \`TranslationSingleE2eDataController\` is a legacy controller — it exposes only \`/generate\` and \`/clean\`, with no \`generateStandard\` endpoint, so this call would 404. The project rule reads: "New E2E data controllers extend \`AbstractE2eDataController\`; new tests call \`.generateStandard()\`, not the deprecated \`.generate()\` (legacy controllers without the base keep \`.generate()\`)." This fixture is one of those legacy controllers, and `keyNameWhitespace.cy.ts` uses it the same way.

On `generateTestDataObject`: `translationSingleTestData` is already produced by it, in `e2e/cypress/common/apiCalls/testData/testData.ts`.

e2e/cypress/e2e/translations/invisibleCharacters.cy.ts
```diff
@@ -0,0 +1,87 @@
...
+  it('marks a non-breaking space in a translation cell', () => {
+    visitTranslations(projectId);
+    waitForGlobalLoading();
+    gcyAdvanced({
+      value: 'translations-table-cell',
+      key: 'nbsp key',
+      language: 'en',
+    })
+      .find('[data-cy="invisible-character"][data-cy-kind="nonBreakingSpace"]')
+      .should('have.length.at.least', 1);
```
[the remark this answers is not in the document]
dkrizan: Partly taken. The editor assertions were matching CodeMirror CSS class names, which are styling details rather than a test contract — fixed in 8634f49: the decorations now carry `data-cy="invisible-character-editor"` plus `data-cy-kind`, and the spec selects those instead.

Declining the rest, with reasons:
- The raw `[data-cy="..."][data-cy-kind="..."]` attribute selector inside `.find()` still targets only `data-cy` attributes, never text or structural CSS. There is no `findDcyAdvanced` helper — `findDcy` takes a single value and cannot express a second attribute — and this is the pattern the negative test in this same file already uses.
- `editCell()` is the shared helper every translations spec uses to open a cell. Its text lookup is over test-owned data this spec creates (`Bonjour`, `Zero`), which the rule discourages rather than forbids; the prohibition is on translated UI text. Replacing it would mean not using the shared helper.

webapp/src/component/editor/Editor.tsx
```diff
@@ -16,6 +16,11 @@ import {
 import { Direction } from 'tg.fixtures/getLanguageDirection';
 import { useScrollMargins } from 'tg.hooks/useScrollMargins';
 import { visibleKeyNameSpacesPlugin } from './utils/codemirrorVisibleWhitespace';
+import {
+  invisibleCharactersPlugin,
+  invisibleCharactersTooltip,
+} from './utils/codemirrorInvisibleCharacters';
```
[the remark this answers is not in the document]
dkrizan: Consistent with the file as it stands: `Editor.tsx` already imports its sibling extension relatively on line 18 (`./utils/codemirrorVisibleWhitespace`), and the new import sits directly beside it. `tg.*` aliases are used in this file for cross-tree imports (`tg.fixtures/...`, `tg.hooks/...`, `tg.component/...`), which is the distinction being followed. No lint rule forbids the relative form. Leaving it consistent with its neighbour.

webapp/src/fixtures/__tests__/invisibleCharacters.test.ts
```diff
@@ -0,0 +1,83 @@
+import {
+  findInvisibleCharacters,
+  INVISIBLE_CHARACTERS,
+} from '../invisibleCharacters';
```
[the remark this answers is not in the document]
dkrizan: This matches the established convention: every test in `webapp/src/fixtures/__tests__/` imports its subject relatively — `../keyName`, `../qaUtils`, `../organizationRole`, `../cropDiffContext`, `../isElementInput`. Switching only this one to the alias would make it the odd one out.

webapp/src/component/InvisibleCharacter.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </span>
+      ) : (
+        <StyledNonBreakingSpace
+          data-cy="invisible-character"
+          data-cy-kind={char.kind}
+        >
+          {char.value}
+        </StyledNonBreakingSpace>
+      )}
+    </Tooltip>
```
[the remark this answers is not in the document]
dkrizan: Declining, but it was independently evaluated rather than dismissed.

The interaction is real: an `InvisibleCharacter` marker inside a glossary or QA-flagged span means both tooltips open, and the small character label can render over the parent card. It is transient — moving toward the parent card dismisses the inner one after `leaveDelay` — and both tooltips are informational, so nothing is blocked and nothing is lost.

Against that, `stopPropagation` on the marker's hover and touch events carries a concrete regression risk: MUI attaches its listeners to the child element and opens on `onMouseOver`, which bubbles. A pointer entering a glossary term directly over one of its invisible characters would have that event swallowed, and the parent tooltip would never open. Trading a transient overlap for a tooltip that silently fails to appear is the worse outcome.

On the suggested component test: this codebase has no React component tests at all — `webapp/src/**/*.test.ts` is pure functions only, and there is no `@testing-library` or jsdom render setup. Adding one would introduce a new testing pattern for a cosmetic issue.

Recorded as a known limitation rather than fixed. Happy to revisit if it proves annoying in practice.

JanCizmar:
Thanks that you looked into it. ♥️

Screenshot or preview would be nice :)

dkrizan:
### What gets marked

| Character | Code point | Rendering |
|---|---|---|
| No-break space | `U+00A0` | faint blue tint behind the space |
| Narrow no-break space | `U+202F` | faint blue tint |
| Figure space | `U+2007` | faint blue tint |
| Zero-width space | `U+200B` | thin orange bar |
| Zero-width no-break space (BOM) | `U+FEFF` | thin orange bar |
| Soft hyphen | `U+00AD` | thin orange bar |

Zero-width characters get a bar rather than a tint because they occupy no width.

### One of each character

02-six-types-closeup

Rows `01`–`03` are the NBSP family, tinted. Note `02` (narrow no-break space) is a correspondingly narrow sliver — the tint tracks the character's real width. Rows `04`–`06` are the zero-width characters, each an orange bar.

### Regular spaces are never marked

03-negative-controls

Ordinary spaces, em space `U+2003`, ideographic space `U+3000` and tabs all render untouched. Only characters that are genuinely non-breaking or zero-width are flagged, so the marker keeps its meaning.

### Grid overview

01-grid-overview

How it reads in a normal translation list.

### Real French typography

04-french-column

French requires a no-break space before `?` and inside `« »`. The English row above is unmarked; the French row below shows four legitimate NBSPs, one of them directly against an ICU placeholder chip.

### Inside the editor

05-editor-markers

`invisible.13 all six in a row` opened for editing. The three adjacent NBSP-family characters merge into one blue block; the three zero-width characters render as orange bars.

The **red dots** between the bars are not from this PR, they are CodeMirror's built-in `highlightSpecialChars()`, which ships as part of `minimalSetup` and already flagged `U+00AD`, `U+200B` and `U+FEFF` before this change.

### Hover names the character

06-hover-tooltip

Hovering any marker identifies it by name and code point. The pink outline is just the editor's focused state.

### Key names too

07-key-name-markers

Invisible characters in the key name itself are marked in the key column.

webapp/src/fixtures/invisibleCharacters.ts
bdshadow, line 1: i think this must be a part of the https://github.com/tolgee/editor/ repo

webapp/src/component/editor/utils/codemirrorInvisibleCharacters.ts
bdshadow, line 1: this one should probably also be part of the https://github.com/tolgee/editor/ repo

webapp/src/ee/translationMemory/components/content/TmEntryText.tsx
```diff
@@ -44,7 +45,7 @@ export const TmEntryText: React.VFC<Props> = ({ text, locale }) => {
   if (placeholders.length === 0) {
     return (
       <StyledPlaceholdersWrapper dir={direction} lang={locale}>
-        {text}
+        {renderWithInvisibleCharacters(text, 'tm-entry-text')}
```
bdshadow: shouldn't `renderWithInvisibleCharacters` be also wrapped into `useMemo` the same way it is done for the placholders above? So that it is not run per cell per render across the whole grid?

webapp/src/component/editor/utils/codemirrorInvisibleCharacters.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export const invisibleCharactersPlugin = (): Extension[] => [
+  invisibleCharactersField,
+];
+
+export const invisibleCharactersTooltip = (
+  getLabel: (char: InvisibleChar) => string
+): Extension =>
+  hoverTooltip((context, pos, side) => {
+    const found = findInvisibleCharacters(context.state.doc.toString()).find(
```
bdshadow: `findInvisibleCharacters` is called on each hover instead of reading the decoration set the StateField already maintains.

The field already computes the full list on every doc change; the tooltip just can't see it, because the field's value is narrowed to a DecorationSet which no longer carries the InvisibleChar. Widen the field's state to keep both:
```ts
type Found = { index: number; char: InvisibleChar };

type InvisibleCharactersState = {
  found: Found[];
  decorations: DecorationSet;
};

function buildState(state: EditorState): InvisibleCharactersState {
  const found = findInvisibleCharacters(state.doc.toString());
  const builder = new RangeSetBuilder<Decoration>();
  found.forEach(({ index, char }) => {
    if (char.kind === 'zeroWidth') {
      builder.add(index, index, zeroWidthDecoration);
    } else {
      builder.add(index, index + char.value.length, nonBreakingSpaceDecoration);
    }
  });
  return { found, decorations: builder.finish() };
}

const invisibleCharactersField = StateField.define<InvisibleCharactersState>({
  create: buildState,
  update(value, tr) {
    return tr.docChanged ? buildState(tr.state) : value;
  },
  provide: (field) =>
    EditorView.decorations.from(field, (value) => value.decorations),
});
```

EditorView.decorations.from takes a second argument that projects the field's value down to the DecorationSet, so the field can hold whatever you like.

The tooltip then reads it instead of re-deriving it:
```ts
hoverTooltip((view, pos, side) => {
  const state = view.state.field(invisibleCharactersField, false);
  const hit = state?.found.find(({ index, char }) =>
    side < 0
      ? pos > index && pos <= index + char.value.length
      : pos >= index && pos < index + char.value.length
  );
  if (!hit) {
    return null;
  }
  return {
    pos: hit.index,
    end: hit.index + hit.char.value.length,
    above: true,
    create: () => { /* unchanged */ },
  };
});
```

e2e/cypress/e2e/translations/invisibleCharacters.cy.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it('marks a zero-width space in a translation cell', () => {
+    visitTranslations(projectId);
+    waitForGlobalLoading();
+    gcyAdvanced({
+      value: 'translations-table-cell',
+      key: 'zero width key',
+      language: 'en',
+    })
+      .find('[data-cy="invisible-character"][data-cy-kind="zeroWidth"]')
```
bdshadow: we have a rule not to use raw CSS in such cases.

Currently, there's no chained equivalent of `gcyAdvanced`. Probably makes sense to add a chainable version of it:

```ts
// cypress/support/registerCommands.ts
Cypress.Commands.add(
  'findDcyAdvanced',
  { prevSubject: 'element' },
  (subject, { value, ...other }) =>
    cy.wrap(subject).find(
      `[data-cy="${value}"]${Object.entries(other)
        .map(([key, v]) => `[data-cy-${key}="${v}"]`)
        .join('')}`
    )
);

// cypress/support/custom.d.ts
findDcyAdvanced(selector: { value: Value; [key: string]: string }): Chainable;

gcyAdvanced({ value: 'translations-table-cell', key: 'nbsp key', language: 'en' })
  .findDcyAdvanced({ value: 'invisible-character', kind: 'nonBreakingSpace' })
  .should('have.length.at.least', 1);
```

bdshadow (changes requested):

webapp/src/fixtures/invisibleCharacters.ts
dkrizan: I deliberately choose to not add it to editor, so I dont need to touch two repos and keep the solution simplier. But if you think it is better approach - I can do it.

bdshadow: i would move it there. 
Yes, the current solution is simpler, but we already have this split into 2 repos, and if we break i once, we'll break it next time too, which means that we don't need this separation. 
I think we can discuss if we want to keep the editor repo in the future, or if we want to move everything to platform. But while it exists, i think the correct way is to separate it

webapp/src/ee/translationMemory/components/content/TmEntryText.tsx
dkrizan: Good catch, done in b6ca518.

webapp/src/component/editor/utils/codemirrorInvisibleCharacters.ts
dkrizan: Done in b6ca518, thanks for the snippet.

e2e/cypress/e2e/translations/invisibleCharacters.cy.ts
dkrizan: Added `findDcyAdvanced` next to `findDcy`, spec uses it everywhere now. b6ca518

webapp/src/fixtures/invisibleCharacters.ts
dkrizan: You're right. `getPlaceholders` is already there and used by both the editor and the read-only rendering, so this fits the same way. Moved: https://github.com/tolgee/editor/pull/11

webapp/src/component/editor/utils/codemirrorInvisibleCharacters.ts
dkrizan: Moved in the same PR.

e2e/cypress/e2e/translations/invisibleCharacters.cy.ts
dkrizan: Reverted part of this in 0f4a96c. The editor markers select by CSS class again, not `data-cy`.

Reason: the markers are CodeMirror decorations, and once they moved to the editor package the `data-cy` moved with them. `@tginternal/editor` has no `data-cy` anywhere, so to keep `dataCyType.d.ts` complete the platform codegen would have had to scan `node_modules/@tginternal/editor/src` — both the script and the vite plugin. That felt like the wrong direction: `data-cy` is the app's contract, and `cm-keyname-space-indicator` is already identified by class.

The read-only markers are platform components and keep proper `data-cy`, so `findDcyAdvanced` stays and is used there. Only the three editor assertions use classes, with a comment saying why.

Worth noting the class selectors are what passed CI originally — I only switched them to `data-cy` for a CodeRabbit suggestion, nothing needed it.

dkrizan:
Editor side is merged and released as 1.20.0, so this is unblocked — bumped in 0757b8b and out of draft.

The registry, the CodeMirror plugin and the marker styles now live in `@tginternal/editor`; what stays here is the React marker component and the translations. All threads resolved.

@bdshadow ready for another look when you have a moment.

JanCizmar (approved):

bdshadow (approved):

JanCizmar:
Mega cool, thx 👍

PR #3853, opened 2026-08-11
fix: hard-delete project fails on branch merge, snapshot and content delivery FKs

bdshadow:
## Problem

The orphan-project purge scheduler (`DeletedOrganizationProjectPurgeScheduler` → `ProjectHardDeletingService.hardDeleteProject`) keeps failing in production, so projects of deleted organizations pile up (`deleted_at` set on the org, projects never removed). The scheduler processes projects in a batch, so a single bad project aborts the whole run.

Sentry showed:

```
update or delete on table "branch" violates foreign key constraint
"FK5jtc226unswiodyqp8r4isds6" on table "content_delivery_config"
  at io.tolgee.service.project.ProjectHardDeletingService.hardDeleteProject
```

Rather than fix that one constraint, I audited every FK into `branch` and `key` in `schema.xml`. All are non-deferrable with no `ON DELETE` action, and four were unhandled — fixing only the reported one would have left the scheduler dying on the next project.

## Fix

Deleting a project's branches is what breaks, so **`BranchService.deleteAllByProjectId` now clears everything that references them** before removing the branch rows:

| Constraint | Problem |
|---|---|
| `content_delivery_config.branch_id → branch` | never detached (the config itself is project-level: the hard delete removes it later, the mirror import keeps it) |
| `branch_merge.{source,target}_branch_id → branch` | never deleted |
| `branch_key_snapshot.branch_id → branch` (+ `branch_translation_snapshot` / `branch_key_meta_snapshot` children) | never deleted. `branch_key_snapshot.project_id → project` also blocked the final project-row delete |
| `branch.origin_branch_id → branch` (self-FK) | `deleteAllByProjectId` is a derived delete issuing one `DELETE` per row in unspecified order |

Keys, imports, tasks and language stats also FK `branch`; those stay the caller's job, since they are deleted by their own services well before branches.

Separately, `branch_merge_change` FKs `key`. Both the single- and multi-key delete paths already cleared it via `BranchMergeService.deleteChangesByKeyIds`; only `KeyService.deleteAllByProject` didn't. It now does the same by project — fixing every caller of it, not just the hard delete.

**`ProjectHardDeletingService` is unchanged.** Its original call ordering was always fine; the referencing rows simply weren't being cleared. `ProjectContentClearer` only loses code — the three private helpers it had grown for this same FK chain are gone, absorbed into the branch service.

`BranchService.deleteBranchMergeChangesByKeyIds` was a no-op that nothing overrode and nothing called — apparently the hook intended for this cleanup, never wired up. Removed (verified unused across all Tolgee repos; billing compiles clean against this branch).

## Where the queries live

Each delete is a project-scoped query on the repository that owns the entity it removes, so `AbstractBranchService` holds no SQL and needs no `EntityManager`:

| Entity | Repository |
|---|---|
| `BranchMergeChange` | `BranchMergeChangeRepositoryOss` (existing) |
| `BranchMerge` | `BranchMergeRepositoryOss` (new OSS base) |
| `KeySnapshot` | `KeySnapshotRepositoryOss` (new OSS base) |
| `TranslationSnapshot`, `KeyMetaSnapshot` | own repositories (new) |
| `ContentDeliveryConfig`, `Branch` | existing repositories |

`BranchMerge` and `KeySnapshot` previously had EE-only repositories, which OSS can't reach — the EE ones now extend the new OSS bases, matching how `BranchMergeChangeRepository` and `BranchRepository` already work. The two snapshot children get their own repositories rather than methods on `KeySnapshotRepositoryOss`, following `KeyComment` / `KeyCodeReference` (also owned children with `cascade = ALL`). Bulk JPQL bypasses those cascade mappings, which is why the child deletes are explicit — the same reason EE's `BranchSnapshotService` deletes them explicitly too.

## Tests

`ProjectHardDeletingServiceTest`:
- `deletes project with content delivery configs on branches` — `ContentDeliveryConfigBranchingTestData`
- `deletes project with branch merges and snapshots` — `ProjectImportTargetTestData`, which carries merge, snapshot and origin-branch rows

Both reproduce the FK violation without the fix (verified by reverting and re-running: `PSQLException`) and pass with it. The export/import suite and the `clear strategy` policy guard cover the shared `ProjectContentClearer` path.

## Supersedes #3832

#3832 covered the same purge path. Its soft-deleted-import and task gaps have since landed on `main` via #3804 (`ImportService.deleteAllByProject` now queries including-deleted; `AbstractTaskService.deleteAllByProjectId` clears notifications, `task_assignees` and `task_key`). Its only surviving piece — the `branch_merge_change` cleanup — is included here, so #3832 is closed.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/service/project/ProjectHardDeletingServiceTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/service/project/ProjectHardDeletingServiceTest.kt
@@ -10,8 +10,10 @@ import io.tolgee.batch.data.BatchJobType
 import io.tolgee.batch.request.DeleteKeysRequest
 import io.tolgee.development.testDataBuilder.data.BaseTestData
 import io.tolgee.development.testDataBuilder.data.BatchJobsTestData
+import io.tolgee.development.testDataBuilder.data.ContentDeliveryConfigBranchingTestData
 import io.tolgee.development.testDataBuilder.data.ContentDeliveryConfigTestData
 import io.tolgee.development.testDataBuilder.data.MtSettingsTestData
+import io.tolgee.development.testDataBuilder.data.ProjectImportTargetTestData
 import io.tolgee.development.testDataBuilder.data.ProjectWithQaEntitiesTestData
 import io.tolgee.development.testDataBuilder.data.SuggestionsTestData
 import io.tolgee.development.testDataBuilder.data.TaskTestData
@@ -130,6 +132,32 @@ class ProjectHardDeletingServiceTest : AbstractSpringTest() {
     testDataService.saveTestData(testData.root)
     executeInNewRepeatableTransaction(platformTransactionManager) {
       projectHardDeletingService.hardDeleteProject(testData.projectBuilder.self.refresh())
+    }
+  }
+
+  @Test
+  fun `deletes project with content delivery configs on branches`() {
+    val testData = ContentDeliveryConfigBranchingTestData()
+    testDataService.saveTestData(testData.root)
+    executeInNewRepeatableTransaction(platformTransactionManager) {
+      projectHardDeletingService.hardDeleteProject(testData.projectBuilder.self.refresh())
+    }
+
+    executeInNewTransaction {
+      projectService.find(testData.projectBuilder.self.id).assert.isNull()
+    }
+  }
+
+  @Test
+  fun `deletes project with branch merges and snapshots`() {
+    val testData = ProjectImportTargetTestData()
+    testDataService.saveTestData(testData.root)
+    executeInNewRepeatableTransaction(platformTransactionManager) {
+      projectHardDeletingService.hardDeleteProject(testData.targetProject.refresh())
+    }
+
+    executeInNewTransaction {
+      projectService.find(testData.targetProject.id).assert.isNull()
     }
   }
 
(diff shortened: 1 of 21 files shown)


backend/data/src/main/kotlin/io/tolgee/service/branching/ProjectBranchingCleaner.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.tolgee.repository.branching.BranchMergeChangeRepositoryOss
+import jakarta.persistence.EntityManager
+import org.springframework.stereotype.Service
+
+/**
+ * Branching rows referencing a project's keys and branches. None of these FKs has an `ON DELETE`
+ * action, so every caller must clear them before deleting the keys and branches they point at.
+ */
+@Service
+class ProjectBranchingCleaner(
```
dkrizan: I don't think we need a new service to clear branches. `BranchService` already does that.

`branchService.deleteAllByProjectId(projectId)` should be one call that just works, so nobody has to remember which foreign keys can break

Merge changes: `BranchMergeService` already deletes them by key ids, and both
key delete paths call it. Only `deleteAllByProject` in `KeyService` misses it.
A by-project variant there fixes it for every caller.

Everything else could be inside `BranchService.deleteAllByProjectId`. Before it
deletes the branch rows, it should clear what reference them: the merges, the
snapshots, the origin reference, etc.

Then `ProjectHardDeletingService` shouldn't need any reordering.

dkrizan (approved):

PR #3861, opened 2026-08-19
feat: apps domain model & schema (apps alpha 1/7)

JanCizmar:
First slice of the Tolgee Apps alpha, re-staged for review. This series lands on the `feat-apps-alpha` integration branch; each PR compiles and passes tests on its own.

## What's here

- **Entities**: `App` (registered once per server, owned by an organization, holds the app-level credentials and manifest snapshot incl. `icon`/`version`), `AppInstall` (one per installing organization, with its own principal account and `grantedScopes` as the consent record), `AppSecret` (hashed, `prefix…suffix` display, lazy expiry for roll-with-grace rotation), `AppEnabledForProject`, and `AppManifestFailureKind`.
- **`UserAccount.isAppPrincipal`** — marks an install's own account; the people-facing query filters (sign-in, seats, listings, the org-migration job) land together with the principal service in a later slice.
- **Repositories** for all of the above, including the paged/searched installing-organizations query used by the owner's installations view.
- **One consolidated Liquibase changeset** (`1787000000000-1..6`) instead of the nine incremental ones from development.

## Not here yet (later slices)

Manifest fetching/validation (2), registration & installs incl. limits (3), credentials & token flow (4), lifecycle channel & self-registration (5), manifest health & re-consent (6), activity/analytics incl. the `activity_revision.app_install_id` column and entity annotations (7).

## Notes for review

- Activity annotations are deliberately absent from the entities here so this slice is pure data model; they join in slice 7 with the ActivityTypes that reference these classes.
- The preview environment DB needs `recreateDb` on its next deploy (it carries checksums of the squashed dev changesets); production has never seen any of these changesets.

## Summary by CodeRabbit

* **New Features**
  * Added foundational support for registering and installing applications.
  * Added app availability across organizations and project-level enablement.
  * Added management of app permissions, credentials, secrets, and installation metadata.
  * Added tracking for app manifest health and failure status.
  * Added support for installation-specific principal accounts.
* **Data & Security**
  * Added secure storage and lifecycle tracking for app secrets.
  * Added database support for applications, installations, granted permissions, and project associations.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
@@ -139,13 +139,9 @@ data class UserAccount(
   var isDemo: Boolean = false
 
   /**
-   * Whether this account is an app install acting as itself rather than a person — see
-   * [io.tolgee.model.apps.AppInstall.principal].
-   *
-   * It is a live, enabled account so that everything resolving the current principal to a row keeps
-   * working, which is precisely why it must be filtered out of every query that counts or lists
-   * people: seats, the administration user list, the sign-in lookups, and the migration job that
-   * hands an organization to every user without one.
+   * An app install acting as itself rather than a person — see
+   * [io.tolgee.model.apps.AppInstall.principal]. Must be filtered out of every query that counts
+   * or lists people (seats, user listings, sign-in lookups, the org-backfill job).
    */
   @Column(name = "is_app_principal", nullable = false)
   @ColumnDefault("false")
--- a/backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
@@ -138,6 +138,19 @@ data class UserAccount(
   @ColumnDefault("false")
   var isDemo: Boolean = false
 
+  /**
+   * Whether this account is an app install acting as itself rather than a person — see
+   * [io.tolgee.model.apps.AppInstall.principal].
+   *
+   * It is a live, enabled account so that everything resolving the current principal to a row keeps
+   * working, which is precisely why it must be filtered out of every query that counts or lists
+   * people: seats, the administration user list, the sign-in lookups, and the migration job that
+   * hands an organization to every user without one.
+   */
+  @Column(name = "is_app_principal", nullable = false)
+  @ColumnDefault("false")
+  var isAppPrincipal: Boolean = false
+
   @OneToMany(mappedBy = "userAccount", fetch = FetchType.LAZY, orphanRemoval = true)
   var slackUserConnection: MutableList<SlackUserConnection> = mutableListOf()
 
--- a/backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
@@ -1,8 +1,8 @@
 package io.tolgee.model.apps
 
+import io.hypersistence.utils.hibernate.type.json.JsonBinaryType
 import io.tolgee.model.Organization
 import io.tolgee.model.StandardAuditModel
-import io.tolgee.model.UserAccount
 import jakarta.persistence.Column
 import jakarta.persistence.Entity
 import jakarta.persistence.EnumType
@@ -14,13 +14,13 @@ import jakarta.persistence.OneToMany
 import jakarta.persistence.Table
 import jakarta.persistence.UniqueConstraint
 import org.hibernate.annotations.ColumnDefault
+import org.hibernate.annotations.Type
 import java.util.Date
 
 /**
- * A published app, registered once and installed by any number of organizations. Its credentials
- * are the app's only long-lived ones: the token endpoint exchanges them plus an install id for the
- * short-lived tokens that reach an organization's data, so revoking them — which stamps
- * [tokensInvalidBefore] — cuts the app off everywhere at once.
+ * A published app, registered once per server and installed by any number of organizations. The
+ * registering organization owns it: it holds the app-level credentials and can remove the app from
+ * every organization that installed it.
  */
 @Entity
 @Table(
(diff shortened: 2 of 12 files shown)


JanCizmar:
@coderabbitai review

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
```diff
@@ -0,0 +1,147 @@
...
+  @Column(name = "app_id", nullable = false)
+  lateinit var appId: String
+
+  @Column(nullable = false)
+  lateinit var manifestUrl: String
+
+  /**
+   * The scopes the manifest currently requests, comma-joined — refreshed on registration and by the
+   * periodic manifest check. Diffed against each install's granted scopes to tell an organization
+   * that the app is asking for more than it consented to. Null until the manifest was first read.
```
JanCizmar: Null until the manifest was first read? I guess manifest is read when the app is installed immediately, so how can it be null?

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
```diff
@@ -0,0 +1,147 @@
...
+  @Column(nullable = false)
+  @ColumnDefault("''")
+  var version: String = ""
+
+  /** See [io.tolgee.dtos.apps.AppManifest.icon]; image URLs are stored already resolved. */
+  @Column(name = "icon", length = 500)
+  var icon: String? = null
+
+  /**
+   * Null for an app backfilled from an install that predates this table: no app-level credential was
```
JanCizmar: We just creating first alpha of apps, so how can some install predate this table?

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
```diff
@@ -0,0 +1,147 @@
...
+   */
+  @Column(name = "client_id", length = 64)
+  var clientId: String? = null
+
+  /**
+   * Signs outbound lifecycle deliveries, so it is stored in plaintext — the same trade-off
+   * [io.tolgee.model.webhook.WebhookConfig.webhookSecret] already makes. It authenticates
+   * nothing towards Tolgee. Null on a backfilled app until the first delivery mints one.
+   */
+  @Column(name = "webhook_secret")
```
JanCizmar: Why do you specify the column names everywhere when they are auto generated fine?

backend/data/src/main/kotlin/io/tolgee/model/apps/AppSecret.kt
```diff
@@ -0,0 +1,62 @@
...
+)
+class AppSecret : StandardAuditModel() {
+  @ManyToOne(fetch = FetchType.LAZY, optional = false)
+  lateinit var app: App
+
+  /** The hash never belongs in activity data — the prefix and suffix below identify the secret. */
+  @Column(name = "secret_hash", length = 128, nullable = false)
+  lateinit var secretHash: String
+
+  @Column(name = "secret_prefix", length = 16, nullable = false)
```
JanCizmar: Secret prefix and suffix can be simply stored in single "name" field.

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
```diff
@@ -0,0 +1,147 @@
...
+  @ColumnDefault("0")
+  var manifestFailureCount: Int = 0
+
+  @Column(name = "manifest_first_failed_at")
+  var manifestFirstFailedAt: Date? = null
+
+  @Column(name = "manifest_last_error", length = 500)
+  var manifestLastError: String? = null
+
+  /**
```
JanCizmar: This is self exmplanatory. Please reveiew all the comments and only comment when the field name is not self-explanatory. Try to keep the comment shorter and easy to read and understand.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
```diff
@@ -0,0 +1,103 @@
...
+  ],
+  indexes = [
+    Index(columnList = "organization_id"),
+    Index(columnList = "author_id"),
+    Index(columnList = "registered_app_id"),
+    Index(columnList = "principal_id"),
+  ],
+)
+class AppInstall : StandardAuditModel() {
+  /**
```
JanCizmar: Self-explanatory.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
```diff
@@ -0,0 +1,103 @@
...
+  @ManyToOne(fetch = FetchType.LAZY, optional = false)
+  lateinit var organization: Organization
+
+  /**
+   * Who registered the install. A historical "created by" record only — an organization's app
+   * outlives the employee who created it, so nothing operational may depend on this account still
+   * existing or being enabled. What the install may do comes from [grantedScopes].
+   */
+  @ManyToOne(fetch = FetchType.LAZY, optional = false)
+  lateinit var author: UserAccount
```
JanCizmar: This should be stored in activity, but we don't need the relation

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
```diff
@@ -0,0 +1,103 @@
...
+   * import's, a batch job's — works without the person who registered the install still being
+   * around, and so that it is obvious in the data who wrote what.
+   *
+   * It carries no organization role and no project permission, so it grants nothing: the install's
+   * capability is [grantedScopes] alone. It cannot sign in and is excluded from seats and user
+   * listings — see [UserAccount.isAppPrincipal].
+   */
+  @ManyToOne(fetch = FetchType.LAZY, optional = false)
+  @JoinColumn(name = "principal_id")
+  lateinit var principal: UserAccount
```
JanCizmar, line 53: Is this common practice to create an user account for each app? It doesn't neem right to me. It seems to me like that we should change the architecture that apps can do stuff, but not create user account for every app. However, if this is common practice then it makes sense. And maybe it's just much more easy then adding the possible actor everywhere in the code. What do you think?

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
```diff
@@ -0,0 +1,103 @@
...
+
+  @Column(nullable = false)
+  lateinit var manifestUrl: String
```
JanCizmar: App stores manifestUrl, so why does AppInstall? Same with the other fields below...

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
```diff
@@ -0,0 +1,103 @@
...
+  lateinit var version: String
+
+  @Column(nullable = false)
+  lateinit var baseUrl: String
+
+  /** See [io.tolgee.dtos.apps.AppManifest.icon]; image URLs are stored already resolved. */
+  @Column(name = "icon", length = 500)
+  var icon: String? = null
+
+  /** The whole manifest — far too large for activity data; the scalar fields above cover it. */
```
JanCizmar: why manifest is stored on install level? Why do we need to store manifest at all?

JanCizmar: For storing jsons, we should use JSONB, not text

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
```diff
@@ -0,0 +1,103 @@
...
+  lateinit var manifestJson: String
+
+  @Enumerated(EnumType.STRING)
+  @ElementCollection(targetClass = Scope::class, fetch = FetchType.EAGER)
+  @CollectionTable(
+    name = "app_install_granted_scope",
+    joinColumns = [JoinColumn(name = "app_install_id")],
+  )
+  @Column(name = "scope", nullable = false)
+  var grantedScopes: MutableSet<Scope> = mutableSetOf()
```
JanCizmar: Maybe storing scopes as array in database is more effective then using collection table.

backend/data/src/main/kotlin/io/tolgee/model/UserAccount.kt
```diff
@@ -138,6 +138,19 @@ data class UserAccount(
...
+   *
+   * It is a live, enabled account so that everything resolving the current principal to a row keeps
+   * working, which is precisely why it must be filtered out of every query that counts or lists
+   * people: seats, the administration user list, the sign-in lookups, and the migration job that
+   * hands an organization to every user without one.
+   */
+  @Column(name = "is_app_principal", nullable = false)
+  @ColumnDefault("false")
+  var isAppPrincipal: Boolean = false
+
```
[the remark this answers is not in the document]
JanCizmar: Deliberate sequencing: nothing can create an app principal before the principal service lands (slice 3 of this series), so the people-query filters ship together with it and its tests there. findActive(id) stays unfiltered for authentication, as you note.

backend/data/src/main/kotlin/io/tolgee/repository/apps/AppEnabledForProjectRepository.kt
```diff
@@ -0,0 +1,61 @@
...
+   */
+  @Query(
+    """
+    delete from AppEnabledForProject e
+    where e.appInstall.app.id = :appEntityId
+      and e.project.organizationOwner.id <> e.appInstall.app.organization.id
+    """,
+  )
+  @org.springframework.data.jpa.repository.Modifying
+  fun deleteByAppIdAndProjectOrganizationNotOwner(appEntityId: Long)
```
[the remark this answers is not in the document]
JanCizmar: Fixed in f1653f759: @Modifying(clearAutomatically = true).

backend/data/src/main/kotlin/io/tolgee/repository/apps/AppInstallRepository.kt
```diff
@@ -0,0 +1,76 @@
...
+   * projects may enable on top of its own installs.
+   */
+  @Query(
+    """
+    select i from AppInstall i
+    where i.app.availableToAllOrganizations = true and i.organization.id <> :organizationId
+    order by i.name
+    """,
+  )
+  fun findAvailableToOtherOrganizations(organizationId: Long): List<AppInstall>
```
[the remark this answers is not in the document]
JanCizmar: Dropped the query in f1653f759. The available-apps listing will be App-based (AppRepository.findAvailableToInstall) when the availability service lands in slice 3; returning another tenant's install was a leftover shape from before App existed as its own aggregate.

backend/data/src/main/kotlin/io/tolgee/repository/apps/AppInstallRepository.kt
```diff
@@ -0,0 +1,76 @@
...
+      and (:search is null
+        or lower(i.organization.name) like lower(concat('%', cast(:search as text), '%'))
+        or lower(i.organization.slug) like lower(concat('%', cast(:search as text), '%')))
+    """,
+  )
+  fun findInstallingOrganizations(
+    appEntityId: Long,
+    search: String?,
+    pageable: Pageable,
+  ): Page<Organization>
```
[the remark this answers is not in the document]
JanCizmar: Fixed in f1653f759: the JPQL now orders by organization.name, organization.id (deterministic across pages); the controller slice will not pass a competing sort.

backend/data/src/main/kotlin/io/tolgee/repository/apps/AppSecretRepository.kt
```diff
@@ -0,0 +1,31 @@
...
+  fun findAllByAppIdOrderByCreatedAtDesc(appId: Long): List<AppSecret>
+
+  fun findAllByAppIdAndRevokedAtIsNull(appId: Long): List<AppSecret>
+
+  fun findByIdAndAppId(
+    id: Long,
+    appId: Long,
+  ): AppSecret?
+
+  fun countByAppIdAndRevokedAtIsNull(appId: Long): Long
```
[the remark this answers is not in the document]
JanCizmar: Fixed in f1653f759: replaced the revoked-only pair with expiry-aware findActiveByAppId(appId, now) / countActiveByAppId(appId, now), so lazy expiry is encoded in the only queries later slices can reach for.

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
JanCizmar: You're right — registration always reads the manifest before the row exists, so the null state was fiction left over from the POC. manifestScopes is non-null now (and the comment says what it's for, not when it's null).

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
JanCizmar: It can't — that was the POC's backfill concept leaking through. Gone in f1653f759: clientId and webhookSecret are non-null, and the mint-on-demand fallbacks disappear from the later service slices too.

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
JanCizmar: Removed everywhere the name matched the generated snake_case; kept only length/columnDefinition attributes since those change the DDL.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppSecret.kt
JanCizmar: Merged into a single name column storing exactly what the UI shows (tgpubs_ab…yz), built once at issue time.

backend/data/src/main/kotlin/io/tolgee/model/apps/App.kt
JanCizmar: Did that pass over all four entities: docs remain only where the name can't carry the meaning (the principal, lazy expiry, plaintext webhook secret trade-off, manifestScopes diffing) and they're shorter.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
JanCizmar: Removed in the same doc pass.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
JanCizmar: Dropped — and applied the same reasoning to App.author and AppEnabledForProject.author: who-did-it belongs to the activity log (slice 7), no relation columns needed.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
JanCizmar: It is established practice: GitHub Apps get a bot user per installation, GitLab has service accounts for the same purpose, Slack apps act as bot users. The alternative is a polymorphic actor (userId-or-installId) on every table that records who did something — comments, imports, batch jobs, activity — which touches far more code and every future feature must remember both cases. A dedicated account row makes every existing user FK just work, and the isAppPrincipal flag plus no-role/no-permission keeps it out of people semantics. So: both common practice and the cheaper architecture here — I'd keep it.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
JanCizmar: Agreed, they were redundant since the app record became authoritative. AppInstall now carries only organization-specific state: the app FK, the principal, and grantedScopes.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
JanCizmar: Moved to App and stored once, as JSONB. We keep the manifest because the module definitions (the dashboard pages an app adds to project menus) are read from it on every render without re-fetching from the app's host; storing requested scopes alone wouldn't cover that.

backend/data/src/main/kotlin/io/tolgee/model/apps/AppInstall.kt
JanCizmar: Done — same EnumArrayType/varchar[] pattern Permission.scopes already uses, collection table gone.

PR #3862, opened 2026-08-20
feat: app manifest fetching & validation (apps alpha 2/7)

JanCizmar:
Second slice of the apps alpha series (after #3861): everything involved in reading an app-controlled manifest URL safely, before any service uses it.

## What's here

- **`appsRestTemplate`** — the HTTP client for app-controlled hosts: bounded connect/socket timeouts, redirects disabled (a followed redirect reaches an address nobody validated), and a DNS resolver that re-checks every address the client actually connects to against the SSRF guard — which is what closes DNS rebinding (validate-then-connect TOCTOU).
- **`AppManifestHttpClient`** — bounded body read plus a wall-clock deadline, so a host trickling bytes can neither OOM the server nor pin a worker thread.
- **`AppManifestFetcher`** — all validation of third-party manifest content: absolute-http URL rules, rejection of manifests served from Tolgee's own origin (the iframe sandbox stops isolating same-origin apps), scope parsing, field length limits, dashboard-page module validation, icon resolution (emoji / native icon name / image URL resolved against the base URL, http(s) only), and explicit rejection of unsupported manifest features.
- **`UrlSecurity`** — extended with the blocked-address checks the DNS resolver and validators share.
- **Config**: `tolgee.apps.enabled` and `tolgee.apps.allow-local-addresses` (dev/test escape hatch, documented as dangerous). Further properties arrive with the slices that use them.
- **Tests**: unit tests for the fetcher and the HTTP client, and a test for the rest template's SSRF/DNS behavior.

## Review focus

This is the security-critical parsing slice — the SSRF guard, the same-origin rejection, and the icon URL handling are the parts worth the slowest read.

## Summary by CodeRabbit

- **New Features**
  - Added configurable app manifest requests with timeouts and response-size limits.
  - Added support for app metadata, dashboard pages, icons, scopes, and protocol versions.
  - Added validation for manifest structure, supported modules, URLs, icons, and duplicate pages.
  - Added clear errors for unavailable, invalid, or same-origin manifests.

- **Security**
  - Blocked redirects and connections to private, reserved, and unsafe network addresses.
  - Restricted manifest icons and entries to safe HTTP(S) origins.

- **Tests**
  - Added coverage for retrieval, validation, timeouts, oversized responses, redirects, and address blocking.


backend/data/src/main/kotlin/io/tolgee/dtos/apps/AppManifest.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonProperty
+
+/**
+ * The alpha of Tolgee Apps supports exactly one module type: `project-dashboard-page`. Any other
+ * top-level property or module key a manifest declares is captured into [unknownProperties] /
+ * [AppManifestModules.unknownModules] and rejected by the fetcher, so an app author gets an explicit
+ * error rather than silently having an unsupported capability ignored.
+ */
+data class AppManifest(
```
JanCizmar: Should be called ...Dto or ...Request

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+    /** The manifest's icon with a relative image URL already resolved against the base URL. */
+    val icon: String? = null,
+  )
+
+  fun fetch(url: String): FetchResult {
+    urlSecurity.validateUrl(url, allowLocalAddresses = appsProperties.allowLocalAddresses)
+
+    val rawJson = appManifestHttpClient.fetchBody(url)
+
+    val manifest =
```
JanCizmar: to separet method

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+      try {
+        objectMapper.readValue<AppManifest>(rawJson)
+      } catch (e: Exception) {
+        throw BadRequestException(Message.APP_MANIFEST_INVALID, listOf(e.message ?: ""))
+      }
+
+    val scopes =
```
JanCizmar: To separate method.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+
+@Component
+class AppManifestFetcher(
+  private val appManifestHttpClient: AppManifestHttpClient,
+  private val objectMapper: ObjectMapper,
+  private val urlSecurity: UrlSecurity,
+  private val appsProperties: AppsProperties,
+  private val tolgeeProperties: TolgeeProperties,
+) {
+  data class FetchResult(
```
JanCizmar: move outside of this class or to the bottom

JanCizmar:
@coderabbitai review

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+    val port = if (uri.port == -1) defaultPortOf(scheme) else uri.port
+    return "$scheme://$host:$port"
+  }
+
+  private fun defaultPortOf(scheme: String): Int {
+    if (scheme == "https") return 443
+    return 80
+  }
+
+  private fun validateDashboardPages(manifest: AppManifest) {
```
JanCizmar: I would move this to separate class and split this to more methods.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * An emoji or a native icon name passes through as-is; anything containing a slash is an image
+   * URL and must resolve to absolute http(s) against the base URL — the browser will load it from
+   * the app's host directly, so a broken or non-http value must be refused here.
+   */
+  private fun resolveIcon(manifest: AppManifest): String? {
+    val icon = manifest.icon?.trim()
+    if (icon.isNullOrEmpty()) return null
+    if (icon.length > MAX_ICON_LENGTH) {
+      throw BadRequestException(
```
JanCizmar: It would be better to get all manifest erros and then throw all the errors.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+      requireResolvableHttpUrl("project-dashboard-page '${page.key}' entry", manifest.baseUrl, page.entry)
+    }
+  }
+
+  /**
+   * An emoji or a native icon name passes through as-is; anything containing a slash is an image
+   * URL and must resolve to absolute http(s) against the base URL — the browser will load it from
+   * the app's host directly, so a broken or non-http value must be refused here.
+   */
+  private fun resolveIcon(manifest: AppManifest): String? {
```
JanCizmar: I would move this to separate IconResolver class.

backend/data/src/test/kotlin/io/tolgee/unit/apps/AppManifestFetcherTest.kt
```diff
@@ -0,0 +1,198 @@
...
+      client,
+      jacksonObjectMapper(),
+      UrlSecurity(InternalProperties()),
+      AppsProperties().apply { this.allowLocalAddresses = true },
+      tolgeeProperties,
+    )
+  }
+
+  @Test
+  fun `rejects a local manifest URL by default, without ever fetching it`() {
```
JanCizmar, line 26: Tests first, helper functions, data to the bottom.

backend/app/src/main/kotlin/io/tolgee/configuration/RestTemplateConfiguration.kt
```diff
@@ -38,10 +48,70 @@ class RestTemplateConfiguration {
...
+    val httpClient =
+      HttpClientBuilder
+        .create()
+        .disableCookieManagement()
+        .disableRedirectHandling()
+        .setConnectionManager(connectionManager)
+        .setDefaultRequestConfig(
+          RequestConfig
+            .custom()
+            .setConnectionRequestTimeout(Timeout.ofMilliseconds(APPS_TIMEOUT_MS))
```
JanCizmar: We can make the timeouts configurable using property.

backend/data/src/main/kotlin/io/tolgee/dtos/apps/AppManifest.kt
JanCizmar: [claude] Renamed: `AppManifest` → `AppManifestDto`, nested types are now `AppManifestModulesDto` and `ProjectDashboardPageModuleDto`.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
JanCizmar: [claude] Split into separate private methods — `fetch` is now a thin orchestrator: validate URL → fetch body → `parseManifest` → validator → build `FetchResult`.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
JanCizmar: [claude] Done — parsing lives in its own `parseManifest` method now.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
JanCizmar: [claude] Moved `FetchResult` to the bottom of the class.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
JanCizmar: [claude] Extracted all content validation into a new `AppManifestValidator` component. `AppManifestFetcher` only orchestrates now. Added `AppManifestValidatorTest` (9 tests).

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
JanCizmar: [claude] The validator now collects every content error and throws one `APP_MANIFEST_INVALID` with all of them as params, so the app author fixes the manifest in one round. The same-origin rejection keeps its own code (`APP_MANIFEST_SAME_ORIGIN_AS_TOLGEE`) since it is a security refusal, not a fixable content error, and is checked after the content pass.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
JanCizmar: [claude] Extracted into a new `AppIconResolver` component with `validate` (feeds the collected errors) and `resolve` (produces the stored value) sharing one internal resolution. Added `AppIconResolverTest`.

backend/data/src/test/kotlin/io/tolgee/unit/apps/AppManifestFetcherTest.kt
JanCizmar: [claude] Rewrote the test class in this layout: tests first, private helpers below, companion test data at the bottom. Applied the same layout to the two new test classes.

backend/app/src/main/kotlin/io/tolgee/configuration/RestTemplateConfiguration.kt
JanCizmar: [claude] Replaced the hardcoded constant with a `tolgee.apps.request-timeout-ms` property (default 5000, documented via `@DocProperty`) wired into the apps `RestTemplate`.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+  private fun rejectTolgeeOrigin(manifest: AppManifest) {
+    val appOrigin = originOf(manifest.baseUrl) ?: return
+    val tolgeeOrigins =
+      listOfNotNull(tolgeeProperties.frontEndUrl, tolgeeProperties.backEndUrl)
+        .mapNotNull { originOf(it) }
+
+    if (appOrigin in tolgeeOrigins) {
+      throw BadRequestException(Message.APP_MANIFEST_SAME_ORIGIN_AS_TOLGEE, listOf(appOrigin))
+    }
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed. When neither `frontEndUrl` nor `backEndUrl` is configured, the validator now falls back to the origin of the current HTTP request (via `RequestContextHolder`), so the same-origin check is effective on default installations too. Covered by unit tests including the normalization cases (case, default ports, path).

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+          )
+        }
+      }
+      if (!seenKeys.add(page.key)) {
+        throw BadRequestException(
+          Message.APP_MANIFEST_INVALID,
+          listOf("duplicate project-dashboard-page key '${page.key}'"),
+        )
+      }
+      requireResolvableHttpUrl("project-dashboard-page '${page.key}' entry", manifest.baseUrl, page.entry)
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed. Each `project-dashboard-page` entry is now resolved against `baseUrl` and must stay on the app's own origin — an absolute entry pointing elsewhere (including Tolgee's origin) is rejected.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestFetcher.kt
```diff
@@ -0,0 +1,248 @@
...
+        throw BadRequestException(Message.APP_MANIFEST_INVALID, listOf("invalid icon: ${e.message}"))
+      }
+    if (!resolved.isAbsoluteHttpUrl()) {
+      throw BadRequestException(
+        Message.APP_MANIFEST_INVALID,
+        listOf("icon must resolve to an absolute http(s) URL"),
+      )
+    }
+    return resolved.toString()
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed. A slash-free icon value must not contain `:`, so `javascript:alert(1)` can no longer be stored as an "emoji". Values with a slash are resolved and must be absolute http(s) URLs.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestHttpClient.kt
```diff
@@ -0,0 +1,74 @@
...
+            throw BadRequestException(
+              Message.APP_MANIFEST_FETCH_FAILED,
+              listOf("unexpected status ${response.statusCode.value()}"),
+            )
+          }
+          readBounded(response.body)
+        })
+      } catch (e: RestClientException) {
+        throw BadRequestException(Message.APP_MANIFEST_FETCH_FAILED, listOf(e.message ?: ""))
+      }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed. The parse and fetch failures now pass the original exception as the cause to `BadRequestException`, so nothing is swallowed from logs.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+@Component
+class AppIconResolver {
+  fun validate(
+    icon: String?,
+    baseUrl: String,
+    errors: MutableList<String>,
+    field: String = "icon",
+  ) {
+    resolveInternal(icon, baseUrl).error?.let { errors.add("$field $it") }
```
JanCizmar: Don't use this pattern when we manipulate the erorrs, that are passed. Throw exception with errors instead.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
```diff
@@ -0,0 +1,99 @@
...
+    baseUrl: String,
+  ): String? = resolveInternal(icon, baseUrl).value
+
+  private fun resolveInternal(
+    rawIcon: String?,
+    baseUrl: String,
+  ): Resolution {
+    val icon = rawIcon?.trim()
+    if (icon.isNullOrEmpty()) return Resolution()
+    if (icon.length > MAX_ICON_LENGTH) {
```
JanCizmar: Instead of doinf such if statemens, extract each check to separate method. I wold mot make this class component. I would store the errors as property of this class. When after all checks, we have any errors, I would throw an exception with list of all errors.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
```diff
@@ -0,0 +1,99 @@
...
+ * an arbitrary host (an internal LAN device, a tracking beacon). Confining it to the origin the
+ * admin already trusts by installing the app removes that.
+ */
+@Component
+class AppIconResolver {
+  fun validate(
+    icon: String?,
+    baseUrl: String,
+    errors: MutableList<String>,
+    field: String = "icon",
```
JanCizmar: Can it be anythng else then icon?

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
```diff
@@ -0,0 +1,216 @@
...
+import org.springframework.web.context.request.ServletRequestAttributes
+import java.net.URI
+
+/**
+ * Validates everything a third-party manifest declares. Content problems are collected and thrown
+ * together as one [Message.APP_MANIFEST_INVALID], so the app author fixes the manifest in one
+ * round instead of one error at a time. The same-origin rejection keeps its own code — it is a
+ * security refusal, not a fixable content error.
+ */
+@Component
```
JanCizmar: Don't make it component, rather make it class function.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
```diff
@@ -0,0 +1,216 @@
...
+class AppManifestValidator(
+  private val tolgeeProperties: TolgeeProperties,
+  private val appIconResolver: AppIconResolver,
+) {
+  fun validate(manifest: AppManifestDto) {
+    val errors = mutableListOf<String>()
+    validateStringFields(manifest, errors)
```
JanCizmar: Don't fill passed property. Rather store all the errors as class property and fill it there.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestHttpClient.kt
```diff
@@ -0,0 +1,74 @@
...
+    return body ?: throw BadRequestException(Message.APP_MANIFEST_FETCH_FAILED)
+  }
+
+  /**
+   * The socket timeout only bounds the gap between two reads, so a host trickling bytes just under
+   * it can hold a worker thread for as long as it likes. The wall-clock deadline is what actually
+   * bounds that.
+   */
+  internal fun readBounded(
```
JanCizmar, line 41: This is weird. Cannot this be setup on the restTemplate level?

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
JanCizmar: [claude] Done. `AppIconResolver` no longer takes or fills a passed-in error list. A new instance validates one icon and accumulates problems in its own `errors` property; `collectErrors()` returns them. The single throw happens at the aggregator (`AppManifestValidator`), so all manifest errors — icon ones included — still come back in one `APP_MANIFEST_INVALID`.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
JanCizmar: [claude] Done. It is no longer a `@Component` — a new `AppIconResolver(icon, baseUrl)` per icon. Errors are a class property; each check is its own method (`isTooLong`, `validateSymbol`, `validateImageUrl`, `resolveImageUrl`). It doesn't throw itself so the manifest validator can gather icon errors together with the rest and throw once with the full list — happy to make it throw standalone instead if you'd rather, but that would split icon errors from the single manifest rejection.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
JanCizmar: [claude] Right — within the resolver it's always just an icon, so I removed the `field` parameter entirely. The caller adds the context: the app icon is prefixed `icon …` and a page icon `project-dashboard-page '<key>' icon …`.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
JanCizmar: [claude] Done — dropped `@Component`. It's a plain class now, instantiated per manifest (`AppManifestValidator(manifest, tolgeeProperties).validate()` from the fetcher).

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
JanCizmar: [claude] Done. `errors` is a class property on `AppManifestValidator`; every check method fills it directly, and `validate()` throws once at the end with all of them. No error list is passed around anymore.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestHttpClient.kt
JanCizmar: [claude] It can't be fully replaced by RestTemplate/HttpClient config, for two reasons this loop handles that no timeout setting does:

1. **Size cap.** RestTemplate buffers the whole response body into a `String` with no limit, so a malicious app could stream gigabytes into heap. We only get the 256 KB cap by reading the stream ourselves and stopping early.
2. **Wall-clock cap.** The `appsRestTemplate` already sets connect/socket/connection-request/response timeouts. But the socket (response) timeout only bounds the gap between two reads — a host that trickles one byte just under it holds the worker thread indefinitely (slowloris). Apache HttpClient 5 has no total-request-duration setting, so the wall-clock deadline in this loop is the only thing that bounds it.

Since we already have to stream for the size cap, the time budget rides along in the same loop for free. I kept both and left the comment explaining the socket-timeout gap. If you'd prefer, I can move the two constants (max size, max read time) onto `AppsProperties` so they're configurable alongside `request-timeout-ms` — say the word.

JanCizmar:
@coderabbitai review

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
```diff
@@ -0,0 +1,109 @@
...
+ */
+class AppIconResolver(
+  rawIcon: String?,
+  private val baseUrl: String,
+) {
+  private val icon = rawIcon?.trim()
+  private val errors = mutableListOf<String>()
+
+  /** Problems with the icon, empty when it is valid. */
+  fun collectErrors(): List<String> {
```
JanCizmar, line 28: It looks to me like now collectErrors doesn't execute the resolve code that actually can throw some other errors. I think that validator and fetcher should store the same instance of this class. Both validation and resolution should do one round, collect all errors. The resolution should throw too if there is any error. Have you tested that validation throws all the possible errors?

JanCizmar: [claude] Good catch, fixed the design. `AppIconResolver` now runs validation and resolution in one memoized pass: a private `validate()` runs every check once, storing both the collected errors and the resolved value. `collectErrors()` and `resolve()` both trigger that single pass, so they can't diverge, and `resolve()` throws `APP_MANIFEST_INVALID` if there is any error. The fetcher now builds ONE `AppIconResolver` instance and hands it to the validator, so validation (which merges its errors) and the fetcher's `resolve()` share the same instance and computed result — no recompute.

On testing all error paths: yes. The icon URL path (parse failure, non-http, off-origin) is exercised by `collectErrors()`, and there are tests for each: malformed URL, non-http, off-origin, markup/whitespace, over-length, scheme-in-symbol, plus a validator test proving an off-origin app icon comes back together with the other manifest errors, and a resolver test that `resolve()` throws.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
```diff
@@ -0,0 +1,109 @@
...
+    if (value.length <= MAX_ICON_LENGTH) return false
+    errors.add("exceeds $MAX_ICON_LENGTH characters")
+    return true
+  }
+
+  private fun validateSymbol(value: String) {
+    if (value.contains(':')) {
+      errors.add("must be an emoji, a native icon name, or an image URL")
+    }
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Tightened. A slash-free value is now also refused if it contains angle brackets or whitespace, so `` is rejected rather than stored. (The frontend renders slash-free icons as escaped React text or a native-icon-name lookup, never as HTML, so there was no stored-XSS path — but storing markup as an 'icon' is junk, so this enforces the contract.)

backend/data/src/main/kotlin/io/tolgee/util/UrlSecurity.kt
```diff
@@ -59,18 +59,44 @@ class UrlSecurity(
...
+  private fun isReservedIpv4(address: InetAddress): Boolean {
+    val bytes = address.address
+    if (bytes.size != 4) return false
+    val first = bytes[0].toInt() and 0xFF
+    val second = bytes[1].toInt() and 0xFF
+    if (first == 100 && second in 64..127) return true
+    if (first == 192 && second == 0 && (bytes[2].toInt() and 0xFF) == 0) return true
+    if (first == 198 && second in 18..19) return true
+    return first >= 240
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed. `isReservedIpv4` now blocks all of `0.0.0.0/8` (not just `0.0.0.0`), which is a loopback alias on Linux, plus the three documentation ranges (192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24) and 6to4-relay 192.88.99.0/24. Test extended with `0.1.2.3` and the new ranges.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
```diff
@@ -0,0 +1,211 @@
...
+  private fun parseAbsoluteHttpUrl(value: String): URI? {
+    val uri =
+      try {
+        URI(value)
+      } catch (e: Exception) {
+        return null
+      }
+    if (uri.scheme?.lowercase() !in HTTP_SCHEMES || uri.host.isNullOrBlank()) return null
+    return uri
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed by naming the unused exception `_` in `parseAbsoluteHttpUrl` (and `AppIconResolver.baseOrigin`) — detekt's SwallowedException allows `_`. These catches intentionally map a parse failure to `null` (the value is a validation input, there's no error to propagate), so there's no cause to keep here, unlike `resolveImageUrl` which does surface `e.message`.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
```diff
@@ -0,0 +1,211 @@
...
+        .mapNotNull { url -> parseAbsoluteHttpUrl(url)?.let { originOf(it) } }
+    return configured + listOfNotNull(currentRequestOrigin())
+  }
+
+  private fun currentRequestOrigin(): String? {
+    val attributes = RequestContextHolder.getRequestAttributes() as? ServletRequestAttributes ?: return null
+    val request = attributes.request
+    return parseAbsoluteHttpUrl("${request.scheme}://${request.serverName}:${request.serverPort}")
+      ?.let { originOf(it) }
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Correct, and this is the open decision I flagged earlier (@JanCizmar) — the request-origin fallback is best-effort and spoofable via the Host header. It only matters when neither `frontEndUrl` nor `backEndUrl` is set (a bare default install). Practical exploitability is low: to actually abuse a same-origin app the attacker must both spoof `Host` AND serve script from Tolgee's real origin (which they don't control) — but as a guard it does not hold on a default install, which is exactly the case it exists for.

I did not change registration semantics here because the clean fix is a product decision:
- **Option A (recommended):** require `frontEndUrl` (or `backEndUrl`) to be configured before an app can be registered, and reject registration otherwise. Deterministic, no Host trust. Cost: self-hosters must set the URL (arguably they should anyway).
- **Option B:** keep the fallback but treat it as advisory and document that `frontEndUrl` must be set for the same-origin guard to be effective.

@JanCizmar which way do you want it? I'll implement A or B on your call.

JanCizmar:
@coderabbitai review

backend/data/src/main/kotlin/io/tolgee/service/apps/AppManifestValidator.kt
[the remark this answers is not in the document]
JanCizmar: [claude] Going with Option B per Jan. Behavior is unchanged — the request-origin fallback stays — but I documented the caveat on `rejectTolgeeOrigin`: the reliable comparison is against `frontEndUrl`/`backEndUrl` and a deployment that wants this guard to hold must set one of them; the request-origin fallback is best-effort only and Host-spoofable, not a boundary to rely on. Once the apps docs exist they'll carry the same 'set frontEndUrl' note.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppIconResolver.kt
```diff
@@ -0,0 +1,121 @@
...
+    if (resolvedUri.scheme?.lowercase() !in HTTP_SCHEMES || resolvedUri.host.isNullOrBlank()) {
+      errors.add("must resolve to an absolute http(s) URL")
+      return null
+    }
+    if (originOf(resolvedUri) != baseOrigin()) {
+      errors.add("must be on the app's own origin")
+      return null
+    }
+    return resolvedUri.toString()
+  }
```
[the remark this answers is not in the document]
JanCizmar: [claude] Per Jan: keeping image-URL icons, and relaxing the same-origin restriction rather than adding it to. Loading a third-party `` cross-origin is an ordinary, safe browser primitive; the same-origin check we had actually rejected the normal case (an app hosting its logo on a CDN separate from its baseUrl), and it was redirect-bypassable anyway. The residual — a blind, opaque, no-feedback GET from the admin's browser — is accepted: it gives an attacker who can already register an app (and already runs JS in the iframe) nothing they don't already have, and can't exfiltrate anything. So there is no same-origin constraint to bypass now. Icon URLs are still required to be absolute http(s) (no `javascript:`/`data:`/markup) and length-capped.

PR #3863, opened 2026-08-20
chore: skip CI on emergency releases via pull request labels

bdshadow:
Part 2 of 2. **Merge tolgee/billing#297 first** — once this starts sending `skip-tests`, a billing that doesn't declare the input rejects the dispatch with HTTP 422.

## Problem

An emergency fix waits for the platform suite three times: on the PR, again on the push to main (a squash merge of an unmoved main reproduces the same tree), and a third time in billing. Roughly **85–120 minutes** from merge to production.

## Two labels

| Label | ① PR | ② push:main | ③ billing composed run |
|---|---|---|---|
| *(none)* | runs | runs | runs |
| `test-cloud-only` | skipped | skipped | **runs** |
| `skip-all-tests` | skipped | skipped | skipped |

`test-cloud-only` keeps the run that matters. Nothing in *this* repo tests the cloud composition — `settings.gradle` includes `:billing-app` only when `../billing/app` exists, which never holds in platform CI, so ① and ② build `server-app` without billing. Billing's run is the only one that builds platform + ee + billing, i.e. the image that actually reaches cloud. Time to production drops to **~55 min with no pre-merge wait**.

`skip-all-tests` is the 2am option and blocks nothing.

## Why labels, not a commit marker

A squash merge puts the PR title into the commit subject, and semantic-release feeds that subject straight into the changelog and the GitHub release notes — a `[skip-tests]` marker would be *published*. A label is also revocable when the emergency turns out not to be one.

## The constraint that shapes the implementation

**The run must still complete successfully with every job skipped.** Release triggers on `workflow_run` of this workflow, so suppressing the run outright (a `[skip ci]` commit) means **no release at all**. Two consequences:

- `everything-passed` had to stop counting `skipped` as failure — otherwise the run goes red and the whole chain stops. A job skipped because its dependency failed is still caught, since the dependency reports it.
- Only the eleven jobs without a `needs:` clause carry the condition; the other five skip transitively.

## Async follow-up

`Test (async)` runs the waived suite off the release path, so `skip-all-tests` no longer means "never tested" — it means "nothing *blocks* the release". It can't stop a bad release, it shortens how long one goes unnoticed to ~20 min.

There's no flag for this: what makes it asynchronous is that it's a separate run nothing subscribes to. Release matches `workflows: ["Test"]` by **exact name**, so `Test (async)` is invisible to it, and both fire on the same push. Doing it inside `Test` with `continue-on-error` wouldn't work — `workflow_run` fires when the *run* completes, not per job.

⚠️ **That name is load-bearing.** Calling this workflow `Test` would fire a second release.

`skip-all-tests` also gets a second billing dispatch with no `release-version`: billing runs the composed suite and its `trigger-release` job skips itself. Signal only, no second release.

## Commits

1. `chore: skip CI on emergency releases via pull request labels`
2. `chore: extract the emergency label lookup into a composite action` — pure refactor, no behaviour change
3. `chore: run the tests an emergency release waived, off the release path`

The extraction is split out because writing it as "no behaviour change" caught a real bug: collapsing the event switch made a manual `workflow_dispatch` of Test inherit a waiver from whatever PR last touched `HEAD` — a manual run means *run the tests*.

## Before merging

Labels need creating:

```bash
gh label create test-cloud-only --repo tolgee/tolgee-platform --color FBCA04 \
  --description "Emergency: skip platform CI; billing's composed run gates cloud"
gh label create skip-all-tests --repo tolgee/tolgee-platform --color D93F0B \
  --description "Emergency: run no tests anywhere, including the cloud image"
```

## Known sharp edges

- **Self-hosted artifacts ship untested under either label.** `release.yml` publishes `tolgee/tolgee:VERSION`, `:latest`, the tag and the GitHub release *before* it dispatches billing — so ③ can't gate them, and the without-EE configuration is covered only by ①/②. Cloud is protected; self-hosted is not. The structural fix is reordering the chain, which is separate work.
- **Label before you merge.** ②, async and billing all read labels live via the API, so labelling any time before the merge works. Labelling *after* races the push event. The PR's own run reads the frozen event payload, so it won't retroactively skip — that only wastes compute, since nothing blocks the merge.
- **A red `Test (async)` reaches nobody's review queue.** It has no PR attached; GitHub emails the actor and that's it. Worth wiring a Slack notification, or this is an audit trail rather than a safety net.
- **Worth confirming on first use** that a reusable workflow doesn't emit its own `workflow_run` event (the design assumes it doesn't). If it did, an async run would fire a redundant Release — a no-op, but `concurrency: {group: release}` on `release.yml` would be cheap insurance.

## Testing

The label matcher was exercised across 13 cases including substring traps (`no-skip-all-tests-please`, `skip-all-tests-later`, `xskip-duplicate-tests` correctly don't match), the `force-run-tests` truth table across all six input combinations, and the `everything-passed` gate across 7 result mixtures. All four YAML files parse.

### DIFF
--- a/.github/actions/resolve-test-policy/action.yaml
+++ b/.github/actions/resolve-test-policy/action.yaml
@@ -1,0 +1,68 @@
+name: Resolve test policy
+description: Reads the emergency test labels off the pull request behind a commit
+
+inputs:
+  sha:
+    description: >-
+      Commit to look up. Empty falls back to the event: payload labels on
+      pull_request, github.sha on push, none otherwise.
+    required: false
+    default: ""
+  github-token:
+    description: Token used for the pull request lookup
+    required: true
+
+outputs:
+  emergency:
+    description: true when either emergency label waived this commit's tests
+    value: ${{ steps.resolve.outputs.emergency }}
+  skip-billing-tests:
+    description: true only for skip-all-tests
+    value: ${{ steps.resolve.outputs.skip-billing-tests }}
+
+runs:
+  using: composite
+  steps:
+    - id: resolve
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        EVENT_NAME: ${{ github.event_name }}
+        PR_LABELS: ${{ join(github.event.pull_request.labels.*.name, ',') }}
+        REPO: ${{ github.repository }}
+        SHA: ${{ inputs.sha }}
+      run: |
+        set -uo pipefail
+        # A lookup error must read as "no labels", never as a waiver.
+        lookup() {
+          gh api "repos/$REPO/commits/$1/pulls" \
+            --jq '[.[].labels[].name] | join(",")' 2>/dev/null || echo ""
+        }
+
+        if [ -n "$SHA" ]; then
+          labels=$(lookup "$SHA")
+        elif [ "$EVENT_NAME" = "pull_request" ]; then
+          labels="$PR_LABELS"
+        elif [ "$EVENT_NAME" = "push" ]; then
+          labels=$(lookup "$GITHUB_SHA")
+        else
+          # A manual run means "run the tests" — never inherit a waiver from
+          # whatever pull request last touched HEAD.
+          labels=""
+        fi
+        echo "Labels: ${labels:-<none>}"
+
+        emergency=false
+        skip_billing=false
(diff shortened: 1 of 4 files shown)


bdshadow:
Closed in favour of tolgee/billing#300

PR #3869, opened 2026-08-24
feat: app credentials & token flow (apps alpha 4/7)

JanCizmar:
Slice 4 of 7 (base: `feat-apps-alpha`).

## What's here
- **Secret service** — `issue`/`expireOthers` (roll-with-grace), `revoke` (stamps `App.tokensInvalidBefore` on force, killing prior tokens), `findLiveMatching` (constant-time hash compare), first-use/last-use stamping, `MAX_LIVE_SECRETS`. Builds on slice 3's `issueInitial`/`list`.
- **Token endpoint** — `POST /v2/public/apps/token` mints an install-context (M2M) token from client-credentials; `POST /v2/projects/{id}/apps/{installId}/token` mints a user-context token for the dashboard iframe.
- **App authentication filter** — grafted into the shared chain: app tokens (audience `tg.app`) are tried first and fall through to the user-JWT path on a non-app token; revocation cutoff via `tokens_invalid_before`; acting-as-user (`X-Tolgee-Act-As-User-Id`, narrow-only) for install context.
- **Access control** — `@DenyAppAccess` (app tokens never reach app-administration routes) / `@AllowAppOwnInstallAccess` (self routes), enforced by `AppAccessInterceptor` after project authorization; `SecurityService` app scope/permission branches.
- **Self-service** — discovery, own-installations (with enabled projects), and secret issue/revoke/list for the app; operator secret list/roll/revoke on owned-apps.

## Boundary decisions (flagging)
- **`rollSecret` is delivery-free here** — roll-with-grace rotation is this slice; broadcasting the rotated secret over the lifecycle channel is slice 5, so `AppSecretRotationModel` carries no delivery yet.
- **`AppSelfRefreshController` deferred to slice 6** (manifest re-fetch + re-consent).
- Activity attribution (`@RequestActivity`) is slice 7.

## Tests
Token mint/validate, credential + token endpoint, full authorization matrix (install vs user context, scope enforcement, `@DenyAppAccess`, revocation cutoff), self-installations, secret rotation (grace/expiry/limits/force-cutoff). Base-auth regression (`ProjectApiKeyAuthenticationTest`, `ProjectPermissionFilterTest`) run green to confirm the shared filter chain is intact. All pass locally; CI not gated per the slice workflow.

## Summary by CodeRabbit

* **New Features**
  * Added OAuth 2.0 client-credentials support for obtaining short-lived, installation-scoped app access tokens.
  * Added APIs to discover app installations, enabled projects, and installation details.
  * Added project-scoped token creation with read-only support.
  * Added app secret listing, rotation, one-time issuance, expiration, and revocation.
* **Security**
  * Strengthened token validation, scope enforcement, access controls, and token invalidation.
  * Added safeguards against unauthorized administrative and MCP access.
* **Configuration**
  * Added configurable app access-token expiration, defaulting to one hour.


JanCizmar:
@coderabbitai review

JanCizmar:
@coderabbitai review

JanCizmar:
@coderabbitai review

JanCizmar:
@coderabbitai review

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppInstallationsDiscoveryController.kt
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Lets an app discover its installations from its app-level credentials alone — the entry point of
+ * the M2M flow, since minting an install-scoped token needs the install id this call returns.
+ */
+@RestController
+@CrossOrigin(origins = ["*"])
+@AppAccessNeutral
+@ConditionalOnProperty(name = ["tolgee.apps.enabled"], havingValue = "true")
+@RequestMapping(value = ["/v2/public/apps/installations"])
```
JanCizmar: Why it's listed under public, when it's not public at all?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppInstallationsDiscoveryController.kt
```diff
@@ -0,0 +1,61 @@
...
+    summary = "List the calling app's installations",
+    description =
+      "Authenticates with the app-level client credentials and returns every installation of the " +
+        "app on this server, with the projects each one is currently enabled for. The install ids " +
+        "are what the token endpoint exchanges for install-scoped access tokens.",
+  )
+  fun list(
+    @RequestBody @Valid body: AppCredentialsRequest,
+  ): CollectionModel<AppSelfInstallationModel> {
+    val app = appCredentialAuthenticator.authenticate(body.clientId, body.clientSecret)
```
JanCizmar: I believe that the only reason for providing clientSecret should be retrieving of short lived token.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Lets a published app manage its own app-level secrets unattended. Authenticates with the app-level
+ * credentials themselves, not a token — those credentials never become a session, and nothing here
+ * touches anything but the app's own secrets.
+ */
+@RestController
+@CrossOrigin(origins = ["*"])
+@AppAccessNeutral
+@ConditionalOnProperty(name = ["tolgee.apps.enabled"], havingValue = "true")
+@RequestMapping(value = ["/v2/public/apps/app-secrets"])
```
JanCizmar: Not public.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
```diff
@@ -0,0 +1,86 @@
...
+  private val appSecretService: AppSecretService,
+  private val appSecretModelAssembler: AppSecretModelAssembler,
+) {
+  @PostMapping("/list")
+  @RateLimited(5, isAuthentication = true)
+  @Operation(
+    summary = "List the calling app's own app-level secrets",
+    description = "Returns every app-level secret, revoked ones included, without disclosing any of them.",
+  )
+  fun list(
```
JanCizmar: Why would the app need to list it's secrets?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
```diff
@@ -0,0 +1,86 @@
...
+  @PostMapping("/issue")
+  @RateLimited(5, isAuthentication = true)
+  @Operation(
+    summary = "Issue an additional app-level client secret for the calling app",
+    description =
+      "Mints a fresh app-level secret and returns it — the only place it is ever disclosed. The " +
+        "secret this call authenticated with keeps working, so the app can store the new one and " +
+        "only then revoke the old one.",
+  )
+  fun issue(
```
JanCizmar: Why would app need to issue a secret for self? The secrets are delivered via webhook on the beginning. I don't think we need this.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
```diff
@@ -0,0 +1,86 @@
...
+ * credentials themselves, not a token — those credentials never become a session, and nothing here
+ * touches anything but the app's own secrets.
+ */
+@RestController
+@CrossOrigin(origins = ["*"])
+@AppAccessNeutral
+@ConditionalOnProperty(name = ["tolgee.apps.enabled"], havingValue = "true")
+@RequestMapping(value = ["/v2/public/apps/app-secrets"])
+@Tag(name = "App Self Service")
+class AppSelfAppSecretsController(
```
JanCizmar: I am not sure about this, but I believe we don't need this controller at all.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsController.kt
```diff
@@ -0,0 +1,67 @@
...
+@ConditionalOnProperty(name = ["tolgee.apps.enabled"], havingValue = "true")
+@RequestMapping(value = ["/v2/apps/self"])
+@Tag(name = "App Self Service")
+class AppSelfInstallationsController(
+  private val authenticationFacade: AuthenticationFacade,
+  private val appEnablementService: AppEnablementService,
+  private val appSelfInstallationModelAssembler: AppSelfInstallationModelAssembler,
+) {
+  @GetMapping("/installations")
+  @AllowAppOwnInstallAccess
```
JanCizmar: We should have special annotation for app level endpoints.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsController.kt
```diff
@@ -0,0 +1,67 @@
...
+    summary = "List the calling app's own installations",
+    description =
+      "Returns the install the calling install-context token belongs to, together with the " +
+        "projects the app is currently enabled for and the organization owning each of them. " +
+        "Requires a token from the client-credentials grant — a user-context token (the one the " +
+        "dashboard iframe gets) is refused, because it acts for a single user who need not be a " +
+        "member of every project the install is enabled for. A collection is returned so that an " +
+        "app holding several installs on one server stays representable.",
+  )
+  fun getSelfInstallations(): CollectionModel<AppSelfInstallationModel> {
```
JanCizmar: Listing app's installations doesn't sound like install context to me at all. It's a project level context, so why do you require the install context?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
```diff
@@ -0,0 +1,81 @@
...
+ * Client-credentials token endpoint. Modelled on the OAuth 2.0 client-credentials grant (RFC 6749
+ * §4.4) but JSON-encoded with a Tolgee error envelope, so use Tolgee's app SDK, not a stock OAuth
+ * library. An app's backend exchanges its `client_id` + `client_secret` for a short-lived
+ * install-context token, so subsequent API calls carry the token rather than the raw secret.
+ */
+@RestController
+@CrossOrigin(origins = ["*"])
+@AppAccessNeutral
+@ConditionalOnProperty(name = ["tolgee.apps.enabled"], havingValue = "true")
+@RequestMapping(value = ["/v2/public/apps"])
```
JanCizmar, line 21: Not public.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
```diff
@@ -0,0 +1,81 @@
...
+  ): AppAccessTokenModel {
+    if (body.grantType != GRANT_TYPE_CLIENT_CREDENTIALS) {
+      throw BadRequestException(Message.APP_UNSUPPORTED_GRANT_TYPE)
+    }
+
+    val app = appCredentialAuthenticator.authenticate(body.clientId, body.clientSecret)
+
+    // App-level credentials identify an app installed by any number of organizations, so they
+    // cannot imply an install on their own.
+    val installId = body.installId ?: throw BadRequestException(Message.APP_INSTALL_ID_REQUIRED)
```
JanCizmar: Without install id, we can simply mint the app level token that can authenticate the app only to do the app level operations.

JanCizmar (commented):
Found some major API issues to fix.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Done — removed the whole controller (commit 9f65343). Agreed: the app never needs to self-manage secrets (initial secret from registration + lifecycle webhook; rotation is owner-driven via owned-apps), and self-service issue-via-client_secret was itself the "a leaked secret can mint fresh secrets and survive a force-revoke" foothold the review flagged. Also removed AppSecretRevokeRequest + APP_SECRET_ID_REQUIRED.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Moot — controller removed in 9f65343 (you're right it wasn't really public either).

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Removed with the controller (9f65343). Agreed there's no reason for an app to list its own secrets.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Removed (9f65343). Agreed — secrets are delivered (registration + webhook), not self-issued.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Agreed. This is the app-level-token idea: without install_id the token endpoint mints an app-level token that authorizes only app-level operations (installation discovery), and client_secret is then only ever exchanged for a token. It is a change to the token/auth core (a third context beside install- and user-context) + a dedicated @AllowAppLevelAccess annotation, so I have not landed it yet — plan posted as a top-level comment; awaiting your go-ahead given limited credits.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppInstallationsDiscoveryController.kt
JanCizmar: [claude] Agreed — client_secret should only buy a token; every other call uses the short-lived token. Part of the app-level-token change (see plan comment).

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppInstallationsDiscoveryController.kt
JanCizmar: [claude] Agreed — once discovery authenticates with an app-level token it moves off /v2/public to a normal authenticated path. Part of the app-level-token change.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsController.kt
JanCizmar: [claude] Agreed — I'll add a dedicated @AllowAppLevelAccess (app-level context), distinct from @AllowAppOwnInstallAccess (install-context). Part of the app-level-token change.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsController.kt
JanCizmar: [claude] Agreed — listing an app's installations is app-level, not install-context; it collapses into the single app-level discovery endpoint. Part of the app-level-token change.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Mild pushback here: in this codebase /v2/public/** is the Spring-Security permit-all namespace = 'no pre-existing session required', not 'anyone'. The token endpoint is the bootstrap (you cannot already hold a token), so it must live there — like any OAuth token endpoint. The OTHER endpoints you flagged should indeed move off /public once they authenticate with a token; only /token stays.

JanCizmar:
[claude] **Plan for the app-level-token model** (your review comments on discovery / self-installations / `install_id` / "not public" / a dedicated annotation).

Design:
- **Token endpoint**: `install_id` becomes optional. No `install_id` → mint an **app-level token** (a third context beside install- and user-context) that carries the app id and authorizes only app-level operations. With `install_id` → install-context token (unchanged). `client_secret` is then only ever exchanged for a token.
- **`@AllowAppLevelAccess`** — new annotation for app-level endpoints; `AppAccessInterceptor` admits an app-level token there and denies it everywhere else (it can never reach project data: no bound project → empty scopes).
- **Discovery** authenticates with the app-level token (Bearer), drops the `client_secret` body auth, and moves off `/v2/public` to a normal authenticated path. `AppInstallationsDiscoveryController` + `AppSelfInstallationsController` collapse into **one** app-level "list my installations" endpoint.
- Only `/v2/public/apps/token` stays public (the bootstrap); `@AppAccessNeutral` drops off discovery.

This touches the shared auth core (a new token context + auth representation — `AppAuthentication` is install-centric today, so app-level needs a nullable install or a small dedicated auth object). It is the right design and net removes surface, but it is real surgery, so I have not started it — confirming direction first given the credit budget.

Open choice for you: keep a separate install-context "self" endpoint at all, or fold everything into the single app-level list? I lean fold-into-one.

JanCizmar:
[claude] **Implemented the app-level-token model** (commit fa274b9), addressing the discovery / install_id / not-public / annotation comments:

- `POST /v2/public/apps/token` now takes an **optional** `install_id`: omit it for an **app-level token** (app-level operations only), include it for an install-context token. `client_secret` is only ever exchanged for a token.
- Installation discovery moved off the client-secret body auth and off `/v2/public`: the two overlapping listing endpoints collapsed into one **`GET /v2/apps/self/installations`** behind a new **`@AllowAppLevelAccess`**. `AppInstallationsDiscoveryController` deleted.
- An app-level token identifies the app (not a person): it carries a synthetic permission-less principal, binds no project, holds no scopes, and the `AppAccessInterceptor` confines it to app-level endpoints and refuses a non-app caller there at the gate.

Only `/v2/public/apps/token` stays public (the bootstrap). All app-auth + interceptor + MCP suites green, ktlint clean.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    val token = mintToken(app, body.installId)
+    return AppAccessTokenModel(
+      accessToken = token,
+      tokenType = "Bearer",
+      expiresIn = tolgeeProperties.apps.tokenExpiration / 1000,
+    )
+  }
+
+  private fun mintToken(
```
JanCizmar: This should be on service or facade level, but not in controller.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
```diff
@@ -0,0 +1,86 @@
...
+  ): String {
+    if (installId == null) return appTokenService.mintAppLevelToken(app.id)
+    val install =
+      appInstallService.findOwnInstall(app.id, installId)
+        ?: throw NotFoundException(Message.APP_INSTALL_NOT_FOUND)
+    return appTokenService.mintInstallContextToken(install.id)
+  }
+
+  companion object {
+    private const val GRANT_TYPE_CLIENT_CREDENTIALS = "client_credentials"
```
JanCizmar: should not be in controller.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/project/ProjectAppsController.kt
```diff
@@ -99,4 +110,33 @@ class ProjectAppsController(
...
+  @ReadOnlyOperation
+  @Operation(
+    summary = "Mint a user-context app token",
+    description =
+      "Issues a short-lived JWT bound to (install, project, current user) that the dashboard iframe " +
+        "uses to call Tolgee's REST API on behalf of the user. Returns 404 if the install is not " +
+        "enabled for this project.",
+  )
+  fun mintToken(
+    @PathVariable installId: Long,
```
JanCizmar: I think we have an architecture issue here. I don't think that install should be project-level. but organization level. Project level will prevent apps from managing organizaton wide resources like glossaries / translation memories. I believe that install should simply be organization level and project access should be permission thing. App should be informed by webhook that it was enabled for project. The list of enabled projects should be accessible by the app.

JanCizmar: Due to this architectural change, I believe that we need to introduce the organization level scopes. I would use the existing Scope enum, add all the possible organization level scopes with "organizationLevel" flag, so we don't show it in the project api key UI for example.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Done (commit 4cc7827) — the grant (validation + credential auth + app-level vs install-context minting + install resolution) moved into a new AppTokenGrantService; the controller just delegates.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Done (4cc7827) — the grant-type constant and its check moved into AppTokenGrantService with the rest of the grant logic.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/project/ProjectAppsController.kt
JanCizmar: [claude] Agree there is a real gap — with one clarification of the current state: the install is ALREADY organization-level (AppInstall.organization, unique per org+app) and per-project enablement (AppEnabledForProject) is already the separate permission layer; the discovery endpoint already hands the app its enabled-projects list, and the enabled/disabled lifecycle webhook is slice 5. What is genuinely missing is an org-level access DIMENSION on the token: an install-context token today MUST bind to a project (ProjectContextService.bindAppToProject) and resolves permissions as a project-scope intersection, so it cannot reach org-level resources (glossaries, translation memories, org settings) at all. Closing that is exactly your org-level-scopes point below. It reshapes the app permission model — a core assumption of the whole feature — so rather than reopen it inside this credentials PR I'd scope it as its own focused change with a short design. Confirm the direction and I'll take it.

JanCizmar: [claude] Agree with the mechanism. Concretely: add the org-level scopes to the existing Scope enum with an organizationLevel flag; filter them out everywhere a scope set is offered in a PROJECT context (PAK create + validation, the project-API-key UI, and the project side of app manifest consent) so they never surface there; and add an org-permission path for app tokens — an install's granted org-level scopes apply on /v2/organizations/** routes without requiring project enablement, checked the same way project scopes are but at org scope. The load-bearing parts: every place that enumerates or validates Scope (PAK, manifest consent, the scope-consent UI) has to learn the flag, and SecurityService needs an org-context branch parallel to the project one (an org-binding for app tokens). It is foundational enough that I would do it as its own slice with a short design doc first — want me to draft that design?

JanCizmar:
[claude] Rebased on `feat-apps-alpha` (now including the org-scopes migration) and addressed the token-per-project comment:

**User-context tokens are now minted per organization** (commit f6e1fde5a). The token binds (install, user) with no project claim; it works on every project the install is enabled for. Each request still enforces per-project enablement plus the intersection of the install's granted scopes with the iframe user's own permissions on that project — so the token grants nothing new, it just stops being single-project.

Anti-enumeration is preserved with a reformulated rule: the accurate `app_not_enabled_for_project` error is returned only for projects the token's *own user* can see (previously: only the minted-for project); everything else — foreign orgs, nonexistent ids — stays an indistinguishable `app_access_forbidden`. An app in user context can learn nothing its user couldn't already see in the UI.

Note the install-context token was already org-level (it never carried a project), so only the iframe flow changed.

Covered by rewritten/new tests: org-wide use across two enabled projects, accurate error for a member's non-enabled project, opaque error for foreign/nonexistent projects, revocation cutoff, plus the existing narrowing/read-only/invalidations suite — all green.

JanCizmar:
[claude] Ran a focused adversarial security review of the slice-4 auth surface (three reviewers over: the org-wide token isolation, the credential/JWT/grant flow, and the request-time gates — interceptor, acting-as, MCP, super-auth). Also caught and reverted a `settings.gradle` that had a machine-specific absolute billing path committed.

**Result: no bypass, no privilege escalation, no cross-tenant data access.** Verified solid: default-deny app confinement (even mis-registered endpoints fail closed), acting-as is a true `install ∩ acted-as` intersection with membership check, MCP + super-auth denials are centralized, `listInstallations` is app-scoped, constant-time secret compare over SHA-256 of 256-bit secrets, full (clientId, secret, installId) grant cross-check, HS512 with a domain-separated key, bounded rotation. Per-project capping survives the org-wide token change (each request re-intersects with the user's live permissions on that project).

**Two issues found and fixed (commit 31443e7ee):**
1. *Medium* — app-level tokens bypassed the force-revoke cutoff and app-existence check. A leaked `client_secret` could mint an app-level token that kept enumerating installs cross-org until natural expiry, even after force-revoke — contradicting the documented immediate invalidation. The filter now loads the app, rejects a deleted one, and applies `tokensInvalidBefore`.
2. *Low* — the user-context not-enabled error was returned for any project the acting user could see in **any** org, letting an app map the user's cross-org memberships. Now scoped to the install's own organization; foreign/nonexistent projects stay indistinguishable.

**Flagged, not changed (need your call — both pre-existing / outside this diff):**
- *Design:* a plain install-context token skips per-language permission checks (grantedScopes has no language dimension), so an install with `translations.edit` edits all languages regardless of the granter's language restriction. Whether that's acceptable depends on the consent/grant slice that caps who can grant such scopes. The acting-as path does re-apply language restrictions.
- *Latent:* the app-level synthetic principal uses `id = appId`; not reachable today (app-level tokens are confined to endpoints that read only `appId`), but a future `@AllowAppLevelAccess` endpoint doing a lookup by `authenticatedUser.id` could collide with a real user id. Worth a guard when/if that surface grows.

New tests: app-level force-revoke invalidation, app-level token rejected after app deletion, and cross-org opacity even when the iframe user is a member of the foreign project. All affected suites green, ktlint clean.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationOwnedAppsController.kt
```diff
@@ -107,4 +116,71 @@ class OrganizationOwnedAppsController(
...
+    summary = "Revoke an app-level client secret",
+    description =
+      "Ends this one secret at once — how a rotation's grace window is cut short, and how a leaked " +
+        "credential is retired. Revoking the app's only active secret is refused unless `force=true`, " +
+        "so an ordinary revoke cannot cut the app off by mistake; `force` also invalidates every " +
+        "access token already minted from the app's secrets. It does not touch the app's OTHER live " +
+        "secrets, so to fully contain a compromise revoke each secret (a holder of a live secret can " +
+        "mint more) or delete the app, which retires every credential in one step. Idempotent.",
+  )
+  fun revokeSecret(
```
JanCizmar, line 173: Why do we need the force? Why would I want to rotate the only secret?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/project/ProjectAppsController.kt
```diff
@@ -99,4 +110,33 @@ class ProjectAppsController(
...
+  @PostMapping("/{installId}/token")
+  @UseDefaultPermissions
+  // Minting is a POST but does not mutate anything; a read-only (e.g. supporter) session must be able
+  // to mint, and the token it gets carries that read-only flag.
+  @ReadOnlyOperation
+  @Operation(
+    summary = "Mint a user-context app token",
+    description =
+      "Issues a short-lived JWT bound to (install, current user) that the dashboard iframe uses to " +
+        "call Tolgee's REST API on behalf of the user. The token is organization-wide: it works on " +
```
JanCizmar: When the token is organization wide now, we should not generate token in this controller and we should not use the projectId here, right. Also the verbose comment is useless. It's self explanatory that we mint it organization-wide when it's in right controller.

backend/api/src/main/kotlin/io/tolgee/hateoas/apps/AppSelfInstallationModel.kt
```diff
@@ -0,0 +1,37 @@
...
+  val appId: String,
+  val name: String,
+  val version: String,
+  @Schema(
+    description =
+      "Whether this is a native (server-level) install rather than an organization's own. Always " +
+        "false for now — native installs arrive in a later release; the field is present so app " +
+        "backends can rely on its shape without a schema change when they do.",
+  )
+  val native: Boolean,
```
JanCizmar: If it's always false I would remove it now as we don't have anything like native apps anymore.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/AppsTestData.kt
```diff
@@ -50,11 +50,19 @@ open class AppsTestData : BaseTestData("apps-test-owner@test.com", "test_project
...
-      otherProject =
+      val otherProjectBuilder =
         addProject {
           name = "other_project"
           organizationOwner = otherOwnerBuilder.defaultOrganizationBuilder.self
-        }.self
+        }
+      otherProject = otherProjectBuilder.self
+      // The main org's owner is also a member of a project in the OTHER organization, so a
+      // user-context token must still stay opaque there (no cross-org membership enumeration).
```
JanCizmar: useless comments.

backend/api/src/main/kotlin/io/tolgee/hateoas/apps/AppSelfInstallationModel.kt
```diff
@@ -0,0 +1,37 @@
...
+  val scopes: List<String>,
+  @ArraySchema(
+    arraySchema =
+      Schema(
+        description =
+          "Projects the app is currently enabled for. These are the only projects its token may " +
+            "act on; the list changes whenever a project owner enables or disables the app.",
+      ),
+  )
+  val enabledProjects: List<AppSelfEnabledProjectModel>,
```
JanCizmar: I would expose this in separate endpoint with proper pagination, not like this. Please review all the code and look for such an issues with nested lists. It's a bad practice.

backend/api/src/main/kotlin/io/tolgee/hateoas/apps/AppSelfInstallationModelAssembler.kt
```diff
@@ -0,0 +1,37 @@
...
+    enabledProjects: List<AppEnabledProjectDto>,
+  ): AppSelfInstallationModel {
+    return AppSelfInstallationModel(
+      id = install.id,
+      appId = install.app.appId,
+      name = install.app.name,
+      version = install.app.version,
+      native = native,
+      scopes = install.grantedScopes.map { it.value },
+      enabledProjects = enabledProjects.map { toProjectModel(it) },
```
JanCizmar: nested list = bad practice. Also standard project models should be used.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppCredentialAuthenticator.kt
```diff
@@ -0,0 +1,38 @@
...
+    return app
+  }
+
+  // Split out of the service so its @Async proxy on updateLastUsedAsync survives self-invocation.
+  private fun stampUse(secret: AppSecret) {
+    if (secret.lastUsedAt == null) {
+      appSecretService.recordFirstUse(secret.id)
+      return
+    }
+    appSecretService.updateLastUsedAsync(secret.id, secret.lastUsedAt)
```
JanCizmar, line 30: I have removed the first use branch, because it seems useless to me.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppSecretService.kt
```diff
@@ -32,12 +44,132 @@ class AppSecretService(
...
+    graceSeconds: Long,
+  ): RotationResult {
+    val issued = issue(app)
+    val previousExpiresAt = expireOthers(app.id, issued.secret.id, graceSeconds)
+    return RotationResult(issued, previousExpiresAt)
+  }
+
+  /** Puts every active secret but [keepSecretId] on a [graceSeconds] deadline, keeping any earlier one. */
+  @Transactional
+  fun expireOthers(
```
JanCizmar: can be private.

backend/data/src/main/kotlin/io/tolgee/service/security/SecurityService.kt
```diff
@@ -85,6 +85,13 @@ class SecurityService(
   }
 
   fun checkAnyProjectPermission(projectId: Long) {
+    if (authenticationFacade.isAppAuth) {
```
JanCizmar, line 88: Can you explain why do we need this branch?

backend/security/src/main/kotlin/io/tolgee/security/authentication/AppAccessNeutral.kt
```diff
@@ -0,0 +1,9 @@
+package io.tolgee.security.authentication
+
+/**
+ * Marks a `/v2/public/apps` endpoint that authenticates from the request body, so [AppAccessInterceptor]
+ * must ignore any bearer app token the caller also sends (the normal SDK pattern of a default
+ * `Authorization` header would otherwise 403 it). The controller still authenticates the body.
+ */
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+annotation class AppAccessNeutral
```
JanCizmar: Seems to me like we don't need this at all.

backend/security/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFilter.kt
```diff
@@ -88,7 +98,21 @@ class AuthenticationFilter(
     val authorization = request.getHeader("Authorization")
     if (authorization != null) {
       if (authorization.startsWith("Bearer ")) {
-        val auth = jwtService.validateToken(authorization.substring(7))
+        val token = authorization.substring(7)
+
+        // App token first; null means not an app token, fall through to the user JWT.
```
JanCizmar: These comments are so f***useless, please fo trough the changeset and remove all the inline comments unless you are sure that it's super important to explain.

backend/security/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFilter.kt
```diff
@@ -88,7 +98,21 @@ class AuthenticationFilter(
     val authorization = request.getHeader("Authorization")
     if (authorization != null) {
       if (authorization.startsWith("Bearer ")) {
-        val auth = jwtService.validateToken(authorization.substring(7))
+        val token = authorization.substring(7)
```
JanCizmar, line 97: You have added to much code to this method and class. Try to make the addition to this class minimalistic and move the supporting methods to other class.

backend/security/src/main/kotlin/io/tolgee/security/ProjectContextService.kt
```diff
@@ -68,6 +77,17 @@ class ProjectContextService(
             userId,
           )
 
+          if (authenticationFacade.isAppAuth) {
```
JanCizmar, line 82: Move into separate method with documentation comment instead.

backend/security/src/main/kotlin/io/tolgee/security/ProjectContextService.kt
```diff
@@ -114,6 +134,64 @@ class ProjectContextService(
     populateHolders(project)
   }
 
+  /**
+   * Binds an app authentication to the request's project after checking enablement. Until this runs
+   * the app has no bound project, so permission resolution returns nothing — a route that never
+   * reaches here denies rather than skipping the check.
+   */
+  private fun bindAppToProject(project: ProjectDto) {
```
JanCizmar: Move the apps supporting methods to separate class

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsControllerTest.kt
```diff
@@ -0,0 +1,238 @@
...
+
+  @AfterEach
+  fun cleanup() {
+    testDataService.cleanTestData(testData.root)
+  }
+
+  @Test
+  fun `lists the app's installations with the projects each is enabled for`() {
+    asToken(appLevelToken, get(SELF_INSTALLATIONS)).andIsOk.andAssertThatJson {
+      node("_embedded.installations").isArray.hasSize(1)
```
JanCizmar, line 71: Use our custom node DSL with nesting support instead of this.

JanCizmar:
[claude] Addressed the new round (commit 2a5c798). Done:

- **Mint endpoint moved** off the project controller to `POST /v2/organizations/{organizationId}/apps/{installId}/token` — no projectId, org-wide token, verbose comment gone. `AppTokenModel` moved out of the `project` package. ⚠️ **Frontend/SDK caller must switch to the new path** (and the API schema needs regenerating) — that's a webapp change outside this backend PR.
- **AuthenticationFilter** slimmed: app-token auth extracted to `AppTokenAuthenticator`. **ProjectContextService** slimmed: app project-binding extracted to `AppProjectContextBinder`, and the app-permission-denied block is now a documented method.
- **`native` field removed** from the self-install model. **`expireOthers` is private.** **Inline comments swept** across the changeset — kept only the load-bearing security/ordering ones (constant-time compare, second-precision cutoff, issue+expire atomicity, interceptor ordering, the non-transactional manifest-fetch rationale).

Answers to the questions:

- **`force` on revoke / rotating the only secret** — `force` stamps `App.tokensInvalidBefore=now`, the kill switch that invalidates *every* token already minted from *any* of the app's secrets (the "it leaked, cut it now" path); a plain revoke only stops *new* tokens and leaves outstanding ones alive to expiry. Rotating the only secret is the normal no-downtime rotation: rotate **issues the new secret first** (two live), then grace-expires the old, so the app adopts the new one during the window — that's why rotate has no "last secret" guard (only plain revoke does, since revoking the last one would lock the app out).
- **`SecurityService` app branch (`checkAnyProjectPermission`)** — without it an app token falls through to `getProjectPermissionScopesNoApiKey`, i.e. the **registering user's raw project scopes**, ignoring the install's grantedScopes, the per-project enablement, and the app cap — the app would inherit that person's access. The branch routes app auth through `getCurrentPermittedScopes → getAppPermittedScopes` (install grant ∩ user scopes, gated on enablement). It's a correctness requirement, confirmed by the security review.
- **`@AppAccessNeutral`** — mild push-back: it's load-bearing. The token endpoint authenticates by body `client_secret`, but the SDK sends its *current* app token in the `Authorization` header when renewing; without `@AppAccessNeutral` the `AppAccessInterceptor` would 403 that Bearer app token on a non-app endpoint and break the renew (the test `the token endpoint ignores a valid app token in the Authorization header` pins this). We can drop it if we make the SDK strip the header on token calls — your call.
- **`AppCredentialAuthenticator` first-use branch** — already gone in the current code; nothing to do.

**Deferred (one item), want your steer:** the discovery response's `enabledProjects` nested list → a separate **paginated** endpoint (`GET /v2/apps/self/installations/{installId}/projects`) using a **standard project model**, plus switching `AppSelfInstallationsControllerTest` to the nested node DSL. That's a real API redesign touching the SDK consumer, so I scoped it out of this commit — confirm the endpoint shape and I'll do it as the next change. It's the only nested-collection in the changeset (the flat `scopes: List<String>` is fine).

All app/security/data suites green; ktlint clean. (Separately, I posted the DB-query/Redis caching analysis you asked for in chat — happy to drop it here too.)

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppInstallationsDiscoveryController.kt
JanCizmar: [claude] Fixed — discovery moved off `/v2/public` to `GET /v2/apps/self/installations` behind `@AllowAppLevelAccess`, authenticated by an app-level token; the old controller is deleted.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppInstallationsDiscoveryController.kt
JanCizmar: [claude] Fixed — `client_secret` is only ever exchanged for a token now (app-level or install-context); every other call uses the short-lived token.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Fixed — the whole AppSelfAppSecretsController was removed.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Removed — an app has no reason to list its own secrets.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Removed — secrets are delivered (registration + lifecycle webhook), not self-issued.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfAppSecretsController.kt
JanCizmar: [claude] Agreed — the controller is deleted.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsController.kt
JanCizmar: [claude] Added `@AllowAppLevelAccess`; the interceptor admits an app-level token only on those endpoints.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsController.kt
JanCizmar: [claude] Fixed — it is an app-level endpoint now (`@AllowAppLevelAccess`), not install-context.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Done — omitting `install_id` mints an app-level token for app-level operations.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Done — the grant moved to `AppTokenGrantService`; the controller just delegates.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/apps/AppTokenEndpointController.kt
JanCizmar: [claude] Done — moved to `AppTokenGrantService`.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/project/ProjectAppsController.kt
JanCizmar: [claude] Done — mint moved to `POST /v2/organizations/{organizationId}/apps/{installId}/token`, organization-wide, no projectId.

backend/api/src/main/kotlin/io/tolgee/hateoas/apps/AppSelfInstallationModel.kt
JanCizmar: [claude] Removed the `native` field.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/AppsTestData.kt
JanCizmar: [claude] Removed.

backend/api/src/main/kotlin/io/tolgee/hateoas/apps/AppSelfInstallationModel.kt
JanCizmar: [claude] Done — `GET /v2/apps/self/installations/{installId}/projects`, paginated, returning standard `SimpleProjectModel`s.

backend/api/src/main/kotlin/io/tolgee/hateoas/apps/AppSelfInstallationModelAssembler.kt
JanCizmar: [claude] Done — the nested list is gone; the new projects endpoint returns standard `SimpleProjectModel`s.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppCredentialAuthenticator.kt
JanCizmar: [claude] The first-use branch is already gone in the current code — nothing further needed.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppSecretService.kt
JanCizmar: [claude] Done — `expireOthers` is now private.

backend/security/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFilter.kt
JanCizmar: [claude] Done — swept the inline comments across the changeset, keeping only the load-bearing security/ordering notes.

backend/security/src/main/kotlin/io/tolgee/security/authentication/AuthenticationFilter.kt
JanCizmar: [claude] Done — app-token auth extracted to `AppTokenAuthenticator`; the filter just delegates.

backend/security/src/main/kotlin/io/tolgee/security/ProjectContextService.kt
JanCizmar: [claude] Done — extracted to a `throwAppScopePermissionDenied` method with a doc comment.

backend/security/src/main/kotlin/io/tolgee/security/ProjectContextService.kt
JanCizmar: [claude] Done — moved to `AppProjectContextBinder`.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsControllerTest.kt
JanCizmar: [claude] Done — the test uses the `andAssertThatJson { node(...) }` DSL now, and the nested list is gone (projects are a separate paged endpoint).

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationOwnedAppsController.kt
JanCizmar: [claude] `force` stamps `App.tokensInvalidBefore=now` — the kill switch that invalidates every token already minted from any of the app's secrets (the "it leaked, cut it now" path); a plain revoke only stops new tokens. Rotating the only secret is the normal no-downtime rotation: rotate issues the new secret first, then grace-expires the old, so the app adopts the new one during the window — which is why rotate has no "last secret" guard (only plain revoke does).

backend/data/src/main/kotlin/io/tolgee/service/security/SecurityService.kt
JanCizmar: [claude] Without it an app token falls through to `getProjectPermissionScopesNoApiKey` = the registering user's raw project scopes, ignoring the install's grantedScopes, per-project enablement, and the app cap — the app would inherit that person's access. The branch routes app auth through `getCurrentPermittedScopes → getAppPermittedScopes` (install grant ∩ the acting user's scopes, gated on enablement).

backend/security/src/main/kotlin/io/tolgee/security/authentication/AppAccessNeutral.kt
JanCizmar: [claude] Mild push-back: it is load-bearing. The token endpoint authenticates by body `client_secret`, but the SDK sends its current app token in the `Authorization` header when renewing; without `@AppAccessNeutral` the AppAccessInterceptor 403s that Bearer app token on a non-app endpoint and breaks the renew (pinned by the test `the token endpoint ignores a valid app token in the Authorization header`). We can drop it if we make the SDK strip the header on token calls — your call.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/apps/AppSelfInstallationsControllerTest.kt
JanCizmar: That's cool, but you can also nest the nodes like this node("_embedded.installations[0]") {
 node(...)
}

So rewrite it so it looks better.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationOwnedAppsController.kt
JanCizmar: OK, I understand. Then you should properly document the parameter.

backend/security/src/main/kotlin/io/tolgee/security/authentication/AppTokenAuthenticator.kt
```diff
@@ -0,0 +1,164 @@
...
+  ): AppAuthentication? {
+    if (!tolgeeProperties.apps.enabled) return null
+
+    val claims =
+      try {
+        appTokenService.validateToken(token)
+      } catch (e: AuthExpiredException) {
+        throw e
+      } catch (_: AuthenticationException) {
+        return null
```
JanCizmar: Maybe we can prefix the token with some tolgee conventional prefix, so we throw proper exception here instead of guessing that it's not app auth and continuing with JWT auth. If we had the prefix, we would be sure user is trying to authenticate with app token and we can throw immediately.

backend/security/src/main/kotlin/io/tolgee/security/authentication/AppAccessNeutral.kt
JanCizmar: [claude] Agreed, removed. Dropped the `@AppAccessNeutral` annotation, its usage on the token endpoint, and the interceptor branch. The token endpoint stays public and authenticates by body `client_secret`; a stale app token in the Authorization header now correctly 403s under default-deny, so the SDK must strip it on token calls (SDK-side fix rather than server-side tolerance).

backend/security/src/main/kotlin/io/tolgee/security/authentication/AppTokenAuthenticator.kt
JanCizmar: [claude] Done. App tokens now carry a `tgapp_` prefix (`AppTokenService.TOKEN_PREFIX`). The authenticator recognises an app token by the prefix, so a malformed one fails deterministically here instead of falling through to JWT auth as if it were a user token. `validateToken` strips the prefix before parsing and rejects an unprefixed token outright.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationOwnedAppsController.kt
JanCizmar: [claude] Documented — added a `@Parameter` description on `force` (in addition to the `@Operation` prose) spelling out that it revokes even the last active secret and invalidates every token already minted, and that an ordinary revoke of the only active secret is refused so the app cannot be cut off by mistake.

backend/security/src/main/kotlin/io/tolgee/security/ProjectContextService.kt
```diff
@@ -114,6 +132,51 @@ class ProjectContextService(
...
+    // APP_ACCESS_FORBIDDEN. Otherwise the differing codes let an app enumerate project ids across tenants.
+    val knowsProject = appAuth.tokenProjectId == project.id
+
+    if (appAuth.tokenProjectId != null && !knowsProject) {
+      throw PermissionException(Message.APP_ACCESS_FORBIDDEN)
+    }
+
+    if (!appEnablementService.isEnabledForProject(project.id, appAuth.appInstall.id)) {
+      if (knowsProject) throw PermissionException(Message.APP_NOT_ENABLED_FOR_PROJECT)
+      throw PermissionException(Message.APP_ACCESS_FORBIDDEN)
```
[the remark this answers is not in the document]
JanCizmar: [claude] Fixed. Good catch — this was a real leak: `setup(projectId)` threw `PROJECT_NOT_FOUND` for a missing project *before* `AppProjectContextBinder` could return the opaque `APP_ACCESS_FORBIDDEN`, so a nonexistent id and an existing-but-unreachable id answered with different codes. A missing project now maps to `APP_ACCESS_FORBIDDEN` for app auth, so the two are indistinguishable. The committed `stays opaque on a foreign organization's project` test already pins `999999999 -> APP_ACCESS_FORBIDDEN`.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationOwnedAppsController.kt
```diff
@@ -107,4 +115,67 @@ class OrganizationOwnedAppsController(
...
+    @PathVariable organizationId: Long,
+    @PathVariable appId: Long,
+    @RequestBody(required = false) @Valid body: RollAppSecretRequest?,
+  ): AppSecretRotationModel {
+    val app = appService.getOwned(organizationId, appId)
+    val request = body ?: RollAppSecretRequest()
+    val rotation = appSecretService.rotate(app, request.graceSeconds)
+    return AppSecretRotationModel(
+      secret = appSecretModelAssembler.toModelWithSecret(rotation.issued.secret, rotation.issued.plaintextSecret),
+      previousExpiresAt = rotation.previousExpiresAt?.time,
```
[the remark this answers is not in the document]
JanCizmar: [claude] Clarified in the schema. `expireOthers` only ever shortens (`current == null || current.after(expiry)`) and returns the max outgoing expiry, so `previousExpiresAt` is the deadline of the secret this rotation just superseded — the latest of the outgoing expiries. A still-older secret already inside a shorter grace window from an earlier rotation keeps its own earlier deadline; the `@Schema` now says exactly that. Kept it a single scalar rather than growing the response to per-secret expiries.

backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/apps/AppSecretRotationTest.kt
```diff
@@ -0,0 +1,324 @@
...
+  @Test
+  fun `a rotation keeps the old secret alive through the grace window`() {
+    val rolled = roll(graceSeconds = ONE_DAY)
+
+    rolled
+      .at("/secret/secret")
+      .asText()
+      .assert
+      .startsWith(AppService.APP_CLIENT_SECRET_PREFIX)
+    rolled.at("/previousExpiresAt").asLong().assert.isGreaterThan(0)
```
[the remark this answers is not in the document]
JanCizmar: [claude] Done — the `previousExpiresAt` (and new-secret prefix) checks now go through `andAssertThatJson` (`node("previousExpiresAt").isNumber.isPositive`), keeping the positive-value assertion.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppSecretService.kt
```diff
@@ -32,12 +47,148 @@ class AppSecretService(
...
+    secretId: Long,
+    previousLastUsedAt: Date?,
+  ) {
+    runSentryCatching {
+      val now = currentDateProvider.date
+      val throttle = LAST_USED_THROTTLE_MS
+      if (previousLastUsedAt != null && now.time - previousLastUsedAt.time < throttle) {
+        return@runSentryCatching
+      }
+      appSecretRepository.updateLastUsedById(secretId, now)
```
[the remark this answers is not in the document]
JanCizmar: [claude] Leaving as-is, deliberately. The existing `ApiKeyService.updateLastUsedAsync` writes `lastUsedAt` unconditionally on *every* use with no throttle at all; this app-secret path already throttles to ~1 write / 60s. The race you describe is real but self-limiting: a burst right after the window yields a bounded handful of redundant writes, last-writer-wins on a monotonic timestamp — no unbounded growth and no correctness issue on a non-critical field. An atomic conditional UPDATE would exceed the platform's own convention here for little benefit. Happy to add it if you'd rather match a stricter bar — leaving this open for your call.

PR #3870, opened 2026-08-24
docs: rewrite the Docker Hub overview and keep its source in the repo

dkrizan:
The [tolgee/tolgee Docker Hub page](https://hub.docker.com/r/tolgee/tolgee) has not had a content update since **October 2021**. It has taken 1.66M pulls since. This moves its source into the repo and rewrites it.

## What is wrong with the current page

| | |
| --- | --- |
| **Both doc links 404** | They point at the old `server_and_web_app` docs tree. These are the page's only two routes into the docs. |
| **You cannot log in by following it** | It never mentions that the admin password is generated on first boot into `/data/initial.pwd`. |
| **Silent about the bundled PostgreSQL** | The image is built on `postgres:13-alpine` and autostarts a database. Nothing tells a reader to disable it and attach a real one before production. |
| **No reference material** | No tags, platforms, ports, volume, healthcheck or environment variables — the things a Docker Hub reader actually comes for. |
| **Feature list stops at 2021** | No translation memory, glossaries, LLM machine translation, branching, SSO, batch operations, content delivery, webhooks, CLI, SDK range or MCP server. |
| **Screenshot and GIF are from 2021** | 2021 `user-images.githubusercontent.com` uploads of a UI that no longer exists. |
| **Short description is empty** | The repository renders blank in Docker Hub search. |

The root cause is that the description is hand-maintained in the Docker Hub UI, with no copy in the repo — so nobody reviewing a change to `docker/` ever saw it drift.

## What this PR does

- **`docker/DOCKERHUB.md`** — the new page, and from here on the source of truth for it. Quick start (including reading `initial.pwd`), production compose with an external database, tag and image reference, the common `TOLGEE_*` variables, the current feature set, licensing.
- **`.github/workflows/dockerhub-description.yml`** — pushes that file and the short description to Docker Hub. **Manual dispatch only**, no push trigger.

Every claim in the file is taken from `docker/app/Dockerfile`, `gradle/docker.gradle`, the Docker Hub tags API or the current self-hosting docs. All 21 links in it were checked and resolve.

### Why dispatch-only

Updating a Docker Hub description requires the credential to have **Admin** on the repository — push access, which is all the release workflows need, is not enough, and we do not know yet whether `DOCKERHUB_PASSWORD` clears that bar. Dispatch-only keeps that uncertainty off `main`: nothing runs on merge, and the first run is a deliberate click by someone who can read the result. The page changes maybe once a year, so a push trigger buys little anyway.

If the run fails on permissions, the fix is to reissue `DOCKERHUB_PASSWORD` as a PAT with `read/write/delete` scope on an account with Admin on `tolgee/tolgee`. Failing that, the file can simply be pasted into the Docker Hub UI by hand.

## Still needs someone with Docker Hub access

**Categories are unset**, and the action cannot set them — so the repository currently appears in no browse listing. That is a manual change in the Docker Hub UI.

## Follow-up, not included here

The page needs one current screenshot of the translation editor and one in-context GIF, committed to the repo rather than relying on the 2021 uploads. I left the old ones out entirely rather than ship a fresh page with a four-year-old UI in it.

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive Docker Hub documentation covering supported architectures, runtime requirements, quick-start instructions, configuration options, image tags, licensing, and deployment guidance.
  - Documented evaluation and production deployment options, including PostgreSQL requirements and administrator password retrieval.

- **Chores**
  - Added a manually triggered process to synchronize the Docker Hub description with the latest project documentation.

### DIFF
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -1,0 +1,23 @@
+name: Docker Hub description
+
+# Run manually from the Actions tab after changing docker/DOCKERHUB.md.
+# Updating a description requires the Docker Hub credential to have Admin on
+# the repository — push access, which is all the release workflows need, is not
+# enough.
+on:
+  workflow_dispatch:
+
+jobs:
+  sync:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Push description to Docker Hub
+        uses: peter-evans/dockerhub-description@v4
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+          repository: tolgee/tolgee
+          readme-filepath: ./docker/DOCKERHUB.md
+          short-description: "Open-source localization platform — self-host the whole Tolgee server in one container."
--- a/docker/DOCKERHUB.md
+++ b/docker/DOCKERHUB.md
@@ -1,0 +1,157 @@
+[![Docker Image Version](https://img.shields.io/docker/v/tolgee/tolgee/latest?label=docker&logo=docker&logoColor=fff&color=2496ED)](https://hub.docker.com/r/tolgee/tolgee/tags)
+[![GitHub Release](https://img.shields.io/github/v/release/tolgee/tolgee-platform?label=release)](https://github.com/tolgee/tolgee-platform/releases/latest)
+[![License](https://img.shields.io/badge/license-Apache%202%20%2F%20Tolgee%20EL-blue)](https://github.com/tolgee/tolgee-platform/blob/main/LICENSE)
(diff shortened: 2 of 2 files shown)


dkrizan:
Closing this — we will update the Docker Hub description manually once someone has Admin credentials on the repository. The findings and the rewritten page are kept outside the repo for now.

For whoever picks this up, the substantive problems with the current page are:

- Both documentation links 404 (they point at the retired `server_and_web_app` docs tree)
- The quick start never mentions that the initial admin password is generated into `/data/initial.pwd`, so it cannot actually get you logged in
- Nothing says a PostgreSQL runs inside the image, so nobody learns to disable it before production
- No tags, ports, volume, healthcheck or environment variable reference
- The feature list and both images date from October 2021
- The short description is empty, so the repository renders blank in Docker Hub search, and categories are unset

PR #3871, opened 2026-08-24
feat: add language-scoped filter for translations never included in a task

### LINKED ISSUE #3848: Add a language-scoped filter for translations never included in a task
## Problem

When preparing translation batches for an external agency, users need to avoid sending the same translation work more than once. A current workaround is to tag each batch with the task date, send only translations with that tag, and later search for items without a tag. This works as lightweight task tracking, but it is manual and easy to misuse.

Customer feedback identified a need for a filter answering: “Has this been in a task before?”

The filter needs to be scoped to a specific target language: a key may have been included in a task for one language but not for another.

## Proposed solution

Add a filter in the translation/task preparation workflow that can identify whether the selected language’s translation has ever been included in a task, for example:

- `Has been in a task: No` / `Never included in a task`
- optionally the inverse: `Has been in a task: Yes`

The filter should use task history rather than tags, and should continue to work when existing tasks or date tags are used.

## Acceptance criteria

- A user can filter by a selected target language.
- The filter can return only translations that have never been included in any task for that language.
- Task history is used as the source of truth; the result is not dependent on manually maintained tags.
- A translation included in a task for language A is still considered “never included” for language B.
- The result can be used when creating/preparing a new batch so duplicate work is not sent accidentally.
- Existing task and tag workflows remain backward compatible.

## Context

This is a high-priority workflow improvement for teams that send recurring translation batches to external agencies and need a reproducible way to prevent duplicate requests.

bdshadow:
Closes #3848

Adds a language-scoped filter answering "has this translation been in a task before?", so teams sending recurring batches to an external agency can avoid sending the same work twice without relying on manually maintained date tags.

## What's in it

**Translations view** — a new **Tasks** subfilter with `Has been in a task` / `Never in a task`, plus the usual language-scope submenu (all languages / all but base / one specific language). Backed by two new query params, `filterHasTaskInLang` and `filterHasNoTaskInLang`.

**Task creation & Order translations dialogs** — the same two options, scoped per target language server-side via `filterNeverInTask` / `filterHasBeenInTask` on `TranslationScopeFilters`. `select-all` narrows the key set with OR-across-languages semantics, then `calculate-scope` and `create-multiple-tasks` apply the per-language filter, so a key tasked in `de` but not `fr` lands only in the `fr` task.

## Notes for the reviewer

- **Task membership, not an immutable audit log.** The filter reads `task_key`. Rows survive CANCELED/FINISHED tasks and both task types, but are hard-deleted when a key is removed from a task (`TaskService.updateTaskKeys`) or a task is deleted (`TaskService.deleteAll`). So deleting a finished agency task makes its keys eligible again. The parameter docs say this explicitly; flagging it here because the labels ("Never in a task") promise slightly more than the data model delivers. Happy to rename the labels or back this with the activity log if that's the call.
- **AND, not OR.** These filters are AND-ed with the rest of the query rather than joining the OR-ed `translationConditions` bucket, so combining with a state filter narrows rather than re-admits already-tasked keys. Consistent with the existing `filterTaskNumber`.
- **Gated on TASKS / ORDER_TRANSLATION**, mirroring how QA check filters are gated — silently ignored without the feature.
- **Second commit fixes a pre-existing bug** beyond the ticket: `updateSelectedLanguages` cleared `filterTranslationLanguage` when the scoped language was *still* selected rather than when it was gone, and replaced the whole filter object instead of merging — so adding any language to the view wiped every active filter. `filterSuggestionLanguage` had no cleanup at all.
- **Side effect worth knowing:** `TaskCreateForm` previously passed `selectedLanguages={[]}`, which made the suggestion filters in the task/order dialogs settable but inert. They now work.
- Needs tolgee/billing#299 (schema regen for the order-translation endpoint; same branch name).

## Testing

- `TranslationsControllerTaskHistoryFilterTest` — 12 tests covering per-language scoping, canceled/finished/review tasks, multi-language OR, AND with other filters, `select-all` parity, unresolvable tags, and the feature-disabled path.
- `TaskScopeTaskHistoryFilterTest` — per-language scope calculation and a multi-language batch that must not re-task a key in the language that already had it.
- Cypress: `translationFilters.tasks.cy.ts` (both directions) and a regression spec for the language-scope clearing fix.

## Summary by CodeRabbit

* **New Features**
  * Added translation filters for keys that have, or have never had, tasks.
  * Added task-language filtering to refine results by target language.
  * Added task-history filters to task creation and ordering workflows.
  * Prevented keys from being re-added for languages where they were already assigned.
* **Bug Fixes**
  * Cleared language-specific filters when languages are deselected.
* **Tests**
  * Added coverage across languages, task states, task types, and task creation workflows.

### DIFF
--- a/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TaskTestData.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TaskTestData.kt
@@ -272,9 +272,12 @@ class TaskTestData : BaseTestData("tasksTestUser", "Project with tasks") {
     state: TaskState,
     type: TaskType,
     number: Long,
-  ) {
+    taskLanguage: Language = englishLanguage,
+    taskKeys: Collection<KeyBuilder> = keysInTask,
+  ): TaskBuilder {
+    lateinit var createdTask: TaskBuilder
     projectBuilder.apply {
-      blockedTask =
+      createdTask =
         addTask {
           this.number = number
           this.name = name
@@ -285,20 +288,54 @@ class TaskTestData : BaseTestData("tasksTestUser", "Project with tasks") {
               user,
             )
           project = projectBuilder.self
-          language = englishLanguage
+          language = taskLanguage
           author = projectUser.self
           this.state = state
         }
 
-      keysInTask.forEach {
+      taskKeys.forEach {
         addTaskKey {
-          task = blockedTask.self
+          task = createdTask.self
           key = it.self
           done = true
           author = user
         }
       }
     }
+    blockedTask = createdTask
+    return createdTask
+  }
+
+  fun addKeyWithOwnTask(
+    keyName: String,
+    number: Long,
+    taskLanguage: Language = englishLanguage,
+    type: TaskType = TaskType.TRANSLATE,
+    state: TaskState = TaskState.NEW,
+  ): KeyBuilder {
+    lateinit var createdKey: KeyBuilder
+    projectBuilder.apply {
+      createdKey =
+        addKey(null, keyName) {
+          addTranslation("en", "Translation of $keyName")
+          addTranslation("cs", "Překlad $keyName")
+        }
+      val ownTask =
+        addTask {
+          this.number = number
+          this.name = "Task of $keyName"
+          this.type = type
+          this.state = state
+          project = projectBuilder.self
+          language = taskLanguage
+          author = projectUser.self
+        }
+      addTaskKey {
+        task = ownTask.self
+        key = createdKey.self
+      }
+    }
+    return createdKey
   }
 
   fun createManyOutOfTaskKeys(): List<KeyBuilder> {
--- a/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TaskTestData.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TaskTestData.kt
@@ -272,12 +272,9 @@ class TaskTestData : BaseTestData("tasksTestUser", "Project with tasks") {
     state: TaskState,
     type: TaskType,
     number: Long,
-    taskLanguage: Language = englishLanguage,
-    taskKeys: Collection<KeyBuilder> = keysInTask,
-  ): TaskBuilder {
-    lateinit var createdTask: TaskBuilder
+  ) {
     projectBuilder.apply {
-      createdTask =
+      blockedTask =
         addTask {
           this.number = number
           this.name = name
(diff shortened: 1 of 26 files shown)


bdshadow:
@coderabbitai please review

JanCizmar:
I would be super grateful for video preview that I can send the TAB member who proposed this improvement.

backend/data/src/main/kotlin/io/tolgee/development/testDataBuilder/data/TaskTestData.kt
```diff
@@ -272,9 +272,12 @@ class TaskTestData : BaseTestData("tasksTestUser", "Project with tasks") {
     state: TaskState,
     type: TaskType,
     number: Long,
-  ) {
+    taskLanguage: Language = englishLanguage,
```
dkrizan: `taskLanguage` and `taskKeys` have zero callers as well as return (`return createdTask`)
I would just use the default values straight in the fun body

dkrizan (dismissed):
~~Just one nitpick, otherwise approved.~~

dkrizan (commented):
With "all but base" selected, a key that was in a German task but not in a French one
matches both "Never in a task" and "Has been in a task". Both options mean "in at least
one language", so they overlap. With a single language selected it works fine.

I think this mechanism already exists in other filters (Translations, Suggestions, QA
checks), so it is not new. But the issue asks for it per translation, and one row covers
several languages at once, so maybe it matters more here. One option would be to drop
"all but base" and "all". 

What do you think?

bdshadow:
@dkrizan i agree, nice catch, thank you. Fixed

bdshadow:
https://github.com/user-attachments/assets/dbeba5b5-7c96-415b-aeb9-389d17a2ec48

backend/data/src/main/kotlin/io/tolgee/dtos/request/translation/TranslationFilters.kt
```diff
@@ -208,6 +208,25 @@ $PATTERN_GRAMMAR_DOC""",
...
+        "so this reflects current task membership rather than an immutable audit log.",
+  )
+  var filterHasTaskInLang: List<String>? = null
+
+  @field:Parameter(
+    description =
+      "Select only keys not currently linked to a task in any of the provided languages, the exact " +
+        "complement of filterHasTaskInLang. Tasks in all states (including CANCELED and FINISHED) " +
+        "and of both types (TRANSLATE, REVIEW) count. The link is dropped when a key is removed " +
+        "from a task or when the branch holding it is deleted, so a key can reappear here after " +
```
dkrizan, line 225: description is probably unnecessarily long (I would omit the last sentence), but I leave it to you, approved anyway

dkrizan (approved):

Issue #2737, opened 2024-12-02
Remove organization dropdown, add project dropdown


JanCizmar:
User's usually don't want to switch organization, but they usually want to switch projects.

![image](https://github.com/user-attachments/assets/ce89094a-edee-4a08-852f-1734dbbff90f)

- [ ] Remove organization dropdown
- [ ] Add project dropdown 

@ZuzanaOdstrcilova do we need design proposal?

JanCizmar:
@stepan662 thoughts?

stepan662:
Make sense, so @StaNov is doing UI also?

It would be useful to still be able to click on project (when you just want to go to the dashbord) and have the switch separately (only the arrow clickable). Could be some general component.

ZuzanaOdstrcilova:
We have design here: 
- https://www.figma.com/design/wVmz098ZrSxswOI0nbIjQ2/Design-system-Tolgee?node-id=1487-49289&t=UTwrMRyFHZFlenH3-1

Aditya-NV-06:
Can I work on this

PR #3873, opened 2026-08-25
feat: add project switcher dropdown to the project breadcrumb

### LINKED ISSUE #2150: Add dropdown to projects
- [ ] Add the same dropdown as for organizations to projects 

image

- [ ] remove + button in Organizations dropdown

dkrizan:
Closes #2150

## What

- Adds a project switcher next to the project name in the breadcrumb, so users can jump between projects of the current organization without going back to the list.
- Removes the "+" (create organization) button from the breadcrumb organization dropdown, per @ZuzanaOdstrcilova's note that creating an organization belongs in settings rather than in the quick switcher.

## How

The switcher reuses the existing generic `SwitchPopover` (search, paging, keyboard nav) that already backs the organization dropdown, and the existing `ProjectSearchSelectItem` for rows. No backend change — `/v2/organizations/{slug}/projects` already supports `search` and paging.

The project name stays a link to the project dashboard; the dropdown is a separate arrow button beside it. Projects are scoped to the *viewed project's* organization (`project.organizationOwner`), not the globally preferred organization, so opening a shared link to a project in another org lists the right projects even before the preferred-organization state catches up.

Creating an organization is still available from the user avatar menu → Switch organization → "+", which is now the single entry point.

## Deliberately out of scope

- The Figma visual polish listed in the 2024 comment (new colors, 8px corners, elevation) — this PR only adds the control.
- No "+ new project" in the project dropdown, per the same comment.
- #2737 ("remove organization dropdown") is a separate decision and is untouched here.

## Testing

- New `e2e/cypress/e2e/projects/projectSwitching.cy.ts`: switching to another project of the organization, and the project name still linking to the dashboard.
- `organizationsSwitching.cy.ts`: asserts the breadcrumb dropdown no longer offers "+".
- `organizationsList.cy.ts` and `administration/base.cy.ts` repointed at the user-menu switcher for their organization-creation assertions.

All four specs pass locally (21/21), plus webapp/e2e eslint + tsc.

## Summary by CodeRabbit

- **New Features**
  - Added a project switcher to project navigation with search, pagination, and direct navigation to selected projects.
  - Project names now open the project dashboard.
  - Organization switching is available through the global user menu.

- **Bug Fixes**
  - Organization creation is no longer offered from the organization breadcrumb switcher.

- **Tests**
  - Added end-to-end coverage for project switching and organization switcher behavior.

### DIFF
--- a/e2e/cypress/e2e/projects/projectSwitching.cy.ts
+++ b/e2e/cypress/e2e/projects/projectSwitching.cy.ts
@@ -1,0 +1,36 @@
+import { login, setBypassSeatCountCheck } from '../../common/apiCalls/common';
+import { projectListData } from '../../common/apiCalls/testData/testData';
+import { waitForGlobalLoading } from '../../common/loading';
+import { enterProject } from '../../common/projects';
+import { gcy } from '../../common/shared';
+
+describe('Project switching', () => {
+  beforeEach(() => {
+    setBypassSeatCountCheck(true);
+    projectListData.clean();
+    projectListData.generate();
+    login('projectListDashboardUser', 'admin');
+    enterProject('test_project');
+  });
+
+  afterEach(() => {
+    projectListData.clean();
+    setBypassSeatCountCheck(false);
+  });
+
+  it('switches to another project of the organization', () => {
+    gcy('project-switch').click();
+    gcy('switch-popover-item').contains('Project 2').click();
+    waitForGlobalLoading();
+
+    gcy('navigation-item').contains('Project 2').should('be.visible');
+    cy.url().should('not.contain', '/translations');
+  });
+
+  it('keeps the project name linking to the dashboard', () => {
+    gcy('navigation-item').contains('test_project').click();
+    waitForGlobalLoading();
+
+    gcy('project-dashboard-language-count').should('be.visible');
+  });
+});
--- a/webapp/src/component/organizationSwitch/OrganizationSwitch.tsx
+++ b/webapp/src/component/organizationSwitch/OrganizationSwitch.tsx
@@ -62,11 +62,6 @@ export const OrganizationSwitch: React.FC<React.PropsWithChildren<Props>> = ({
     onSelect?.(organization);
   };
 
-  const handleCreateNewOrg = () => {
-    handleClose();
-    history.push(LINKS.ORGANIZATIONS_ADD.build());
-  };
-
   const isCommunitySurface = selectedSurface === 'community';
 
   const switchLabel = isCommunitySurface ? (
@@ -89,7 +84,6 @@ export const OrganizationSwitch: React.FC<React.PropsWithChildren<Props>> = ({
         selected={preferredOrganization}
         onSelect={handleSelectOrganization}
         anchorEl={anchorEl.current!}
-        onAddNew={handleCreateNewOrg}
         communitySelected={isCommunitySurface}
         onCommunityNavigate={
           hasCommunityContributions
--- a/webapp/src/component/projectSwitch/ProjectSwitch.tsx
+++ b/webapp/src/component/projectSwitch/ProjectSwitch.tsx
@@ -1,0 +1,117 @@
+import { useCallback, useRef, useState } from 'react';
+import { IconButton } from '@mui/material';
+import { useTranslate } from '@tolgee/react';
+import { useHistory } from 'react-router-dom';
+
+import { ArrowDropDown } from 'tg.component/CustomIcons';
+import { ProjectSearchSelectItem } from 'tg.component/projectSearchSelect/ProjectSearchSelectItem';
+import { SwitchPopover } from 'tg.component/SwitchPopover/SwitchPopover';
+import { LINKS, PARAMS } from 'tg.constants/links';
+import { components } from 'tg.service/apiSchema.generated';
+import { useApiInfiniteQuery } from 'tg.service/http/useQueryApi';
+
+type ProjectModel = components['schemas']['ProjectModel'];
+
+type Props = {
+  project: ProjectModel;
+};
+
+export const ProjectSwitch: React.FC<Props> = ({ project }) => {
+  const anchorEl = useRef<HTMLButtonElement>(null);
+  const [isOpen, setIsOpen] = useState(false);
+  const [search, setSearch] = useState('');
+  const { t } = useTranslate();
+  const history = useHistory();
+
+  const slug = project.organizationOwner?.slug || '';
+
+  const query = {
+    search: search || undefined,
+    size: 20,
+    sort: ['name'],
+  };
+
+  const projectsLoadable = useApiInfiniteQuery({
+    url: '/v2/organizations/{slug}/projects',
+    method: 'get',
+    path: { slug },
+    query,
+    options: {
+      keepPreviousData: true,
+      enabled: isOpen && Boolean(slug),
+      getNextPageParam: (lastPage) => {
+        if (
+          lastPage.page &&
+          lastPage.page.number! < lastPage.page.totalPages! - 1
+        ) {
+          return {
+            path: { slug },
+            query: {
+              ...query,
+              page: lastPage.page!.number! + 1,
+            },
+          };
+        } else {
+          return null;
+        }
+      },
+    },
+  });
+
+  const items = (projectsLoadable.data?.pages
+    .flatMap((page) => page._embedded?.projects)
+    .filter(Boolean) ?? []) as ProjectModel[];
+
+  const totalElements =
+    projectsLoadable.data?.pages[0]?.page?.totalElements ?? 0;
+
+  const handleSearchChange = useCallback((value: string) => {
+    setSearch(value);
+  }, []);
+
+  const handleSelect = (selected: ProjectModel) => {
+    setIsOpen(false);
+    history.push(
+      LINKS.PROJECT_DASHBOARD.build({ [PARAMS.PROJECT_ID]: selected.id })
+    );
+  };
+
+  return (
+    <>
+      {/* Anchored on the preceding breadcrumb link (avatar + project name) rather
+          than on this button, so the popover lines up with the project the same way
+          the organization one lines up with its label. Rendering this switch anywhere
+          other than directly after that link silently misplaces the popover. */}
+      <IconButton
+        ref={anchorEl}
+        sx={{ p: 0, color: 'inherit' }}
+        onClick={() => setIsOpen(true)}
+        data-cy="project-switch"
+        aria-label={t('projects_switch_label', 'Switch project')}
+        size="small"
+      >
+        <ArrowDropDown width={20} height={20} />
+      </IconButton>
+
+      <SwitchPopover
+        open={isOpen}
+        onClose={() => setIsOpen(false)}
+        onSelect={handleSelect}
+        anchorEl={
+          (anchorEl.current?.previousElementSibling as HTMLElement) ??
+          anchorEl.current!
+        }
+        selectedId={project.id}
+        items={items}
+        isLoading={projectsLoadable.isFetching}
+        hasNextPage={projectsLoadable.hasNextPage ?? false}
+        fetchNextPage={() => projectsLoadable.fetchNextPage()}
+        totalElements={totalElements}
+        renderItem={(item) => <ProjectSearchSelectItem data={item} />}
+        searchPlaceholder={t('projects_search_placeholder', 'Search projects')}
+        headingText={t('projects_title', 'Projects')}
+        onSearchChange={handleSearchChange}
+      />
+    </>
+  );
+};
--- a/webapp/src/views/projects/BaseProjectView.tsx
+++ b/webapp/src/views/projects/BaseProjectView.tsx
@@ -6,6 +6,7 @@ import { BaseView, BaseViewProps } from 'tg.component/layout/BaseView';
 import { NavigationItem } from 'tg.component/navigation/Navigation';
 import { SmallProjectAvatar } from 'tg.component/navigation/SmallProjectAvatar';
 import { OrganizationSwitch } from 'tg.component/organizationSwitch/OrganizationSwitch';
+import { ProjectSwitch } from 'tg.component/projectSwitch/ProjectSwitch';
 import { LINKS, PARAMS } from 'tg.constants/links';
 import { useProject } from 'tg.hooks/useProject';
 import { BatchOperationsSummary } from './translations/BatchOperations/OperationsSummary/OperationsSummary';
@@ -44,7 +45,10 @@ export const BaseProjectView: React.FC<React.PropsWithChildren<Props>> = ({
         [PARAMS.PROJECT_ID]: project.id,
       }),
       <SmallProjectAvatar key="avatar" project={project} />,
-      branching ? <GlobalBranchSelector key="branch" /> : undefined,
+      <React.Fragment key="suffix">
+        <ProjectSwitch project={project} />
+        {branching && <GlobalBranchSelector />}
+      </React.Fragment>,
     ]);
   }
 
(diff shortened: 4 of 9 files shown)


dkrizan:
Translation key `projects_switch_label` ("Switch project") has been created in Tolgee, tagged `draft: project-switch-dropdown`.

No screenshot was attached to it: the key is an `aria-label` on an icon-only button, so it renders no visible text and `getVisibleKeys()` returns no bounding box for it. Worth attaching manually if we want the in-context view to cover it.

dkrizan:
### Breadcrumb

breadcrumb

### Dropdown open

dropdown open

### With branching enabled

branching breadcrumb

branching dropdown open

e2e/cypress/e2e/projects/projectSwitching.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+import { projectListData } from '../../common/apiCalls/testData/testData';
+import { waitForGlobalLoading } from '../../common/loading';
+import { enterProject } from '../../common/projects';
+import { gcy } from '../../common/shared';
+
+describe('Project switching', () => {
+  beforeEach(() => {
+    setBypassSeatCountCheck(true);
+    projectListData.clean();
+    projectListData.generate();
```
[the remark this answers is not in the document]
dkrizan: Not applicable here. `generateStandard()` requests `/generate-standard`, which only exists on controllers extending `AbstractE2eDataController`. `ProjectListDashboardE2eDataController` is a plain class exposing only `/generate` and `/clean`, so this change would 404. The guideline carves this out: legacy controllers without the base class keep `.generate()`.

e2e/cypress/e2e/projects/projectSwitching.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  afterEach(() => {
+    projectListData.clean();
+    setBypassSeatCountCheck(false);
+  });
+
+  it('switches to another project of the organization', () => {
+    gcy('project-switch').click();
+    gcy('switch-popover-item').contains('Project 2').click();
```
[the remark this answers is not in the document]
dkrizan: Keeping `contains()` here. The guideline marks text queries on *test-owned* data as discouraged rather than forbidden, and `gcy("switch-popover-item").contains(name)` is exactly what `organizationsSwitching.cy.ts` and the shared `switchToOrganization()` helper already do against this same component — adding a `data-cy-item` to the shared `SwitchPopover` for a test-only affordance would key on the same string while diverging from how the org switcher is tested. The `navigation-item` assertions (26, 31) are deliberately about the visible project name, which is what the user actually sees after switching.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,110 @@
...
+        anchorEl={anchorEl.current!}
+        selectedId={project.id}
+        items={items}
+        isLoading={projectsLoadable.isFetching}
+        hasNextPage={projectsLoadable.hasNextPage ?? false}
+        fetchNextPage={() => projectsLoadable.fetchNextPage()}
+        totalElements={totalElements}
+        renderItem={(item) => <ProjectSearchSelectItem data={item} />}
+        searchPlaceholder={t('projects_search_placeholder')}
+        headingText={t('projects_title')}
```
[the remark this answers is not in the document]
dkrizan: Applied — both now pass defaults (`Search projects` / `Projects`).

JanCizmar:
I would propose also removing the dropbar from organization level. We don't want people switch / create organizations often. As we have discussed this before.

Anty0:
I'd pull in @ZuzanaOdstrcilova on this decision so we don't break any use cases by removing that. While uncommon, I don't see an alternative for people who are already used to switching organizations that way.

dkrizan:
I didn’t remove the organization switch intentionally, as the discussion didn’t make it clear that we genuinely wanted to do so. Perhaps I’ll address this as a follow-up and maintain the scope of this PR as it is.

Worth asking @ZuzanaOdstrcilova for her opinion.

dkrizan:
Popover alignment fixed — it was anchored on the arrow button, so it centred on the arrow instead of the project. It now anchors on the project label, the same way the organization switcher anchors on its own label.

Measured on the same page: organization popover centre 132 vs its anchor 131; project popover centre 305 vs its anchor 304.

Also rebased onto current `main`.

### With branching enabled





The branch selector sits in the same breadcrumb wrapper as the arrow, so anchoring on the wrapper would have shifted the popover right by ~80px only when branching is on. Anchoring on the label keeps it in the same place either way.

ZuzanaOdstrcilova:
I wouldn’t remove the organization switcher either, even though we discussed it in the past. 
It’s now important for people who want to switch between their organization’s projects and community projects.

If we don’t want people to create unnecessary organizations, I would hide the “+ New organization” option. 
In that case, however, it would be important to include a “+” option in the main organization switcher.

dkrizan:
> I wouldn’t remove the organization switcher either, even though we discussed it in the past. It’s now important for people who want to switch between their organization’s projects and community projects.
> 
> If we don’t want people to create unnecessary organizations, I would hide the “+ New organization” option. In that case, however, it would be important to include a “+” option in the main organization switcher.

The “New organization” button was removed in this PR, while the same button remains in the main organization switcher. So I guess PR is ready for review.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,117 @@
...
+      >
+        <ArrowDropDown width={20} height={20} />
+      </IconButton>
+
+      <SwitchPopover
+        open={isOpen}
+        onClose={() => setIsOpen(false)}
+        onSelect={handleSelect}
+        anchorEl={
+          (anchorEl.current?.previousElementSibling as HTMLElement) ??
```
Anty0, line 101: Anchoring on `previousElementSibling` couples this component to its own position as the first child of the suffix fragment in `BaseProjectView`.

`OrganizationSwitch` avoids this by rendering label + arrow inside one element it owns and anchoring on that. That also lets the comment above the button go away.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,117 @@
...
+};
+
+export const ProjectSwitch: React.FC<Props> = ({ project }) => {
+  const anchorEl = useRef<HTMLButtonElement>(null);
+  const [isOpen, setIsOpen] = useState(false);
+  const [search, setSearch] = useState('');
+  const { t } = useTranslate();
+  const history = useHistory();
+
+  const slug = project.organizationOwner?.slug || '';
```
Anty0: `ProjectModel.organizationOwner` is nullable (legacy user-owned projects, see `Project.kt`). For such a project the arrow still renders, the query never fires (`enabled: isOpen && Boolean(slug)`), and the popover opens to "Nothing found" with no loading state. Consider rendering `ProjectSwitch` only when `project.organizationOwner` is set.

Related: a non-member viewing a public project also gets "Nothing found" because the org projects endpoint filters by the caller's permissions. Probably acceptable, but worth deciding consciously.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,117 @@
...
+  };
+
+  const projectsLoadable = useApiInfiniteQuery({
+    url: '/v2/organizations/{slug}/projects',
+    method: 'get',
+    path: { slug },
+    query,
+    options: {
+      keepPreviousData: true,
+      enabled: isOpen && Boolean(slug),
```
Anty0: Gating the query on `isOpen` has a side effect in `SwitchPopover`: on the very first open `totalElements` is still 0, so `displaySearch` stays `undefined`, the search input mounts with `display: none`, and its `autoFocus` is a no-op. When the data arrives and the input becomes visible nothing refocuses it (its `key` is `Number(open)`, which hasn't changed), so typing/arrow keys go nowhere until the user clicks the field. Second and later opens are fine because `displaySearch` has latched.

`OrganizationPopover` doesn't hit this because its query runs at mount. Worth a quick manual check with 11+ projects in the org (search only shows above the threshold of 10).

e2e/cypress/e2e/projects/projectSwitching.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+    waitForGlobalLoading();
+
+    gcy('navigation-item').contains('Project 2').should('be.visible');
+    cy.url().should('not.contain', '/translations');
```
Anty0: This passes for any URL that isn't the translations view, including the projects list, a 404, or an error page. `handleSelect` builds `LINKS.PROJECT_DASHBOARD` explicitly, so that's the thing to pin. The sibling test already uses `gcy('project-dashboard-language-count').should('be.visible')` as the dashboard signal; suggest the same here.

e2e/cypress/e2e/projects/projectSwitching.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+    login('projectListDashboardUser', 'admin');
+    enterProject('test_project');
+  });
+
+  afterEach(() => {
+    projectListData.clean();
+    setBypassSeatCountCheck(false);
+  });
+
+  it('switches to another project of the organization', () => {
```
Anty0, line 21: Tests don't exercise search and pagination. Since those are quite rarely in use, breakage there could go unnoticed for long - I've seen that happen before.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { components } from 'tg.service/apiSchema.generated';
+import { useApiInfiniteQuery } from 'tg.service/http/useQueryApi';
+
+type ProjectModel = components['schemas']['ProjectModel'];
+
+type Props = {
+  project: ProjectModel;
+};
+
+export const ProjectSwitch: React.FC<Props> = ({ project }) => {
```
Anty0: Optional: this file merges the trigger button with the infinite-query plumbing, whereas the feature it's modelled on splits them (`OrganizationSwitch.tsx` = trigger/anchor/open state, `OrganizationPopover.tsx` = query + `SwitchPopover` wiring; same split in `ProjectSearchSelect` / `ProjectSearchSelectPopover`). The pagination block here is a near copy of `OrganizationPopover`'s. Following the same split would make the two switchers read the same way.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,117 @@
...
+    .filter(Boolean) ?? []) as ProjectModel[];
+
+  const totalElements =
+    projectsLoadable.data?.pages[0]?.page?.totalElements ?? 0;
+
+  const handleSearchChange = useCallback((value: string) => {
+    setSearch(value);
+  }, []);
+
+  const handleSelect = (selected: ProjectModel) => {
```
Anty0: Optional: `history.push` lives inside the component here, while `OrganizationSwitch` right beside it in `BaseProjectView` takes `onSelect` and lets the view own the navigation. Minor with one consumer, but it's the opposite split from the neighbouring control.

Also: picking the already-selected project from `/translations` or `/settings` navigates to the dashboard rather than being a no-op, so the user loses their place inside the project. For the org switch that doesn't matter since the org list is where you'd land anyway.

webapp/src/views/projects/BaseProjectView.tsx
```diff
@@ -44,7 +45,10 @@ export const BaseProjectView: React.FC<React.PropsWithChildren<Props>> = ({
         [PARAMS.PROJECT_ID]: project.id,
       }),
       <SmallProjectAvatar key="avatar" project={project} />,
-      branching ? <GlobalBranchSelector key="branch" /> : undefined,
+      <React.Fragment key="suffix">
+        <ProjectSwitch project={project} />
```
Anty0, line 49: Note: the breadcrumb now has two interaction models side by side. Clicking the org label opens its switcher; clicking the project label navigates and the dropdown is a separate arrow. #2150 asks for "the same dropdown as for organizations". Users trained by the org control may click the project name expecting the switcher. The PR states the choice but doesn't argue it.

webapp/src/component/organizationSwitch/OrganizationSwitch.tsx
```diff
@@ -62,11 +62,6 @@ export const OrganizationSwitch: React.FC<React.PropsWithChildren<Props>> = ({
     onSelect?.(organization);
```
Anty0: Dropping `onAddNew` removes the "+" from every consumer, not just the breadcrumb. `ProjectCreateView.tsx` renders `<OrganizationSwitch ownedOnly />`; a maintainer who can reach the create-project page but owns no organization now sees an empty "Nothing found" dropdown with no way to create one from there. The PR description says only the breadcrumb loses the button. Either keep `onAddNew` wired for the `ownedOnly` surface, or state the wider effect in the description.

Anty0 (changes requested):
Lgtm, just a few small things, two of which I just don't see why they were done this way. ^^

webapp/src/views/projects/BaseProjectView.tsx
dkrizan: I think this was already decided in #2737. Stepan suggested keeping the project name as a link to the dashboard, with only the arrow opening the switcher (https://github.com/tolgee/tolgee-platform/issues/2737#issuecomment-2511394557), and Zuzana confirmed it with the Figma design: "on click the project name - will go to the dashboard, on click the arrow - will open a dropdown with projects" (https://github.com/tolgee/tolgee-platform/issues/2737#issuecomment-2511611777).

So I would keep it this way. I'll add this to the PR description.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
dkrizan: Changed. `ProjectSwitch` now renders the avatar, the name link and the arrow in one element and anchors on that, same as `OrganizationSwitch`. The comment is gone.

One side effect: `Navigation` did not render `suffix` for element names, so the branch selector would have disappeared. I added that (`Navigation.tsx`), the other two callers with element names pass no suffix.

e2e/cypress/e2e/projects/projectSwitching.cy.ts
dkrizan: Added tests for search and load more. They needed an organization with more than 20 projects, so there is a new `project-switching` fixture (`ProjectSwitchingTestData`, 22 projects) with a controller on `AbstractE2eDataController`, and the spec now uses `generateStandard()`. The load more button in `SwitchPopover` got a `data-cy` for this.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,131 @@
...
+    setIsOpen(false);
+    history.push(
+      LINKS.PROJECT_DASHBOARD.build({ [PARAMS.PROJECT_ID]: selected.id })
+    );
```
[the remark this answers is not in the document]
dkrizan: Fixed. Selecting the current project now just closes the popover, so the branch in the URL stays as it is.

webapp/src/component/projectSwitch/ProjectSwitch.tsx
```diff
@@ -0,0 +1,134 @@
...
+        >
+          <SmallProjectAvatar project={project} />
+          {project.name}
+        </StyledLink>
+        <IconButton
+          sx={{ p: 0, color: 'inherit' }}
+          onClick={() => setIsOpen(true)}
+          data-cy="project-switch"
+          aria-label={t('projects_switch_label', 'Switch project')}
+          size="small"
```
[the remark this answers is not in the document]
dkrizan: Added `aria-haspopup="listbox"` and `aria-expanded`. I skipped `aria-controls` - the listbox is unmounted while the popover is closed, so the attribute would point at an element that does not exist most of the time.

PR #3849, opened 2026-08-07
feat: OAuth 2.1 authorization server (round 1)

bdshadow:
Implements round 1 of OAuth 2.1 authorization-server support for Tolgee (issue #3826): Tolgee becomes an OAuth 2.1 authorization server so non-webapp clients (starting with the browser extension) can authorize with exactly a community-level, consented, project-scoped credential instead of a long-lived PAK/PAT.

## What's included (backend + consent UI)

- **Spring Authorization Server** chain alongside the stateless main chain: Authorization Code + PKCE (S256), public (`NONE`) clients, self-contained **RS256** access tokens, JWKS, rotating refresh tokens (incl. a custom generator + client-auth so public clients can refresh, which stock SAS withholds).
- **Signing keys**: persisted RSA keypair via `FileStorage` (stable `kid` across restarts/replicas), first-boot generation serialized with `LockingProvider`.
- **Token claims & enforcement**: `scope`, a project-set claim `tg.prj`, and `aud` (single REST resource this round). `OAuth2AccessTokenResolver` dispatches Bearer tokens by the `aud` claim and validates RS256 against the JWKS; `SecurityService` intersects the token's scope ∩ project-set ∩ the user's live permissions at every project checkpoint (including the task-assignee fallbacks and `checkAnyProjectPermission`).
- **Consent flow**: SPA-served `/oauth2/consent` + `/oauth2/bootstrap`; a `session-bootstrap` endpoint turns the webapp JWT into an HTTP session (session-fixation-safe). The consent HTTP session is backed by **Spring Session JDBC** so the flow works across replicas without ingress affinity.
- **CIMD** (Client ID Metadata Documents) for URL-form client ids: SSRF-hardened outbound fetch, **fail-closed** (off by default + requires an allow-list), bounded cache. No round-1 consumer (MCP is deferred).
- **Connected apps**: list/revoke a user's authorized apps (`/v2/user/connected-apps`); nightly cleanup of expired and abandoned pre-consent authorizations; RFC 9728 protected-resource metadata.
- **Migrations**: SAS + Spring Session schema via Liquibase (indexed; excluded from diff tooling).

## Out of scope (deferred, per the pitch's No-Goes)

chrome-plugin (Part B) and `@tolgee/web` (Part C) client work, and MCP client / CLI / Figma / DCR / org-level scopes / per-resource (RFC 8707) audiences / automated key rotation.

## Notes

[...]

PR #3879, opened 2026-08-28
review: OAuth 2.1 authorization server

Anty0:
Review of #3849, as changes rather than comments. Base is `bdshadow/oauth-authorization-server`, so this diff is only what the review changed — 13 commits on top of your branch.

Everything here went through a multi-reviewer loop until it converged (4 rounds, ending with 22/23 reviewers approving). Each security fix below was mutation-checked: the fix was reverted and the test confirmed to fail.

## Two things worth your attention first

**1. The consent project picker cannot narrow to a project the user chooses.** `projectOptions` holds only the client's own `project` hint, so a client that sends no hint (the CLI, or an extension connect on a page with no SDK project id) offers "All projects" and nothing else — which is also the preselected default. Not a regression, it has been this way from the start, but it undercuts the project-scoped credential goal. Either populate the picker from the user's accessible projects, or make the default fail-safe (no preselection, Allow disabled until chosen). **Left for you to decide — not changed here.**

**2. Consent is no longer remembered, and this was forced.** Details below; it is the one behavioural change in this PR I would not have made unprompted.

### Remembered consent bricked every reconnect

Consent persisted per client+user, so SAS skipped the screen on the next connect. `PROJECT_ATTRIBUTE` is written only by `OAuth2FlowController.selectProject`, which only the consent SPA calls — so a skipped screen meant no project selection, and the token endpoint refused with `invalid_request`. Both shipped clients set `requireConsent = true`, so a user who approved everything was bricked on their second connect, with changing their password as the only self-service recovery. (A user who *narrowed* scopes was fine, because their stored consent no longer covered the request and the screen re-showed.)

`AlwaysPromptConsentService` fixes it: the screen runs every time. The project is chosen there and is per-authorization state with nowhere durable to live, so remembering consent needs the selection to become durable first — a persisted grant entity holding approved scopes *and* projects. That is the follow-up that unlocks consent-skip; until then re-prompting is what lets a reconnect mint a token at all. `oauth2_authorization_consent` is now written and read by nothing; dropping it is a migration decision, so the table stays.

### Other fixes

- **A de-registered client answered 500.** Dropping a client from config is the documented way to switch it off, but its grants outlive it and SAS throws when it cannot re-hydrate the client for a row. Now refused as an invalid token.
- **OAuth tokens reached `@IsGlobalRoute` handlers under `/v2/projects/**`.** `ProjectAuthorizationInterceptor` returns before narrowing anything for those, so a matching path never meant the token got narrowed. Nothing was exploitable today — I enumerated every `@IsGlobalRoute` site — but the next `@AllowApiAccess(ANY)` one added would have handed every consented token the user's whole account.
- **The consent scope tree stopped blocking dependencies.** `Hierarchy.tsx` forked on `lockedScopes`; any array is truthy, so the CLI client's empty `requiredScopes` switched blocking off for the whole tree, letting a user uncheck `keys.view` while `keys.edit` stayed checked.
- **The bootstrap continue URL repeated the servlet context path** — latent, since Tolgee sets none.
- OAuth failures get their own message codes instead of borrowing `expired_jwt_token`.

## Structure

- `asScopedCredential` loses its default on the `PermissionService` / `SecurityService` signatures, so a new call site has to answer it; all 30 existing ones are explicit.
- `OAuth2IssuerResolver` owns issuer derivation — `issuerUrl` for in-request callers, `configuredBaseUrl` for the startup bean — replacing three call sites with three different fallbacks and two trailing-slash trims.
- `OAuth2ServerProperties` moves into `io.tolgee.configuration.tolgee` as a `TolgeeProperties` property with per-field `@DocProperty`; three of its five properties were invisible to the generated config docs.
- The remaining OAuth2 collaborators move into `io.tolgee.security.oauth2`.
- `revokeAllForPrincipal(String)` becomes `revokeAllForUser(Long)`, so a caller cannot pass a username to a deletion that would silently match nothing.

One factual note for the RFC 9207 code: I checked the SAS 7.1 bytecode, and its authorization response emits only `code` and `state`. We are the sole source of `iss`, not preserving Spring's — the comment said otherwise. It is now emitted unconditionally, derived per request when nothing is configured.

## Tests

New coverage for the OAuth rejection paths in both interceptors and MCP, the consent-info scope source, the bootstrap origin, the issuer resolver, the de-registered client, both previously unpinned `LanguageService` branches, granular `tasks.assigned-access` permissions, the consent-tree disable predicate, and `selectConsentProject`'s `projectId` branch — the one that narrows the token, which had no coverage at any layer.

## Deployment notes

- **`server.forward-headers-strategy` was removed** (earlier in this branch). Proxied deployments must set `tolgee.back-end-url` or the issuer, discovery document and bootstrap redirect will carry the internal URL. The local-dev walkthrough in the README now says so too.
- No backfill for `tasks.assigned-access`: existing granular permission rows and project API keys lose the assignee elevation until the scope is granted. Deliberate, per the product decision.

## Known follow-ups, not done here

- `OAuth2BootstrapView` re-implements the global 401 handling, causing a double logout and a toast the silent-redirect path was meant to replace. Your code, not security-relevant, and no test pins the current behaviour — left for you.
- The two new message codes have no SPA translations; the SPA is not an OAuth client, and new keys need a decision.
- Whether `permittedLanguageIds` should come from `computedPermissions` rather than `directPermissions` is pre-existing for all credential kinds.


Anty0:
Superseded by #3893.

PR #3880, opened 2026-08-28
refactor: replace Spring Authorization Server with an in-house OAuth 2.1 server

JanCizmar:
## Why

Tolgee needs one narrow slice of OAuth: authorization-code + PKCE for public clients, opaque tokens, a per-authorization project picker, and a stateless webapp. Bending Spring Authorization Server to that needed ~450 lines of adapters (session bootstrap trio, public-client refresh converter/provider, consent-service stub, token customizer, registered-client repository), a 33-column schema and a consent table nothing wrote, and the discovery document advertised auth methods and grants no client could use.

Owning the protocol turns out to be *smaller*: net production Kotlin is **+842 / −1,085** (−243), with nothing left to work around.

## How this was done

1. **`test:` commit**: 25 black-box HTTP tests (`OAuth2ProtocolConformanceTest`) written against the SAS implementation first, pinning everything that was previously trusted to the library without a test: redirect_uri / client_id validation, unregistered scopes, the `plain` PKCE method, consent denial and over-approval, code replay revoking issued tokens, cross-client and redirect-mismatched exchange, unsupported grants, refresh scope widening / client binding, token response shape + TTL wiring, discovery. They import no library type and use only the clients registered from `tolgee.oauth2.*` in the test config.
2. **`refactor:` commit**: remove the library and everything that existed for it, replace with:
   - `OAuth2Authorization` entity + repository, lean migration (drops the SAS tables)
   - `OAuth2ClientRegistry` (extension + CLI from config; exact redirect match, PKCE + consent always)
   - `OAuth2AuthorizationService`: validation, consent, single-use code exchange, refresh rotation + narrowing, revocation on replay / cross-client use / expiry / token invalidation
   - `OAuth2AuthorizationServerController`: `/oauth2/authorize` (request + consent form), `/oauth2/token`, RFC 8414 discovery

The conformance tests pass **unchanged** against the new implementation. The existing flow, resolver, cleanup and EE narrowing tests are kept, now driven through the config-registered clients and the entity instead of library types.

## Behaviour changes

- Codes and tokens are stored **hashed** (SAS stored plaintext).
- Grants have a FK to `user_account`; deleting an account revokes its grants (SAS left orphan rows).
- Discovery now lists `token_endpoint_auth_methods_supported: [none]` and only `authorization_code` / `refresh_token`.
- Consent client guard kept: a code is redeemable only once `select-project` bound a project set.

Webapp untouched (same consent URL params and form POST). Spring Session JDBC stays: the JWT→session bridge for the top-level navigation is a property of the stateless webapp, not of the library.

## Tests

- `OAuth2ProtocolConformanceTest` 25/25, `OAuth2AuthorizationCodeFlowTest` 30/30, `OAuth2AccessTokenAuthTest` 26/26, cleanup/service/discovery/protected-resource tests, `:data` and `:security` OAuth tests, EE narrowing tests: all green locally. `ktlintCheck` clean.
- Not run locally: the Cypress consent spec (needs the docker e2e stack). The URL contract it relies on is pinned by backend tests.
- Test Postgres containers need a one-time `docker rm -fv` because the schema changed.

## Known follow-ups (unchanged from before)

Refresh is plain rotate-on-use (no grace window / reuse-detection family revocation); grant lookup is uncached on the API hot path. Both are documented in `docs/oauth/README.md`.


Anty0:
Superseded by #3893.

PR #3885, opened 2026-08-29
fix: make the in-house OAuth 2.1 server release-ready

Anty0:
Polishes the in-house OAuth 2.1 authorization server for release and closes the correctness gaps found while reviewing it. Reviewed against local copies of RFC 6749, 6750, 7009, 7591, 7636, 8252, 8414, 9207, 9700, 9728 and OAuth 2.1 draft-13.

## BREAKING: granular permissions and project API keys lose the task-assignee elevation

`Scope.TASKS_ASSIGNED_ACCESS` gates the task-assignee elevation that used to be unconditional. Role presets expand at
runtime and pick it up, but **granular permissions and project API keys store their scope list literally**, so on
upgrade they simply no longer carry it: a member or key that could view and edit a task purely because it was assigned
to them now needs `tasks.assigned-access` granted explicitly.

**There is deliberately no backfill.** An earlier revision of this PR added one. Restoring the bypass that way means
granting a brand new scope to every permission and API key that holds any scope at all — a lot of grants widened at
once, and not distinguishable afterwards from scopes an admin chose. Losing an elevation fails closed; granting one
does not. The changeset, its rollback and its test were removed.

Operators upgrading should expect this and grant the scope where it is wanted. Role-based permissions need nothing.

## Protocol correctness

- **`/oauth2/authorize` returned 500** for any query string that is not strictly URI-encoded. RFC 6749 §4.1.1 allows any printable ASCII in `state`, and the endpoint replayed the raw query through a builder that *verifies* it — after `client_id` and `redirect_uri` had already validated. The consent URL is now built from the bound parameters.
- **`state` could come back byte-different.** A space written as a bare `+` reads as a space to a form decoder and as a literal plus to a percent decoder; RFC 6749 does not say which a client uses, and a mismatched `state` must be treated as an attack (§10.12). Outgoing values are now RFC 3986 percent-encoded.
- **A relative configured redirect URI matched**, making the code redirect resolve against Tolgee's own origin — the authorization code would go back to Tolgee instead of the client. Configured URIs are now validated as absolute, fragment-free and https-unless-loopback.
- **`invalid_client` answered 401 with no `WWW-Authenticate`** (RFC 9110 §15.5.2 makes that a MUST) on a server advertising `token_endpoint_auth_methods_supported: ["none"]` — nothing to challenge with. Now 400, the RFC 6749 §5.2 default.
- An over-long `state` is refused on the redirect, where the client can still be told, instead of dead-ending on the consent screen after the user has burned a login.
- Repeated spec parameters are rejected on both endpoints (OAuth 2.1 §3.1).
- An unparseable `back-end-url` / `front-end-url` now fails loudly; the validator previously accepted the one bad value it could not inspect.

## Storage and schema

- Grants persist `Scope` **names**, as `ApiKey.scopesEnum` does. `Scope.value` is a `var` reshaped for API cosmetics; renaming one would otherwise silently narrow every live grant that held it.
- `previous_refresh_token_hash` is indexed — it was the only looked-up hash column without one, on the path every failed refresh takes.
- Only the Spring Authorization Server tables are dropped. `spring_session` / `spring_session_attributes` are stock Spring names owned by no Tolgee release, so on a schema shared with another Spring application the drop would have silently signed out that application's users.

## Cleanup

The Spring Session bridge, its `spring-session-jdbc` dependency and its changelog are gone — the consent flow is stateless and carries no cookie. OAuth endpoint paths, the blank-to-null helper and the scoped-credential rule each have one home now.

## Deferred (agreed)

The all-projects option stays. An all-projects token has no route that can enumerate its projects — both
project-listing routes are `@IsGlobalRoute` (denied for OAuth); `getAll` is `ONLY_PAT` besides, and
`getAllWithStatistics` carries no `@AllowApiAccess` at all, so it admits no API credential. Not a boundary
problem: every
request is still narrowed by `coversProject()` and the user's live permissions, so the failure is "cannot discover",
never "reaches further than intended". The browser extension is handed its project by the page it edits; MCP is the
round that needs discovery. Tracked in `docs/oauth/README.md` under round-1 limitations, together with
`implicitProjectId` currently reading `*` as "no project".

## Second review pass

A second multi-reviewer loop (24 rounds, converged to approved) ran over the whole stack. What it changed:

- **The release was not rollbackable.** The backfill wrote a `Scope` constant only this release's binary can
  resolve into `permission.scopes` and `api_key_scopes_enum`, behind an empty `<rollback/>`. Rolling the image
  back would have thrown `Enum.valueOf` on the hot authentication path — every scoped API-key request failing,
  recovery needing hand-written SQL under incident pressure. There is now a real rollback, exercised by tests
  against real rows. A permission whose only scope was the new one becomes role-based `NONE`, because an emptied
  granular row is un-persistable: `PermissionListeners` nulls the empty array and then requires exactly one of
  `scopes` or `type`.
- **CORS at the authorization endpoint.** OAuth 2.1 §3.1 says CORS **MUST NOT** be supported there; the endpoint
  is only ever navigated to. Class-level CORS was removed in favour of per-method.
- **RFC 7009 §2.1**: revocation now verifies the token was issued to the requesting client and refuses if not.
- **RFC 8414 §2**: an issuer with a query or fragment was accepted and would have been published as `iss` and
  concatenated into every endpoint URL. Only a bare origin passes now.
- **A stale `Authorization` header could 401 the recovery path.** The authorization-server paths authenticate
  nobody, so they no longer resolve a credential — including the RFC 9728 document a 401 points a client at. The
  filter still runs, so they keep the per-IP auth rate limit.
- **`getCurrentPermissions` disclosed a project the caller cannot see.**
- **Organization routes** refuse OAuth tokens above the "no organization resolved" early return, so a route
  opened with `@AllowOAuthAccess` cannot pass through ungated.
- `exchangeCode` rejects a blank project selection (an empty one bound `""`, minting a token reaching no
  project); `matchesLoopback` compares userinfo, which its contract already claimed; a mis-annotated
  `@AllowOAuthAccess` endpoint answers 403 instead of 500; the dev proxy is opt-in.

## Follow-up for a separate PR: account-level endpoints take any API credential

OAuth tokens are now gated by `@AllowApiAccess.tokenType` alone, the same as project API keys - the previous
extra rule (must be on a project-scoped path, with `@AllowOAuthAccess` as an escape hatch) is gone, and so is
that annotation.

That makes these reachable by an OAuth token, exactly as they already are by a project API key:
`/v2/user`, `/v2/notification`, `/v2/notification-settings`, `/v2/notifications-mark-seen`, `/v2/user-tasks`,
`/v2/image-upload`.

None of them resolves a project, so `SecurityService.getCurrentPermittedScopes` never runs and no credential's
scopes are consulted on them - a project API key reads them today for the same reason. `/v2/user` is the one
worth weighing: it returns the account's email and server role.

These should most likely be `ONLY_PAT`, but that is a question about **every** API credential rather than about
OAuth, and changing it would alter existing project-API-key behaviour. Deliberately left for its own PR.

Project enumeration is unaffected: `ProjectsController.getAll` stays `ONLY_PAT`, and it narrows by user id
alone (`findPermittedInOrganizationPaged`), not by the credential - so making it `ANY` would need token
scoping added first.

## Verification

`:data:test`, `:security:test`, `:server-app:test` for `io.tolgee.security.*` and `io.tolgee.mcp.*` plus
`ApiKeyControllerTest` and the translation controllers, `ktlintCheck` on all four touched modules, `webapp`
eslint 0 / tsc 0, and `diffChangeLog --rerun-tasks` clean against a from-scratch schema database.

**Not run:** the Cypress specs, including the new `oauth2Consent.cy.ts`. `gradle/docker.gradle` shells out to
`docker buildx`, which is not installed on the machine this was prepared on, so the harness fails before Cypress
starts. Nothing in the specs is known-broken; they are simply unverified here and need a run in CI or on a host
with buildx. The one genuinely unverified expectation is the new `tasks.assigned-access` E2E case's menu-item
set — that scope's backend behaviour is already covered by `TaskAssignedAccessScopeTest`.

## Review notes

Converged through two automated multi-reviewer loops (31 rounds, then 24). Two things a human reader should weight: the authorize endpoint and the refresh-rotation / scope-persistence paths were each restructured more than once under review pressure, and several defects were introduced while fixing earlier ones. The `spring-session-jdbc` dependency removal above was found only by re-checking a reviewer that had stood itself down.

Two decisions from the second loop are recorded rather than acted on: a client's `requiredScopes` stay a
consent-screen affordance rather than a server-side rule (RFC 6749 §3.3 permits issuing narrower than requested,
and it is the user's own account), and `AuthenticationFacade.isScopedCredentialFor` keeps its free-function twin
because `PermissionService` cannot take the facade without the dependency cycle the free function exists to
avoid. Both are documented where the next reader will hit them.


Anty0:
Superseded by #3893.

PR #3887, opened 2026-08-30
feat: app audit logging — activity, PostHog bridge, OTEL (apps alpha 7/7)

JanCizmar:
Slice 7 of 7 — the final slice. Adds **audit logging** for app operations: activity revisions with entity-level change tracking, PostHog attribution via the existing activity bridge, and OTEL metrics. Storage only — **no read API/UI**.

## 1. Activity / audit log
- `@ActivityLoggedEntity` on `App`, `AppInstall`, `AppSecret`. Logged props: `AppInstall.grantedScopes` (the audit-critical one), `App.name`/`version`/`baseUrl`/`manifestScopes`, `AppSecret.revokedAt`/`expiresAt`. Churny manifest-health/timestamp fields are `@ActivityIgnoredProp` so they don't create noise revisions.
- **`grantedScopes` is `Array<Scope>`** — a custom `ScopeArrayPropChangesProvider` diffs it by sorted scope-value set (the default provider compares arrays by reference and never yields a meaningful old→new). A re-consent records `[view,edit] → [view,edit,create]`.
- New `ActivityType`s: `APP_REGISTER`, `APP_INSTALL`, `APP_UNINSTALL`, `APP_UPDATE`, `APP_ENABLE_FOR_PROJECT`, `APP_DISABLE_FOR_PROJECT`, `APP_SECRET_UPDATE`, `APP_AVAILABILITY_UPDATE`. The ones whose operation doesn't dirty a logged prop (enable/disable touch `AppEnabledForProject`; availability uses a native insert; secret ops) carry `saveWithoutModification = true`.
- Set via `@RequestActivity(...)` on the owner/project/self/admin endpoints. App-self refresh has no `OrganizationHolder`, so `AppActivityRecorder` pre-seeds `activityRevision.organizationId` (`ActivityRevisionInitializer` preserves it) → the revision is org-scoped.

## 2. PostHog (via the activity bridge, no manual publish)
`AppActivityRecorder` sets `ActivityHolder.businessEventData["appId"/"appName"]` before the operation's transaction commits; `InterceptedEventsManager` carries it into `OnProjectActivityEvent` → `BusinessEventReportingActivityListener` → PostHog. So each app op's PostHog event is named after the `ActivityType` and carries the app id + name — no duplicate/manual event.

## 3. OTEL
`Metrics.kt`: `tolgee.apps.*` counters — `recordAppRegistered`, `recordAppInstalled`, `recordAppEnabledForProject`, `recordAppTokenMinted`, `recordAppLifecycleDelivery(eventType, outcome)`. `@WithSpan` on `AppManifestFetcher.fetch` and `AppLifecycleDeliveryService.deliverNow` with `app.*` span attributes.

## Scope
No read/list API. `AppEnabledForProject`/`AppAvailability` are deliberately not logged entities (avoids capturing cascade-cleanup deletes as unrelated revisions) — their audit rides `saveWithoutModification` + a seeded project/org.

## Tests
`AppActivityTest` (6): APP_REGISTER org-scoped, APP_INSTALL, APP_ENABLE_FOR_PROJECT (projectId set), owner refresh records grantedScopes old→new, app-self refresh stores an org-scoped revision without an OrganizationHolder, businessEventData carries appId/appName. Existing apps suites stay green.

## Note for reviewers
An app-level token authenticates as `syntheticAppPrincipal(appId)`, so an app-self action's `authorId` is the app's entity id (not null) — the bridge fires for app-self actions attributed to that synthetic id. The test asserts the org-scoping guarantee (the load-bearing behavior).

### DIFF
--- a/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationAppsController.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationAppsController.kt
@@ -2,6 +2,8 @@ package io.tolgee.api.v2.controllers.organization
 
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.tags.Tag
+import io.tolgee.activity.RequestActivity
+import io.tolgee.activity.data.ActivityType
 import io.tolgee.dtos.request.RegisterAppRequest
 import io.tolgee.hateoas.organization.apps.AppInstallModel
 import io.tolgee.hateoas.organization.apps.AppInstallModelAssembler
@@ -54,6 +56,7 @@ class OrganizationAppsController(
 
   @PostMapping
   @RequiresOrganizationScopes([Scope.ORGANIZATION_APPS_MANAGE])
+  @RequestActivity(ActivityType.APP_INSTALL)
   @Operation(
     summary = "Install a Tolgee app",
     description =
@@ -91,6 +94,7 @@ class OrganizationAppsController(
   @PostMapping("/{installId}/refresh")
   @RequiresOrganizationScopes([Scope.ORGANIZATION_APPS_MANAGE])
   @RequiresSuperAuthentication
+  @RequestActivity(ActivityType.APP_UPDATE)
   @Operation(
     summary = "Refresh an installed app's manifest and approve its current scopes",
     description =
@@ -108,6 +112,7 @@ class OrganizationAppsController(
 
   @DeleteMapping("/{installId}")
   @RequiresOrganizationScopes([Scope.ORGANIZATION_APPS_MANAGE])
+  @RequestActivity(ActivityType.APP_UNINSTALL)
   @Operation(
     summary = "Remove app",
     description = "Uninstalls the app from the organization.",
--- a/backend/data/src/main/kotlin/io/tolgee/model/apps/AppSecret.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/model/apps/AppSecret.kt
@@ -1,5 +1,7 @@
 package io.tolgee.model.apps
 
+import io.tolgee.activity.annotation.ActivityLoggedEntity
+import io.tolgee.activity.annotation.ActivityLoggedProp
 import io.tolgee.model.StandardAuditModel
 import jakarta.persistence.Column
 import jakarta.persistence.Entity
@@ -16,6 +18,7 @@ import java.util.Date
  * to issuing it; only the hash is stored.
  */
 @Entity
+@ActivityLoggedEntity
 @Table(
   name = "app_secret",
   uniqueConstraints = [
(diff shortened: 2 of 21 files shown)


backend/data/src/main/kotlin/io/tolgee/service/apps/AppActivityRecorder.kt
```diff
@@ -0,0 +1,32 @@
...
+  private val activityHolder: ActivityHolder,
+) {
+  fun record(
+    app: App,
+    activityType: ActivityType? = null,
+    organizationId: Long? = null,
+    projectId: Long? = null,
+  ) {
+    activityType?.let { activityHolder.activity = it }
+    activityHolder.businessEventData["appId"] = app.id.toString()
```
JanCizmar, line 27: I meant also the string appId.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationAppsController.kt
```diff
@@ -54,6 +56,7 @@ class OrganizationAppsController(
 
   @PostMapping
   @RequiresOrganizationScopes([Scope.ORGANIZATION_APPS_MANAGE])
+  @RequestActivity(ActivityType.APP_INSTALL)
```
JanCizmar, line 59: How do you propagate the activity info to posthog for these events?

backend/data/src/main/kotlin/io/tolgee/model/apps/AppSecret.kt
```diff
@@ -16,6 +18,7 @@ import java.util.Date
  * to issuing it; only the hash is stored.
  */
 @Entity
+@ActivityLoggedEntity
```
JanCizmar, line 21: I would log also the name property changes.

backend/data/src/main/kotlin/io/tolgee/service/apps/AppActivityRecorder.kt
JanCizmar: [claude] Added — `businessEventData["appIdentifier"] = app.appId` (the string manifest id), alongside the numeric `appId` and `appName`. Test asserts all three reach the business event.

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/organization/OrganizationAppsController.kt
JanCizmar: [claude] `@RequestActivity` on the endpoint only sets the event *type*; the app id/name are attached by `AppActivityRecorder` in the service/persister layer — `install` → `AppInstallPersister.create`, `refresh` → `applySnapshot`, `remove` → `AppInstallPersister.remove` — each calls `appActivityRecorder.record(app)` which writes them into `ActivityHolder.businessEventData` before the transaction commits. `InterceptedEventsManager` forwards `businessEventData` into `OnProjectActivityEvent`, and `BusinessEventReportingActivityListener` maps it to the PostHog event's `data`. So all three of these endpoints propagate app id/name; verified by the `business-event bridge carries the app id and name` test (which drives the refresh endpoint).

backend/data/src/main/kotlin/io/tolgee/model/apps/AppSecret.kt
JanCizmar: [claude] Done — `@ActivityLoggedProp` added to `AppSecret.name`.

PR #3888, opened 2026-08-30
perf: cache the per-request app enablement check

JanCizmar:
First of the app-auth caching follow-ups (see the hot-path analysis). Every project-scoped app request runs `AppEnablementService.isEnabledForProject` → a DB query; this caches it.

## What
- New cache `Caches.APP_ENABLEMENTS`, keyed **`installId → Set<enabled projectId>`**, so the per-request check is a set-contains and an uninstall evicts a single key. Created on-demand under the uniform default TTL.
- `AppEnablementCache` is a **separate bean** (so the `@Cacheable` call isn't a self-invocation bypass) exposing `getEnabledProjectIds(installId)` + `evict(installId)` + `evictAll()`.
- `isEnabledForProject(projectId, installId)` → `projectId in appEnablementCache.getEnabledProjectIds(installId)`.

## Eviction (the correctness-critical part)
All evicts live **inside the existing `@Transactional` methods**, so the `TransactionAwareCacheManagerProxy` defers them to commit (a rolled-back change never evicts):
- `enable` → evict the install (new project shows up; also clears a stale negative entry).
- `disable` → evict the install.
- `removeAllForAppInstall` (uninstall) → evict the install.
- `removeAllForProject` (project delete, rare) → `evictAll()` — it touches many installs' sets; per-install eviction would need an extra query.

Missing an evict on enable/disable/uninstall would let a disabled or uninstalled app keep project access from a stale entry — so the tests pin exactly that.

## Tests
`AppEnablementCacheTest` (cache enabled via `tolgee.cache.enabled=true`, repository spied):
1. the check is served from cache on the second call (DB hit once);
2. **disable evicts** — no stale "enabled";
3. **uninstall evicts** — no stale "enabled";
4. **enable evicts the negative entry** — reads enabled immediately.
Existing `ProjectAppsControllerTest`, `AppTokenAuthorizationTest`, `AppRefreshTest` stay green.

## Scope
Item 1 only. Caching `App`/`AppInstall` (incl. the force-revoke eviction) needs the `AppAuthentication`/binder DTO refactor and is a separate follow-up.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/service/apps/AppEnablementCacheTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/service/apps/AppEnablementCacheTest.kt
@@ -1,0 +1,98 @@
+package io.tolgee.service.apps
+
+import io.tolgee.AbstractSpringTest
+import io.tolgee.development.testDataBuilder.data.AppsWithInstallsTestData
+import io.tolgee.repository.apps.AppEnabledForProjectRepository
+import io.tolgee.service.apps.lifecycle.AppLifecycleHttpClient
+import io.tolgee.testing.assert
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.mockito.Mockito
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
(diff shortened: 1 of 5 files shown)


JanCizmar:
Superseded by #3890 — the caching work is combined into a single PR. Closing this one.

PR #3889, opened 2026-08-30
perf: cache App and AppInstall on the app-auth path (auth-cache items 2 & 3)

JanCizmar:
The second app-auth caching follow-up (items 2 & 3 of the hot-path analysis). Moves the per-request app-auth loads off JPA entities onto cacheable DTOs, so an app request doesn't reload the App and AppInstall every time. **Security-preserving: the force-revoke kill switch still stops tokens immediately.**

## DTOs (serializable, `io.tolgee.dtos.cacheable`)
- `AppDto(id, tokensInvalidBefore: Long?)` — deliberately minimal, so a manifest refresh (which never touches `tokensInvalidBefore`) needs no `APPS` eviction.
- `AppInstallDto(id, appEntityId, organizationId, principalUserId, grantedScopes: Array<Scope>)` — refers to the app by id, so a force-revoke touches only the one app snapshot.

## What's cached
`AppInstallService.findForAppAuth` → `@Cacheable(APP_INSTALLS, key=installId)`; `findAppForAppAuth` → `@Cacheable(APPS, key=appEntityId)`. `resolveForAppAuth` composes the cached install DTO (via `@Lazy self` so the proxy isn't bypassed) with the already-cached `userAccountService.findDto(principalUserId)` — zero queries when warm. `AppAuthentication` now holds `AppInstallDto`; `AppTokenAuthenticator`/`AppProjectContextBinder`/`SecurityService` read DTO fields; `checkAppTokenCutoff(AppDto)` keeps the whole-second `iat` comparison.

## Eviction (all inside the existing `@Transactional` methods → deferred to commit)
- **`AppSecretService.revoke` → `@CacheEvict(APPS, key="#appId", condition="#force")`** — the force-revoke kill switch. Non-force revoke doesn't change the AppDto, so the condition scopes it exactly.
- `AppInstallPersister.remove` → evict `APP_INSTALLS[installId]` (uninstall).
- `AppInstallPersister.applySnapshotForOrgInstall` / `applySnapshotForApp` → evict `APP_INSTALLS[installId]` (owner re-consent and app self-refresh both reduce `grantedScopes`; a stale wider set would be a hole).
- `AppOwnerRemovalService` → evict each removed install from `APP_INSTALLS` + the app from `APPS`.
- enable/disable untouched (that's the separate enablement cache).

## Tests
`AppAuthCacheTest` (cache enabled via `tolgee.cache.enabled=true`), all pass:
1. **force-revoke kills immediately** — app-level + install tokens authenticate, then a force-revoke evicts `APPS[appId]` and both tokens get 401 on the very next request (without the evict, the cached AppDto keeps `tokensInvalidBefore=null` and they'd keep working).
2. **uninstall evicts** — the install token 401s on the next request.
3. **scope re-consent evicts** — a refresh that drops a granted scope is enforced immediately (403), not stale-allowed.
Existing `AppCredentialTokenTest`, `AppTokenAuthorizationTest`, `AppRefreshTest`, `AppInstallPrincipalTest` stay green.

## Residual note
On a Caffeine-only (non-Redis) multi-node deployment a force-revoke on one node wouldn't evict the others' local `APPS` cache until TTL. Tolgee's shared-cache deployments use Redis, where the transaction-aware evict is authoritative.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/apps/AppAuthCacheTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/apps/AppAuthCacheTest.kt
@@ -1,0 +1,228 @@
+package io.tolgee.api.v2.controllers.apps
+
+import io.tolgee.constants.Caches
+import io.tolgee.development.testDataBuilder.data.AppsTestData
+import io.tolgee.dtos.cacheable.AppDto
+import io.tolgee.dtos.cacheable.AppInstallDto
+import io.tolgee.fixtures.andIsOk
+import io.tolgee.service.apps.AppManifestHttpClient
+import io.tolgee.service.apps.AppsTestFixtures
+import io.tolgee.service.apps.lifecycle.AppLifecycleHttpClient
+import io.tolgee.testing.AuthorizedControllerTest
+import io.tolgee.testing.ContextRecreatingTest
+import io.tolgee.testing.assert
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.webmvc.test.autoconfigure.AutoConfigureMockMvc
+import org.springframework.http.HttpHeaders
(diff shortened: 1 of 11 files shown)


JanCizmar:
Superseded by #3890 — the caching work is combined into a single PR. Closing this one.

PR #3890, opened 2026-08-30
perf: cache the app request hot path (enablement + App/AppInstall auth)

JanCizmar:
Caches the DB lookups an app runs on every request, from the hot-path analysis. Eviction (not TTL) is the safety mechanism — a disabled/uninstalled/force-revoked app must lose access immediately, so the tests pin exactly that.

## Caches added
- **`APP_ENABLEMENTS`** — `installId → Set<enabled projectId>`. `AppEnablementService.isEnabledForProject` (per project-scoped request) becomes a set-contains. Separate `AppEnablementCache` bean (avoids the `@Cacheable` self-invocation trap).
- **`APPS`** — `AppDto{id, tokensInvalidBefore}` (minimal, so a manifest refresh needs no eviction). Carries the force-revoke cutoff.
- **`APP_INSTALLS`** — `AppInstallDto{id, appEntityId, organizationId, principalUserId, grantedScopes}`.

## Auth path moved off entities onto DTOs
`AppInstallService.findForAppAuth`/`findAppForAppAuth` are `@Cacheable`; `resolveForAppAuth` composes cached DTOs (via `@Lazy self`) + the already-cached `userAccountService.findDto`. `AppAuthentication` holds `AppInstallDto`; `AppTokenAuthenticator`/`AppProjectContextBinder`/`SecurityService` read DTO fields; `checkAppTokenCutoff(AppDto)` keeps the whole-second `iat` comparison.

## Eviction (all inside the existing `@Transactional` methods → deferred to commit)
- **`AppSecretService.revoke` → `@CacheEvict(APPS, key="#appId", condition="#force")`** — the force-revoke kill switch.
- `AppEnablementService` enable/disable/removeAllForAppInstall → evict `APP_ENABLEMENTS[installId]`; removeAllForProject → `evictAll()` (rare).
- `AppInstallPersister.remove` + both refresh paths (`applySnapshotForOrgInstall`/`ForApp`) → evict `APP_INSTALLS[installId]` (uninstall + scope re-consent).
- `AppOwnerRemovalService` → evict each removed install + the app.

Missing any of these would leave a disabled/uninstalled/revoked app authenticating from a stale entry until TTL — so the tests assert immediate loss of access.

## Tests (cache enabled via `tolgee.cache.enabled=true`)
- `AppEnablementCacheTest` (4): served-from-cache; disable/uninstall evict (no stale "enabled"); enable evicts the negative entry.
- `AppAuthCacheTest` (3): **force-revoke kills already-minted app + install tokens on the next request**; uninstall evicts; a refresh that drops a scope is enforced immediately (403).
- Existing `AppCredentialTokenTest`, `AppTokenAuthorizationTest`, `AppRefreshTest`, `ProjectAppsControllerTest`, `AppInstallPrincipalTest` all green.

## Residual note
On a Caffeine-only (non-Redis) multi-node deployment, a force-revoke on one node wouldn't evict the others' local `APPS` cache until TTL. Tolgee's shared-cache deployments use Redis, where the transaction-aware evict is authoritative.


backend/data/src/main/kotlin/io/tolgee/repository/apps/AppEnabledForProjectRepository.kt
```diff
@@ -49,6 +49,11 @@ interface AppEnabledForProjectRepository : JpaRepository<AppEnabledForProject, L
     pageable: Pageable,
   ): Page<Project>
 
+  @Query("select e.project.id from AppEnabledForProject e where e.appInstall.id = :appInstallId")
+  fun findEnabledProjectIdsByInstallId(
+    @Param("appInstallId") appInstallId: Long,
```
JanCizmar, line 54: no reason to introduce the param annotation.

PR #3893, opened 2026-08-31
feat: OAuth 2.1 authorization server

### LINKED ISSUE #3826: OAuth 2.1 authorization server for token-free client auth (MCP & browser extension)
### 1. Problem

Every non-webapp client that talks to a Tolgee server authenticates with a static, long-lived secret, a Project API Key (PAK) or Personal Access Token (PAT) sent as the `X-API-Key` header (`AuthenticationFilter.kt`). The user generates that key in the UI, copies it, and pastes it into the client.

The reason this matters most is community translation. With public projects, a contributor is a logged-in non-member who gets an implicit "community floor" of permissions (view, suggest, comment) on a public project, and there is no anonymous access. To translate in-context, that external person's browser extension needs a credential, and we cannot hand a stranger a project API key (a PAK is bounded to a member's own project access). OAuth is what lets an outside contributor authorize the extension with exactly their community-level access. This is the critical path.

The same paste-a-static-key friction hits our other headless clients:

- Browser extension and SDK (in-context editing, `@tolgee/web` and `tolgee/browser-extension`): the user pastes a key that the extension stores and the SDK replays as `X-API-Key`.
- MCP server (`/mcp/developer`, in the official MCP registry): users paste a PAK/PAT into MCP client config, landing in plaintext JSON on disk, out of step with the MCP OAuth 2.1 spec. Secondary priority, a power-user convenience rather than the driver.
- Tolgee CLI (`tolgee/tolgee-cli`): same paste-a-token model; users expect the seamless "open browser, approve, logged in" flow.

---

### 2. Appetite

2w

---

### 3. Solution

Make the Tolgee platform an OAuth 2.1 authorization server using `spring-security-oauth2-authorization-server`, exposing Authorization Code + PKCE so a client redirects the user to Tolgee, the user consents to a listed set of access, and the client receives a short-lived access token (plus refresh token) instead of a pasted key. All our clients are public clients (no secret); PKCE replaces the secret.

Core mechanics:

- A dedicated `SecurityFilterChain` for the authz-server endpoints (`/oauth2/authorize`, `/oauth2/token`, `.well-known/*`), coexisting with the existing stateless chain.
- Reuse the JWT stack (`JwtService`) as the access-token format so `AuthenticationFilter`'s existing Bearer path validates OAuth tokens. Tokens are audience-bound (RFC 8707): a token minted for MCP carries `aud = <MCP canonical URI>` and is rejected by the REST API, and vice versa.
- The scope model is the existing `Scope` enum wire values (`translations.suggest`, `translation-suggestions.manage`, `translation-comments.add`, and so on). Effective access is `expand(token.scopes)` intersected with `expand(user's live project scopes)`, evaluated per request. This is already exactly what `SecurityService.getCurrentPermittedScopes` does for PAKs. The token only narrows; losing a project or leaving the org kills access immediately.
- Client registration. Our own clients (extension, CLI): pre-registered public clients with a fixed `client_id` and redirect URIs (seeded rows, giving the extension a real `chrome-extension://` anchor and branded consent). Unknown third-party MCP clients: Client ID Metadata Documents (CIMD), where the client's `client_id` is an HTTPS URL we fetch and validate. CIMD is spec-preferred and supported by the MCP clients we target (Claude Code, Claude Desktop, VS Code Copilot). Implemented as a custom `RegisteredClientRepository`, since it is not in SAS. DCR is not implemented in v1 (see No-Goes).

Extension and SDK (the critical path, and a cross-repo change). This is three repos (platform, `tolgee/browser-extension`, and `@tolgee/web` pinned on customer sites), not a one-liner. It needs:

- a PAK-vs-OAuth discriminator so the SDK picks the right header;
- a Bearer branch in `@tolgee/web`'s `client.ts` (today it is unconditionally `X-API-Key`);
- token lifecycle the SDK has never had: 401, ask the extension, refresh, retry;
- capability negotiation via `TOLGEE_READY` (absence means an old SDK), so we carry both the `X-API-Key` and Bearer paths indefinitely;
- the refresh token stays in extension storage; the page/SDK only ever receives short-lived access tokens (page `sessionStorage` is XSS-reachable).

Scope and consent decisions (load-bearing, named now rather than left to the spike):

- Capabilities live in `scope`; the project set lives in a custom claim, since the client doesn't know project IDs before it is authorized.
- "All projects" is a sentinel (including future projects), never an expanded list, otherwise tokens blow past header limits for big orgs and users re-authorize on every new project. GitHub's wording is "all repositories, including future ones".
- A client that knows the project (Figma, the extension) may hint one; consent pre-selects and marks it required, other accessible projects optional. The server validates the user actually holds those scopes on the hint, so it can't leak IDs or nudge access.
- Re-consent / step-up is first-class, not just first login; it maps onto the MCP spec's `insufficient_scope` step-up flow.
- Effective access is evaluated per request (token intersected with live permissions, per the core mechanics above).
- Org-level operations don't fit a per-project selector: v1 covers project scopes plus "all projects"; org-level ops get their own consent scope or are a v1 no-go.

Other clients:

- MCP: Protected Resource Metadata (RFC 9728) on `/mcp/developer`, RFC 8707 audience, RFC 9207 issuer, so spec-compliant clients discover and run the flow, with CIMD for the ones we don't pre-register.
- CLI: seamless login via loopback redirect (`http://127.0.0.1:<port>`) plus PKCE. Keeps `--api-key` for CI and back-compat.

Cross-cutting:

- Connected apps: persist authorizations and revoke a single app, largely free from SAS's `OAuth2AuthorizationService`. The durable grant/refresh/consent store lives in Postgres (built-in JDBC); Redis (optional) is used only for ephemeral TTL data (authorization codes, a revocation denylist), never the system of record.
- Signing keys: today's JWT uses a symmetric HMAC secret auto-generated at boot. OAuth needs an asymmetric keypair plus a JWKS endpoint (external verification), persisted so restarts/replicas don't invalidate tokens, with a rotation-with-overlap story.
- Token lifetimes: short access TTL; refresh rotation with reuse detection plus a grace window tolerating near-simultaneous refresh; a defined grant-invalidation trigger.
- PAK/PAT stay for CI, scripts, and machine-to-machine (see No-Goes).

---

### 4. Basic Drawings

---

### 5. Rabbit Holes

- Scope-to-permission mapping: expose the `Scope` enum as OAuth scopes and reuse the existing `getCurrentPermittedScopes` intersection, including the public-project community floor (derived server-side per request, independent of token scopes). The consent UI reuses the permissions-dialog content plus a project selector ("all projects").
- CIMD implementation (in scope): not in Spring Authorization Server, so we build a custom `RegisteredClientRepository` that resolves the URL-form `client_id`, fetches the metadata doc, validates `client_id` against `redirect_uris`, advertises `client_id_metadata_document_supported` in AS metadata, invalidates consents when a client's `redirect_uris` change, and hardens the outbound fetch against SSRF (allow-list, no internal IPs, size and timeout limits) with caching. This is the security-critical surface of the cycle.
- Extension version negotiation: customer sites pin `@tolgee/web`; an old SDK reading an OAuth token as `X-API-Key` gets a 401. A capability flag on `TOLGEE_READY`, both paths carried for a long time.
- Refresh-token rotation races: clients refresh proactively before expiry, so rotation must tolerate near-simultaneous attempts without tripping reuse-detection.
- Audience binding (RFC 8707): `JwtService` emits one fixed audience today; OAuth needs a per-resource `aud` and rejection of wrong-audience tokens.
- Self-hosted keys and discovery: persisted asymmetric keys, JWKS, issuer and canonical URIs, zero-config on a fresh instance, correct behind reverse proxies (external URL is not the internal URL).
- Two filter chains on Security 7: the authz-server chain coexisting with the stateless chain on the freshly-migrated #3804 baseline.

---

### 6. No-Goes

- Not full/unified session management. Connected-apps plus per-OAuth-grant revocation is in; revoking the webapp's own stateless JWT sessions, "log out everywhere", or a unified view across JWT/PAK/PAT/OAuth is a separate subsystem and a follow-up.
- DCR (RFC 7591) deferred. CIMD covers the MCP clients we target; DCR is a fast-follow only if we hit a DCR-only client. Avoids building two registration paths.
- Figma plugin deferred. Its iframe can't do a redirect (null origin); the right answer is the device authorization grant (RFC 8628), and a large Figma refactor is landing soon. Postponed to a later version to avoid rework; rationale captured so it isn't re-litigated.
- Device grant deferred, unless a headless/SSH/container CLI case forces it earlier.
- Not removing PAK/PAT. They stay for CI, scripts, and machine-to-machine. Additive.
- Not the `client_credentials` grant. PAK already covers headless machine-to-machine.
- Not a general-purpose IdP. No "Login with Tolgee" for third parties, and no change to Tolgee's SSO-client role.
- No Redis-backed system of record for grants. Postgres stays authoritative.
- Org-level operations out of v1 unless the scope model lands cleanly.

---

### 7. Success Criteria

- An external community contributor authorizes the browser extension on a public project via redirect and consent, and suggests a translation, with no key pasted. This is the headline.
- The extension logs in via redirect and consent for members too; no manual key entry.
- The CLI logs in via the browser (open, approve, done); `--api-key` still works for CI.
- A real MCP client (Claude Desktop, Cursor) connects to `/mcp/developer` via redirect and consent, nothing in plaintext JSON.
- An unknown third-party MCP client authorizes via CIMD (URL-form `client_id`, no pasted key, no pre-registration on our side), and a client presenting a `client_id` whose metadata doesn't list its `redirect_uri` is rejected.
- Issued tokens carry exactly the consented scope, and effective access is the token intersected with live permissions evaluated per request (losing a project or leaving the org revokes immediately).
- Audience binding holds: an MCP token is rejected by the REST API and vice versa.
- A user can see connected apps and revoke one without affecting others.
- Works on a fresh self-hosted instance with no config, including persisted signing keys (restarts/replicas don't log everyone out).
- PAK/PAT auth continues to work unchanged.

### Scopes
**MUST:**
- [ ] OAuth server core — Click connect in an app, log in in the browser, and get a short-lived token instead of a pasted key. Demoable with curl/Postman against the endpoint that already accepts Bearer.
- [ ] Scope model & token shape — The token carries exactly what was approved: which actions, on which project. ("All projects" sentinel exists but is for owner clients, not community.)
- [ ] Live permission enforcement — The moment someone loses access to a project, their token stops working there.
- [ ] Audience binding — A token issued for one thing can't be used against another (foreign-audience token is rejected).
- [ ] Consent screen — When an app asks to connect, the person sees a plain summary of what it will be able to do and clicks Allow or Deny. Project is pre-filled from context, not picked.
- [ ] Connected apps & revocation — See the apps you've connected and disconnect any of them for real.
- [ ] Stay signed in — Sign in once and stay signed in; the token renews silently.

- [ ] SDK — OAuth prep (@tolgee/web) — The SDK can speak OAuth, so any app or plugin on it can log in without a key. (Project comes from page config, not decoded from the key.)
- [ ] Chrome plugin — path prep — The plugin's sign-in path is wired and verified end to end. (The page's known project survives the token swap.)

- [ ] MCP client (+ CIMD) — An AI tool (Claude, VS Code, Cursor) connects by logging in, with no key in a config file, and works across your projects.
- [ ] Tolgee CLI — Sign in with one command via the browser and the CLI just works; key option stays for CI.

**OPTIONAL:**

- [ ] Figma plugin

Anty0:
Consolidates the OAuth 2.1 authorization-server chain (#3849 → #3879 → #3880 → #3885) into one branch with a
reviewable history. Supersedes all four, which are now closed. Tracking issue: #3826.

## Why this branch exists

The chain reached its final shape through several reworks — a first implementation on Spring Authorization Server,
a review pass, a replacement of the library with an in-house server, and a release-readiness pass. The commits
record that journey, which is the wrong thing to review. These 21 commits record the destination instead: each one
makes a single incremental change, compiles and lints on its own, and never overrides work done in an earlier one.

## The commits

Groundwork that is independent of OAuth and stands on its own merit:

1. `test:` query permission scope checkboxes through `gcyAdvanced`
2. `fix:` evict the cached user when its tokens are invalidated
3. `refactor:` read the configured base URLs through providers
4. `fix:` stop a project API key inheriting the account's elevations
5. `feat:` gate the task-assignee elevation behind `tasks.assigned-access` — **breaking, see below**

The authorization server, bottom-up:

6. protocol primitives · 7. grant entity + schema · 8. client registry from configuration ·
9. issuer resolution · 10. authorization service · 11. revocation with the user's other credentials ·
12. authorization-server endpoints · 13. consent-flow API

The credential integration:

14. authenticate OAuth2 access tokens as scoped credentials · 15. RFC 6750 bearer challenge ·
16. cleanup of abandoned and expired grants

Then the client-visible surface: 17. one-origin dev frontend · 18. consent screen · 19. e2e coverage · 20. docs.

21. `chore:` skip the email build when Schema Check starts the backend — unrelated CI fix, appended last so it can be
    dropped or split off without touching the 20 above.

## BREAKING: granular permissions and project API keys lose the task-assignee elevation

`Scope.TASKS_ASSIGNED_ACCESS` gates the task-assignee elevation that used to be unconditional. Role presets expand
at runtime and pick it up, but granular permissions and project API keys store their scope list literally, so on
upgrade they no longer carry it: a member or key that could view and edit a task purely because it was assigned to
them now needs `tasks.assigned-access` granted explicitly.

There is deliberately no backfill. Restoring the bypass that way means granting a brand new scope to every
permission and API key that holds any scope at all — a lot of grants widened at once, and not distinguishable
afterwards from scopes an admin chose. Losing an elevation fails closed; granting one does not.

## An OAuth grant is a project API key that can hold several projects

The design rule for this PR is parity: a grant is treated as a PAK with a project set, not special-cased. Three
consequences of that parity are worth stating outright, because they read differently for a token handed to a
third party than for a key a user minted for their own tooling. **All three apply to project API keys today in
exactly the same way**, so each fix belongs to both credentials at once — and therefore to its own PR, not this
one:

- **Account-level endpoints apply no scope narrowing to any API credential.** `GET /v2/user` returns the account's
  email and server role, and `GET /v2/notifications` returns notifications across every project, to a token the
  consent screen presented as "translations.view on project X". Neither endpoint resolves a project, so neither the
  token's scopes nor its project set is consulted.
- **`@IsGlobalRoute` skips every project narrowing.** `AbstractAuthorizationInterceptor.preHandle` returns before
  `preHandleInternal` on a global route, so `coversProject` and the scope intersection are never asked. No
  `@AllowApiAccess @IsGlobalRoute` handler reads project-shaped data today.
- **A project a scoped credential does not cover is answered 403, not 404.** That distinguishes "exists" from "does
  not exist" for any project id the holder cares to probe. It is what a PAK has always answered; this branch gives
  an OAuth token the same answer rather than the 404 it used to get.

`docs/oauth/README.md` carries these plus the known V1 gaps — most notably that the grant stores a project *set*
while consent can only bind one project or all of them. That split is the V1 decision, not an oversight: the
storage and permission checks are the expensive parts to change later, the picker is not.

## Review

This branch went through 8 full review rounds raising 240 findings before this push. Most were fixed; the rest
were either defended and accepted or deliberately deferred with the reason recorded. Fixes were folded into the
commit each finding belonged to, so the history stayed reviewable commit by commit rather than growing a tail of
patches.

One test defect deserves its own note, because it was green for the wrong reason. `testMyTasks`
asserted that a user without task scopes sees an **error snackbar** on a refused task. The refusal is
real — the server answers `operation_not_permitted [TASKS_VIEW]` — but the app *navigates the user
away* rather than toasting, so that assertion only ever passed when some unrelated error toast
happened to be on screen. It now asserts that the task detail never renders. Confirmed by local
reproduction: all 18 permission specs pass (38/38), where the snackbar form failed 14 of them.

Two of the late findings were real defects worth naming:

- **`deleteExpiredBefore` compared the wrong expiry.** It took the first expiry that was set, justified as
  "by construction". The ordering actually comes from operator-configured durations, so a short
  `refresh-token-validity-days` against a long `access-token-validity-minutes` could delete a grant whose access
  token was still live. It now compares the latest of the three.
- **A cross-repo compile break.** Commit 3 renames `FrontendUrlProvider.url`, which breaks the billing repo.

## Approving a consent needs a super token

`POST /v2/oauth2/consent` carries `@RequiresSuperAuthentication`, matching `ApiKeyController.create` and
`PatController.create`. An OAuth grant is one more long-lived credential and its refresh token outlives the session
that approved it, so a lifted JWT alone must not be able to leave one behind. Signing in already yields a super token
(1 hour), and the webapp raises the dialog and retries on its own — so in practice the user is asked at most once per
hour. Opening and reading the consent screen are not gated; only the approval is.

## Companion branch

tolgee/billing#311 (`jirikuchynka/oauth2-consolidated`) follows the `FrontendUrlProvider` rename (2 lines). **It has
to land with this PR** — billing will not compile against the renamed API otherwise, and billing CI resolves the
platform branch by matching name.

## Verification

Rebased onto current `main`. Two conflicts came from `main` moving under the bearer-challenge commit — the
`ExceptionHandlers` imports and constructor, and its `AuthTest` call site — plus `ExceptionHandlersAsyncCapacityTest`,
a test `main` added that constructs `ExceptionHandlers` directly. All three were resolved inside the commit that
introduces the parameter, so no later commit patches an earlier one. `apiSchema.generated.ts` auto-merged; it was
regenerated afterwards and came back byte-identical.

- **Every commit is independently clean**: `ktlintCheck` and `compileTestKotlin` at each one, run with
  `--no-build-cache --rerun-tasks` and judged on gradle's exit status.
- 280 backend tests green across `AuthTest`, `ExceptionHandlers*`, `io.tolgee.security.oauth2.*` and
  `io.tolgee.mcp.*`; `:data:test` green; `e2e` eslint and tsc green; `webapp` eslint green.
- `webapp` tsc cannot run in this environment — a local npm `min-release-age` guard refuses `@tginternal/editor`
  1.20.0 as too recently published and falls back to 1.19.0, which lacks the exports `library/` imports. Nothing in
  the repo is pinned, the affected files are ones this PR does not touch, and CI's Frontend static check covers it.
- Billing compiles against the renamed API (`:billing-app:compileKotlin`, `:billing-test:compileTestKotlin`).

## Summary by CodeRabbit

- **New Features**
  - Added OAuth 2.1 authorization flows with consent, authorization codes, refresh tokens, revocation, PKCE, and discovery metadata.
  - Added a consent screen for selecting permissions and projects.
  - Added OAuth bearer-token access for supported API and MCP requests.
  - Added the `tasks.assigned-access` permission for assigned task workflows.
  - Added protected-resource metadata for MCP integrations.

- **Bug Fixes**
  - Improved permission enforcement, project access checks, token invalidation, and streaming error responses.

- **Documentation**
  - Updated local OAuth and end-to-end setup instructions.


backend/data/src/main/kotlin/io/tolgee/security/oauth2/OAuth2AuthorizationService.kt
```diff
@@ -0,0 +1,396 @@
...
+    client: OAuth2Client,
+    redirectUri: String,
+    params: AuthorizeParams,
+    projectHint: String?,
+  ): OAuth2Grant {
+    val (scopes, challenge) = validateAuthorizeRequest(params)
+
+    val grant =
+      OAuth2Grant().apply {
+        userAccount = userAccountService.get(userId)
```
bdshadow, line 100: nitpicking: `entityManager.getReference` can help here to avoid a `select`

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/oauth2/OAuth2FlowController.kt
```diff
@@ -0,0 +1,159 @@
...
+  private val authenticationFacade: AuthenticationFacade,
+  private val clientRegistry: OAuth2ClientRegistry,
+  private val authorizationService: OAuth2AuthorizationService,
+  private val projectService: ProjectService,
+  private val securityService: SecurityService,
+  private val issuerResolver: OAuth2IssuerResolver,
+) : IController {
+  @PostMapping("/authorize")
+  @Operation(summary = "Record the pending grant the consent screen will act on")
+  fun authorize(
```
bdshadow, line 58: `authorize` has an unconditional INSERT of an oauth2_grant and it's not rate limited, only a global rate: 400/min (~576k rows/day/user)

ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/task/TaskAssignedAccessScopeTest.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+
+/**
+ * The task-assignee elevation is gated on `Scope.TASKS_ASSIGNED_ACCESS`. These tests are about the granular-permission
+ * population: role presets grant the scope by expansion, hand-picked scope sets only carry it if it was granted.
+ */
+class TaskAssignedAccessScopeTest : ProjectAuthControllerTest("/v2/projects/") {
```
bdshadow, line 23: no test covers a PAK whose scope list lacks TASKS_ASSIGNED_ACCESS - the variant that breaks every pre-existing key without a new scope

backend/data/src/main/kotlin/io/tolgee/model/enums/Scope.kt
```diff
@@ -100,6 +101,9 @@ enum class Scope(
             ),
           ),
           HierarchyItem(TASKS_EDIT, listOf(tasksView)),
+          // Deliberately standalone, and deliberately NOT under ALL_VIEW: ALL_VIEW's expansion is `readOnlyScopes`,
+          // which grants server supporters read access, and this scope unlocks task-gated writes.
```
Anty0: Useless comment

backend/data/src/main/kotlin/io/tolgee/service/security/UserAccountService.kt
```diff
@@ -597,12 +603,15 @@ class UserAccountService(
...
-    resetTokensValidNotBefore(userAccount)
+    revokeSessionsAndOAuthGrants(userAccount)
     return userAccountRepository.save(userAccount)
   }
 
-  private fun resetTokensValidNotBefore(userAccount: UserAccount) {
+  private fun revokeSessionsAndOAuthGrants(userAccount: UserAccount) {
     userAccount.tokensValidNotBefore = DateUtils.truncate(currentDateProvider.date, Calendar.SECOND)
+    // The tokensValidNotBefore cutoff is read from the USER_ACCOUNTS cache, which lags per-node without Redis, so the
+    // refresh grant could keep minting tokens on a stale replica; deleting the grants makes revocation topology-safe.
```
Anty0: I'd drop this comment. It is a bit misleading - we don't support running multi node without redis, but we still should call the `revokeAllForUser` no matter, just no comment.

backend/app/src/test/kotlin/io/tolgee/security/oauth2/OAuth2GrantCleanupTest.kt
```diff
@@ -0,0 +1,99 @@
...
+      newOAuth2Grant(testData.user, clientId = "cleanup-test-client").apply {
+        this.codeHash = codeHash
+        this.refreshTokenExpiresAt = refreshExpiresAt?.let { Date.from(it) }
+        this.accessTokenExpiresAt = accessExpiresAt?.let { Date.from(it) }
+        this.codeExpiresAt = codeExpiresAt?.let { Date.from(it) }
+        this.consentExpiresAt = consentExpiresAt?.let { Date.from(it) }
+      }
+    repository.save(grant)
+    return grant.id
+  }
```
Anty0: Maybe can be built in test data too? We could add harness for it and build it there. Then these tests just turn into `testData.addExpiredGrant()` and `testDataService.saveTestData(testData.root)`, then assertiong - test data created in test data class as they should be.

backend/data/src/main/kotlin/io/tolgee/security/oauth2/OAuth2GrantCleanup.kt
```diff
@@ -0,0 +1,44 @@
...
+@Component
+class OAuth2GrantCleanup(
+  private val authorizationService: OAuth2AuthorizationService,
+  private val properties: OAuth2ServerProperties,
+  private val currentDateProvider: CurrentDateProvider,
+) : Logging {
+  @Scheduled(
+    cron =
+      "\${tolgee.oauth2.grant-cleanup-cron:${OAuth2ServerProperties.DEFAULT_GRANT_CLEANUP_CRON}}",
+  )
```
Anty0, line 38: Will this be fine on multi node deployment? Probably yes, just all nodes will try to do it at the same time - kinda waste. Anything we can do about that?

backend/api/src/main/kotlin/io/tolgee/api/v2/controllers/oauth2/OAuth2FlowController.kt
Anty0: I'll leave it as is. We need to do pass over these in the future - there are other endpoints with similar problems. This table is automatically cleaned up, so anything there isn't really persistent.

backend/data/src/main/kotlin/io/tolgee/security/oauth2/OAuth2AuthorizationService.kt
Anty0: Technically, it also ensures the user is enabled - it uses `findActive` underneath. I'll leave it as is just to be sure - this function doesn't need to be optimized since it's only called once throughout the OAuth flow.

PR #3899, opened 2026-09-04
chore: shard ee-test and context-recreating tests, split standard tests into 4 shards

dkrizan:
## What changed

Backend test jobs

- `server-app:runStandardTests` now runs in 4 shards instead of 2.
- `server-app:runContextRecreatingTests` now runs in 2 shards.
- `ee-test:test` now runs in 3 shards. The module gets a `junit-platform.properties` so the existing `TestShardCondition` is picked up.
- The `SHARD_INDEX` / `SHARD_TOTAL` handling moved from `backend/app/build.gradle` to a root `setTestSharding` helper, next to `setTestRetry`, so any test task can opt in.

E2E jobs

- Specs are assigned to shards by measured duration instead of alphabetically by file count. Durations live in `e2e/spec-durations.json`; a spec missing from the file counts as 30 seconds. `E2E_TOTAL_JOBS=20 ./gradlew printE2eGroups` shows the groups.
- 20 E2E jobs instead of 15.

Tests that only passed because of the old order

- `security/login.cy.ts`: the third-party and SSO login contexts now wrap themselves in `setBypassSeatCountCheck`, like every other user-creating spec. They create a fresh user through the OAuth callback and hit the free seat limit whenever earlier specs in the shard leave generated users behind. It passed before only because the alphabetical order ran `seatLimit.cy.ts` right before it.
- `BranchRevisionsTest`: never enabled `Feature.BRANCHING` itself and passed only when an earlier class in the same Spring context had left it force-enabled. Fails on its own with `feature_not_enabled [BRANCHING]`. Now sets `forceEnabled` in setup like every sibling branching test, and resets it in `@AfterEach` so it does not leak onward in the same way.

## Why

PR runs took about 15 minutes. The backend build takes 3 minutes, then the slowest test job takes 10 to 12 minutes. That job was always one of `ee-test:test`, a `runStandardTests` shard, or E2E shard 4, 6 or 7, which held the slow `projects/*` specs.

Measured on the last runs of this branch:

| Job | Before | After |
|---|---|---|
| ee-test:test | 11m 23s | 5m 46s to 6m 50s |
| runStandardTests | 10m 27s / 10m 34s | 5m 57s to 7m 03s |
| runContextRecreatingTests | 9m 01s | 5m 08s / 6m 00s |
| Slowest E2E shard | 11m 58s | 8m 04s |
| Whole run | ~15 min | ~11 min |

## Follow-ups, not in this PR

- `Editor.tsx` and `EditorJson.tsx` drop typed characters the same way the prompt editor did before #3892. With the new order `translations/softDeleteKeys.cy.ts` runs first after boot and created `ky1` instead of `key1` in 2 of 4 runs. The `isExternalValue` guard from #3892 should be applied there too.
- The export format matrix (`testExportFormats`, 48 combinations) runs three times: once on the export page and twice in content delivery, about 4.5 minutes of spec time. With balanced shards this no longer moves the wall time, so it is left as is.
- `e2e/spec-durations.json` is a snapshot. Refresh it when the groups drift apart; the README in `e2e/` says how.

## Summary by CodeRabbit

* **CI & Testing**
  * Improved test distribution across parallel CI jobs for faster, more balanced execution.
  * Increased end-to-end test parallelization from 15 to 20 jobs.
  * Standardized test sharding and improved test environment setup.
  * Improved end-to-end job balancing using recorded test durations.

* **Documentation**
  * Added guidance for end-to-end test sharding, local execution, port configuration, and duration data maintenance.

* **Test Reliability**
  * Updated security and enterprise test setup to ensure required feature and environment conditions are applied consistently.


ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/branching/BranchRevisionsTest.kt
```diff
@@ -25,13 +28,22 @@ class BranchRevisionsTest : ProjectAuthControllerTest("/v2/projects/") {
...
     testDataService.saveTestData(testData.root)
     userAccount = testData.user
     this.projectSupplier = { testData.project }
     currentDateProvider.forcedDate = currentDateProvider.date
+    enabledFeaturesProvider.forceEnabled = setOf(Feature.BRANCHING)
+  }
+
+  @AfterEach
+  fun cleanup() {
+    enabledFeaturesProvider.forceEnabled = null
```
bdshadow, line 46: since it's fixed here, i think it's a good pr to fix it in other places, e.g. `BranchControllerMergingTest`, `AllKeysControllerBranchingTest` and `TaskControllerTest`

.github/workflows/test.yml
bdshadow: probably would be good to add shard index here too

bdshadow (commented):
2 things:
1. nightly `reportIntermittentTests` workflow is not updated. Shouldn't it be sharded too?
2. What do you think about adding a new ci for updating the `spec-durations.json`? For example, when pr is merged to main and main build succeeds, launch an async update job, which will just collect the timings and update the file

dkrizan:
> 2 things:
> 
>     1. nightly `reportIntermittentTests` workflow is not updated. Shouldn't it be sharded too?
> 
>     2. What do you think about adding a new ci for updating the `spec-durations.json`? For example, when pr is merged to main and main build succeeds, launch an async update job, which will just collect the timings and update the file

1. I would leave that one out of this PR. It is a nightly job with its own matrix and nobody waits on it, so sharding it does not buy much. If we want it consistent I would do it separately.

2.  I think it would create noise. The timings differ a bit on every run, so a job committing the file after each merge to main would produce a commit on almost every merge and conflicts for open PRs touching `e2e/`. Maybe a manual `workflow_dispatch` that uploads the regenerated file as an artifact, without committing, could be a middle ground later.

ee/backend/tests/src/test/kotlin/io/tolgee/ee/api/v2/controllers/branching/BranchRevisionsTest.kt
dkrizan: I think fixing it in three classes would not really help. In `ee-test` 83 classes set `forceEnabled`, 49 reset it by hand and 34 do not, so the leak stays whatever we patch here. I would rather do it properly in a separate PR: one `AfterEachCallback` in `ee-test` that resets the flag after every test, and drop the 49 manual resets. That will probably surface a few more tests that only pass because of a leaked flag, so it does not belong in a CI PR.

bdshadow: agree

bdshadow (dismissed):
@dkrizan ok

1. I would do it the same way everywhere. Less duplications and the logic is the same everywhere - easier to investigate and catch a problem sooner
2. ok, agree about the noise. However, the problem is that ther is already a missing oauth test, which run longer than 30 seconds. Artifact is nice solution. Simpler can be to create a branch, where it will be commited and taken from (it's dirtier than an artifact, but i used such strategy in some cases in the past, it worked)

approve

PR #3900, opened 2026-09-05
fix: close websocket authentication holes and authenticate OAuth2 tokens on the socket

Anty0:
Security follow-up to #3893 (OAuth 2.1 authorization server), extended into a cleanup of the websocket authentication path. Nine commits, each reviewable on its own.

Two of them are not about websockets at all — they fix pre-existing problems this work exposed. They are first in the chain, so the websocket commits build on a correct base.

## Two authentication holes

A credential the STOMP resolver had **refused** could still reach a topic, both reachable with an OAuth token:

- `isUserSubscribeAllowed` tested `is ApiKeyDto`, so only a project API key was kept off `/users/{id}/…`. Any other scoped credential — an OAuth grant — reached the account's own notifications.
- The subscribe decision read the principal off the frame, and Spring falls back to the HTTP handshake principal there. The handshake is an ordinary servlet request that the filter chain authenticates, so a refused credential rode in anyway — reliably over SockJS's HTTP transports, which take their principal straight off the handshake request.

The resolver's CONNECT verdict is now stored in the session attributes and is the only thing the subscribe decision reads.

## Three more found while reviewing

- **Event forgery.** Client `SEND` frames were relayed to the simple broker, which fanned them out to every subscriber of the destination. Reproduced against a running server. `MESSAGE` frames were the same hole under a different command name — both carry `SimpMessageType.MESSAGE`, and nothing on Spring's inbound path rejects a client for sending a command that is nominally server-to-client. Both are now dropped.
- **Total websocket outage under `tolgee.authentication.enabled: false`** — the default for a direct Java run.
- **That fix was unreachable from the browser:** the client sent `Bearer undefined` when it had no token — a bad credential rather than no credential.

## Two pre-existing fixes this work exposed

- **A deleted account could still be the initial user.** `findInitialUser()` matched `isInitialUser = true` alone, and `softDeleteUser` left that flag set while stamping `deletedAt`. So a deleted initial user stayed the identity that `tolgee.authentication.enabled: false` runs as, and the one the startup runner finds instead of creating a replacement. The query now excludes deleted accounts and soft delete clears the flag.
- **`checkProjectPermission`'s `credential` parameter was misleading.** It looked like an off-request-thread caller could supply the credential directly. It could not: the check delegates to `checkProjectPermissionNoApiKey`, which reads the credential from the security context regardless — and with no context there it takes the admin and supporter bypasses. So passing the credential *without* setting up a context gave a **more permissive** answer than doing it properly. The parameter is gone; `withSecurityContext` is what makes the check correct, and it now lives next to `TolgeeAuthentication` so both callers share one copy.
  `QaCheckPreviewWebSocketHandler` gained the wrapper it was missing. It accepts a JWT alone today, which is never a scoped credential, so nothing changes — but the day it accepts a project API key or an OAuth token, the check would have silently skipped scope narrowing.

## OAuth parity

An OAuth grant is treated as a project API key that may hold several projects, not special-cased against one: same `keys.view` requirement on a project topic, same refusal on a user topic, same scope narrowing. The consequences of that parity are listed in `docs/oauth/README.md` rather than designed around.

## Webapp

An unresolvable credential meant a full SockJS handshake, CONNECT and SUBSCRIBE every three seconds for as long as the tab stayed open, with no user-visible signal, against a CONNECT that runs under no IP auth rate limit. `onError` now reports whether the ERROR frame carried the server's `Unauthenticated` header, and the client deactivates only for that one — a server-side failure still reconnects.

## Also

- `WebsocketSubscribeAuthorizer` holds the subscribe policy; `WebSocketConfig` is wiring (~215 lines to ~105).
- `DisabledAuthenticationResolver` gives the disabled-authentication identity one owner instead of two. It is looked up per call rather than cached for the JVM's life: that costs a query per credential-less request while authentication is disabled — a dev, demo and e2e mode — and it buys correctness, since the old cache had no eviction.
- `WebsocketEventType` builds the destination strings, replacing five hand-built literals. Typing that surfaced a nullable `BatchJobDto.projectId` publishing to `/projects/null/batch-job-progress`.
- `docs/websocket/README.md` documents the transport-level model for every credential type.

## Not fixed, deliberately

- Every project topic requires exactly `keys.view` whatever the topic carries, so a grant consented to as `keys.view` alone receives `translation-data-modified` payloads whose HTTP readers require `translations.view`. Narrowing it per topic would break existing project API keys; it is the PAK's rule applied unchanged.
- A live subscription is never re-authorized — revocation, expiry and permission changes do not stop an open stream. Identical to a project API key today.
- Neither CONNECT nor SUBSCRIBE runs under the IP auth rate limit, and the websocket applies no SSO liveness check.

## Follow-ups worth a separate PR

Both surfaced while fixing the deleted-initial-user bug, and neither is made worse by this PR:

- `isDeletable` (`accountType != MANAGED && !isInitialUser`) is only surfaced to the UI, not enforced server-side, so nothing actually stops the initial user being deleted.
- Once it is deleted, the next restart recreates an **admin** account under the configured initial username with the password from the on-disk password file.

## Testing

45 websocket tests, including a raw STOMP client that writes frame bytes directly so a test can forge a `MESSAGE` frame no client library will emit; 8 webapp unit tests; 3 repository tests covering the deleted-account guards. Each fix was falsified by reverting it and confirming the matching test — and only it — fails.

## Summary by CodeRabbit

- **New Features**
  - WebSocket connections now support OAuth bearer-token authentication through STOMP.
  - Project and user subscriptions are authorized according to credentials and permissions.
  - Unauthenticated users can connect where supported, while unauthorized subscriptions are refused.
  - Invalid client message frames are denied to improve connection security.
  - WebSocket destinations for notifications, translation updates, batch jobs, and QA events are now handled consistently.
  - Authentication-disabled environments continue to support configured default-user access.

- **Documentation**
  - Added guidance for WebSocket authentication, subscription rules, refusal behavior, reconnects, and supported endpoints.

### DIFF
--- a/backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
@@ -8,6 +8,7 @@ import io.tolgee.security.PAT_PREFIX
 import io.tolgee.security.authentication.DisabledAuthenticationResolver
 import io.tolgee.security.authentication.JwtService
 import io.tolgee.security.authentication.TolgeeAuthentication
+import io.tolgee.security.oauth2.OAuth2AccessTokenResolver
 import io.tolgee.service.security.ApiKeyService
 import io.tolgee.service.security.PatService
 import io.tolgee.service.security.UserAccountService
@@ -20,6 +21,7 @@ import org.springframework.stereotype.Component
 @Component
 class WebsocketAuthenticationResolver(
   @Lazy private val jwtService: JwtService,
+  @Lazy private val oauth2AccessTokenResolver: OAuth2AccessTokenResolver,
   @Lazy private val apiKeyService: ApiKeyService,
   @Lazy private val patService: PatService,
   @Lazy private val disabledAuthenticationResolver: DisabledAuthenticationResolver,
@@ -37,7 +39,9 @@ class WebsocketAuthenticationResolver(
 
     val bearer = extractBearer(authorizationHeader)
     if (bearer != null) {
-      return attempt("Bearer token") { jwtService.validateToken(bearer) }
+      return attempt("Bearer token") {
+        oauth2AccessTokenResolver.tryResolve(bearer) ?: jwtService.validateToken(bearer)
+      }
     }
 
     if (!xApiKeyHeader.isNullOrBlank()) {
--- a/backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
@@ -5,6 +5,7 @@ import io.tolgee.constants.Message
 import io.tolgee.dtos.cacheable.UserAccountDto
 import io.tolgee.exceptions.AuthenticationException
 import io.tolgee.security.PAT_PREFIX
+import io.tolgee.security.authentication.DisabledAuthenticationResolver
 import io.tolgee.security.authentication.JwtService
 import io.tolgee.security.authentication.TolgeeAuthentication
 import io.tolgee.service.security.ApiKeyService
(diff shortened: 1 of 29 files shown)


Anty0:
@coderabbitai review

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
```diff
@@ -133,7 +133,9 @@ interface UserAccountRepository : JpaRepository<UserAccount, Long> {
   @Query("from UserAccount ua where lower(ua.username) = lower(:username) and ua.deletedAt is null")
   fun findActiveOrDisabled(username: String): UserAccount?
 
-  @Query("from UserAccount ua left join fetch ua.emailVerification where ua.isInitialUser = true")
+  @Query(
+    "from UserAccount ua left join fetch ua.emailVerification where ua.isInitialUser = true and ua.deletedAt is null",
```
bdshadow, line 137: these changes in the `UserAccountRepository` now make it so, that a soft-deleted initial user is created again ono the next app start. I don't think it's an exprected behavior. 
Just as an idea: add `isDeletable (accountType != MANAGED && !isInitialUser)` and if user is not `isDeletable`, don't allow to delete it. Anyway there is always a possibility to not create it by enabling `disableInitialUserCreation`. wdyt?

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
```diff
@@ -19,71 +21,51 @@ import org.springframework.stereotype.Component
...
     }
 
     return null
   }
 
+  private fun attempt(
+    credentialKind: String,
+    resolve: () -> TolgeeAuthentication?,
+  ): TolgeeAuthentication? =
+    runCatching(resolve)
```
bdshadow, line 65: wouldn't it be better to catch only `AuthenticationException` here?
it is also mentioned in the docs/websocket/README.md on line 72

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
Anty0: Yee, I considered that option too, but... Once you are deleting you initial user, I have no idea what you expect to happen xD
So I just went the "least resistance" route and if you really delete your initial user, we just create new one on the next start, but it probably isn't the best way to go about it.

I can add enforcement for the `isDeletable`, but it doesn't really allow me to remove any other changes, so I vote "it isn't worth it" and I'd leave it as is.

bdshadow: Jirko, what dou you mean by "it doesn't really allow me to remove any other changes"?

Anty0: I still need to filter out the deleted initial user if I don't want potential existing instance with deleted initial user to stop working, etc. I can instead write a migration or something, but then it would be a bit less resistant to invalid data. Hard to say what is the best approach.

backend/api/src/main/kotlin/io/tolgee/websocket/WebsocketAuthenticationResolver.kt
Anty0: applied ^^

backend/data/src/main/kotlin/io/tolgee/repository/UserAccountRepository.kt
Anty0: Added the `isDeletable` handling on BE side. Let me know if it looks good ^^

bdshadow (approved):

PR #3904, opened 2026-09-08
fix: preserve plural number placeholder tag during machine translation

braindeaf:
## Observed bug

When create a new set of keys for a Rails application, we push `tolgee push` and we get an auto-translation of values for all the languages we support. 

```yaml
    with_attachments:
      one: "%{count} file uploaded:"
      other: "%{count} files uploaded:"
```

However it is not uncommon for the `%{count}` within the auto-translated text to be `#` or `#{count}` or even `%'{'count'}'` which indicates an issue with preserving these interpolation placeholders.

## Summary

When auto-translating a plural key through a machine translation provider that doesn't support ICU plurals natively (Google, DeepL, Azure, AWS — everything except the LLM/prompt-based provider), each plural form is translated in a separate request. To protect the ICU `#` (replace-number) placeholder during that request, `PluralTranslationUtil` substitutes a real example number wrapped in a `<x id="tolgee-number">` tag before sending the text to the provider, then strips the tag back to `#` afterward.

The bug: none of the provider API calls actually told the engine to preserve that tag. Each provider was free to translate through it, reposition it, or drop it entirely — so restoration of `#` was unreliable. In practice this surfaces as things like Ruby/Rails `%{count}` coming back as `#{count}`, a bare `#`, or an escaped literal placeholder after re-export, essentially every time a plural key went through auto-translate with one of these providers.

## Fix

- Added a `containsNumberTag` flag that flows from `PluralTranslationUtil` through `TranslationParams` / `ProviderTranslateParams` to each provider.
- Each provider now opts into its own documented tag-preserving/HTML mode when the flag is set:
  - **DeepL**: `tag_handling=xml` + `ignore_tags=x`
  - **Azure**: `textType=html`
  - **Google**: `format("html")` instead of the hardcoded `"text"`
  - **AWS**: `contentType("text/html")`, which is required for its existing `translate="no"` span wrapping (`HtmlNoTranslatePlaceholderProtector`) to actually take effect
- Baidu is unchanged — its basic translate API doesn't document a tag-exclusion mechanism.
- Updated the two existing tests whose mocks needed a matcher for the new parameter (`DeeplTranslationProviderTest`, `TranslationSuggestionControllerMtTest`).

## Test plan

- [ ] `./gradlew :data:test --tests "*Deepl*" --tests "*PluralTranslationUtil*"`
- [ ] `./gradlew :app:test --tests "*TranslationSuggestionControllerMtTest*"`
- [ ] Manual: auto-translate a plural key (e.g. Rails-style `one`/`other` with `%{count}`) through DeepL, Google, Azure, and AWS, and confirm the exported translation keeps `%{count}` (or the target format's equivalent) intact for both forms.

Note: I wasn't able to compile/run the test suite locally (this environment only had a Java 26 toolchain available; the project pins Java 25 and toolchain auto-provisioning isn't configured here), so please treat CI as the first real build of this change.

https://claude.ai/code/session_017zkLqfP1ezvSqve8C86a6a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved machine translation handling for pluralized text containing protected number placeholders.
  * Translation providers now preserve these placeholders across supported services, reducing the risk of corrupted or altered plural formatting.
  * Provider-specific formatting is automatically selected when protected tags are present, while standard text handling remains unchanged.
  * Improved handling of markup and HTML escaping during translation so translated plural text is restored correctly.

### DIFF
--- a/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translationSuggestionController/TranslationSuggestionControllerMtTest.kt
+++ b/backend/app/src/test/kotlin/io/tolgee/api/v2/controllers/translationSuggestionController/TranslationSuggestionControllerMtTest.kt
@@ -152,11 +152,13 @@ class TranslationSuggestionControllerMtTest : ProjectAuthControllerTest("/v2/pro
         any(),
         any(),
         anyOrNull(),
+        any(),
       ),
     ).thenReturn("Translated with DeepL")
 
     whenever(
       azureCognitiveApiService.translate(
+        any(),
         any(),
         any(),
         any(),
--- a/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/MtServiceManager.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/MtServiceManager.kt
@@ -98,6 +98,7 @@ class MtServiceManager(
         isBatch = params.isBatch,
         pluralFormExamples = params.pluralFormExamples,
         pluralForms = params.pluralForms,
+        containsNumberTag = params.containsNumberTag,
       )
 
     return translateParams
--- a/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/TranslationParams.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/machineTranslation/TranslationParams.kt
@@ -15,4 +15,10 @@ data class TranslationParams(
   val isBatch: Boolean,
   var pluralForms: Map<String, String>? = null,
   val pluralFormExamples: Map<String, String>? = null,
+  /**
+   * True when [text] contains the `<x id="tolgee-number">` tag protecting the ICU `#` plural
+   * placeholder (see [io.tolgee.service.machineTranslation.PluralTranslationUtil]). Providers use this
+   * to switch on their own tag-preserving/HTML mode so the engine doesn't mangle the tag.
+   */
+  val containsNumberTag: Boolean = false,
 )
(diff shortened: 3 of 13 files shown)


dkrizan:
Thanks for this PR, good job finding the number tag problem and covering it with tests.

I pushed a small commit on top of your branch (56e478d), I hope that is fine since the PR allows maintainer edits.

The commit escapes the form text in `PluralTranslationUtil` just before the number tag is inserted, and unescapes it just after the tag is stripped back to `#`. Both steps only run when the tag is present. No provider code changes, unit tests added.

I tested the branch with real Google, DeepL, Azure and AWS keys. In HTML/XML mode the providers return `<`, `>`, `&` and quotes as entities, and nothing decoded them on the way back.

- `# < 5 items` was stored and exported as `# &lt; 5 items` from Google and DeepL.
- Azure treated the bare `<` as broken markup and returned the English text untranslated.
- Only plurals with `#` are affected, the tag is what turns markup mode on.

Three smaller things from the review that I think are still open, what do you think?

- `DeeplApiService` adds `tag_handling` after the user's `optionalParameters`, so a self-hoster who set `tag_handling` themselves would send it twice. Maybe `set` instead of `add`?
- `containsNumberTag` in the cache key changes the key for every translation, not only plurals, so the whole MT cache is dropped on deploy. The flag is already part of `text`, so I think it could be left out of the key.
- The description still mentions an AWS fix, but 51828e4 removed it. Could you update it?

I also noticed one limitation. A source string that literally contains `&amp;` comes back as `&` from Google, Azure and AWS. I think that is inherent to sending the text as markup, not something to fix here.

PR #3908, opened 2026-09-09
fix: reject requests on rate limit bucket contention instead of queueing threads

bdshadow:
## Problem

All rate limit verdicts for a bucket — including "you are rate limited, 429" — are serialized behind a distributed Redisson lock, and `RateLimitService.consumeBucketUnless` blocks on it **indefinitely** (`lock.lock()`).

When many requests share one bucket (e.g. a single API key used by a large fleet of shipped clients), this becomes a self-inflicted denial of service: a turn on the lock costs ~4 Redis round-trips (acquire, bucket get, bucket/strike put, unlock + pub/sub), and Redisson's unfair wake-all handoff degrades to ~25–50 ms per turn with many waiters — so one bucket can only serve a few dozen verdicts per second. Once arrival on the bucket exceeds that, the queue grows without bound (wait = queue position × turn time), clients time out and retry — refilling the queue faster than it drains — and the entire worker pool ends up parked on a single bucket lock. The node stops serving *all* tenants while still passing health checks on the management port.

Per-IP rate limiting (proxy or app-level) cannot catch this shape: each client stays far below any per-IP threshold; it is the aggregate on one *bucket* that kills the node. And the failure mode is concurrency × duration, not request rate — so the fix is to stop queueing worker threads, not to tune limits.

## Fix

`RateLimitService` no longer queues unboundedly on the bucket lock:

1. **Per-bucket concurrency cap (per node)** — an in-flight counter per bucket; above `tolgee.rate-limits.max-concurrent-per-bucket` (default 50) requests are rejected with 429 immediately, without touching the lock. Node-local by design: the protected resource (worker threads) is node-local, and the cap keeps working even when Redis itself is slow. Set to 0 to disable.
2. **Bounded lock wait** — the bucket lock is acquired via the new `LockingProvider.tryWithLocking(name, waitTime, fn)` with `tolgee.rate-limits.lock-wait-ms` (default 500 ms) instead of an infinite `lock.lock()`; timeout → 429. `tryWithLocking` is a default interface method; `RedissonLockingProvider` overrides it only to keep its `isHeldByCurrentThread` unlock guard.

Neither rejection records a strike (the bucket cannot be read or written safely without holding the lock); both use `retryAfter = lock-wait-ms`.

New metric: `tolgee.ratelimit.concurrency_rejections` counter, tagged `reason=concurrency_cap|lock_timeout`.

## Behavior change

A client sending more than `max-concurrent-per-bucket` truly simultaneous requests per node to one bucket now gets 429s for the excess instead of having them queue briefly. The default of 50 covers the largest known legitimate burst — apps with 40+ namespaces fetch them all in parallel on startup through one API key (one bucket), even if the whole burst lands on a single node — while still limiting a single hot bucket to ~25% of a node's worker pool (~200 threads). The cap only bites the flood shape described above.

## Testing

- New unit tests in `RateLimitServiceTest`: rejection within the lock-wait bound when the lock is held; cap rejection happens without waiting for the lock (timing-asserted against a 10 s lock wait); cap on one bucket does not affect other buckets; `max-concurrent-per-bucket = 0` restores queueing. Metric counters asserted throughout.
- Existing `RateLimitServiceTest`, `:security` rate limit tests, and `RateLimitsTest` / `RedisRateLimitsTest` integration suites pass unchanged.

### DIFF
--- a/backend/api/src/main/kotlin/io/tolgee/component/lockingProvider/RedissonLockingProvider.kt
+++ b/backend/api/src/main/kotlin/io/tolgee/component/lockingProvider/RedissonLockingProvider.kt
@@ -31,6 +31,24 @@ open class RedissonLockingProvider(
     }
   }
 
+  override fun <T> tryWithLocking(
+    name: String,
+    waitTime: Duration,
+    fn: () -> T,
+  ): T? {
+    val lock = this.getLock(name)
+    if (!lock.tryLock(waitTime.toMillis(), TimeUnit.MILLISECONDS)) {
+      return null
+    }
+    try {
+      return fn()
+    } finally {
+      if (lock.isHeldByCurrentThread) {
+        lock.unlock()
+      }
+    }
+  }
+
   override fun <T> withLockingIfFree(
     name: String,
     leaseTime: Duration,
--- a/backend/data/src/main/kotlin/io/tolgee/component/LockingProvider.kt
+++ b/backend/data/src/main/kotlin/io/tolgee/component/LockingProvider.kt
@@ -1,6 +1,7 @@
 package io.tolgee.component
 
 import java.time.Duration
+import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.Lock
 
 interface LockingProvider {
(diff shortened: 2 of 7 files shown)


backend/data/src/main/kotlin/io/tolgee/component/LockingProvider.kt
```diff
@@ -11,6 +12,30 @@ interface LockingProvider {
...
+    if (!lock.tryLock(waitTime.toMillis(), TimeUnit.MILLISECONDS)) {
+      return null
+    }
+    try {
+      return fn()
+    } finally {
+      lock.unlock()
+    }
+  }
+
```
Anty0, line 38: Why same function twice?

bdshadow: @Anty0  which one?
`tryLock` is only in this function

backend/api/src/main/kotlin/io/tolgee/component/lockingProvider/RedissonLockingProvider.kt
```diff
@@ -31,6 +31,24 @@ open class RedissonLockingProvider(
...
+    }
+    try {
+      return fn()
+    } finally {
+      if (lock.isHeldByCurrentThread) {
+        lock.unlock()
+      }
+    }
+  }
+
```
Anty0, line 51: Thought this is basically the same function.

PR #3917, opened 2026-09-14
feat: add copy button to translation cells

Anty0:
Requested by a customer: a way to copy a translation without having to open the editor.

Adds a copy button to the translation cell controls, next to edit and comments. It appears in **both the list and the table view**, on any cell that has a translation. Clicking it copies the value and does not open the editor.

#### List view — hovering the button:

pr-list-view

#### After clicking — the icon becomes a checkmark and the tooltip reads "Copied!":

pr-list-view-copied

#### Table view — same control, same behaviour:

pr-table-view

## Notes

- The copied value is the **raw stored text**, so a plural key yields its full ICU source (`{value, plural, one {…} other {…}}`) rather than a rendered variant. This is what the editor shows and what is most useful to paste elsewhere; there is an e2e case pinning it.
- No cell without a translation gets the button — there would be nothing to copy.
- A failed clipboard write does not show the "Copied!" confirmation.
- The confirmation lasts as long as the button does. The controls unmount shortly after the pointer leaves the row, which takes the copied state with it, so there is no separate timer.
- Uses the existing `clipboard_copy` / `copied_to_clipboard` keys.
- Deletes `webapp/src/component/common/ClipboardCopy.tsx`. It had **no call sites** anywhere in the repo, and keeping it would have left a second, dead implementation of the same widget.

#### One way to copy, everywhere

Both `InvitationItem` components used `navigator.clipboard.writeText`, while every other copy action in the webapp uses `copy-to-clipboard`. `navigator.clipboard` only exists in a **secure context**, so on a self-hosted instance served over plain HTTP those two "copy link" buttons threw instead of copying. Both now go through `copy-to-clipboard` like the rest.

They also showed the success snackbar regardless of the outcome; they now report a failed copy instead — new key `invite_user_invitation_copy_error`.

#### e2e fixtures built on the backend

`copyTranslation.cy.ts` gets its data from a `TestData` class and an E2E data controller instead of creating it from the test. `withViews.cy.ts` was migrated the same way, which made `forEachView`'s project deletion redundant — so that helper no longer deletes anything, and no longer needs a project id.

The fixture is named after its data (`FourTranslationsTestData` / `four-translations`) rather than the spec that moved first, since `create4Translations` still backs eight other specs that could migrate onto it.

#### One clipboard stub

`e2e/cypress/common/clipboard.ts` replaces the two ad-hoc stubs the invitation specs carried. `copy-to-clipboard` copies via `execCommand` and falls back to `window.prompt` outside the user gesture, so both are captured and a spec no longer has to know which path its component takes.

## Testing

- `translations/copyTranslation.cy.ts` — 3 cases through `forEachView`, so 6 tests across both views: copies and leaves the editor closed, copies a plural key's raw ICU, and shows no button on an untranslated cell.
- `organizations/organizationInvitations.cy.ts` — adds a case for the invitation list's own copy button, which had no coverage before the mechanism changed underneath it.
- `projects/invitation.cy.ts` — its two clipboard assertions were wrapped in `cy.wrap(() => …)`, which yields the function without calling it, so neither had ever run. They do now.

## Summary by CodeRabbit

- **New Features**
  - Added copy buttons to translation cells for translated, untranslated, and plural content.
  - Copied translations remain visibly confirmed after successful copying.
  - Improved copying of organization and project invitation links with confirmation feedback.

- **Bug Fixes**
  - Invitation links now copy more reliably across supported environments.

- **Tests**
  - Expanded end-to-end coverage for translation-cell and invitation-link clipboard workflows.


webapp/src/views/projects/translations/cell/CopyButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type Props = Omit<
+  React.ComponentProps<typeof ControlsButton>,
+  'onClick' | 'tooltip' | 'children'
+> & {
+  text: string;
+};
+
+export const CopyButton = ({ text, ...props }: Props) => {
+  const [copied, setCopied] = useState(false);
```
dkrizan, line 16: I think `copied` never goes back to false while the row stays hovered or focused. If I copy `en` and then `cs` in the same row, both cells show the checkmark. Maybe a short timeout to reset it, like the old `ClipboardCopy` had? Optional.

webapp/src/views/organizations/members/InvitationItem.tsx
```diff
@@ -68,11 +69,15 @@ export const InvitationItem: React.FC<React.PropsWithChildren<Props>> = ({
   };
 
   const handleGetLink = () => {
-    navigator.clipboard.writeText(
+    const success = copy(
       LINKS.ACCEPT_INVITATION.buildWithOrigin({
         [PARAMS.INVITATION_CODE]: invitation.code,
       })
     );
+    if (!success) {
```
dkrizan, line 77: When `copy()` returns false, I think `copy-to-clipboard` has already shown a `window.prompt` with the link, so the user sees the link and then an error toast right after. Same in the project `InvitationItem`. Not sure it is worth handling, just mentioning it. Optional.

dkrizan (approved):
LGTM 👍
Two small things below, both found by AI review and confirmed by hand. Optional, I leave it to you whether to fix them, both are low-severity rough edges.

webapp/src/views/projects/translations/cell/CopyButton.tsx
Anty0: Yup, it is on purpose. It actually looks fine in practice - it resets the second the buttons disappear. Since they show up on hover, they'll be reset by the time user returns back to the translation.
It would be problem for generic copy button, but fine in this case.

webapp/src/views/organizations/members/InvitationItem.tsx
Anty0: Addressed. ^^

webapp/src/views/projects/translations/cell/CopyButton.tsx
dkrizan: Sorry, I meant something else - it is within one row. Both cells are in the same row, so active is still true and the first button never unmounts: hover a row, copy en, then copy cs without leaving the row, and both cells show the checkmark. Still minor, just wanted to be sure we mean the same thing.

PR #3921, opened 2026-09-17
fix: build the project dashboard billing link from the organization slug

Anty0:
## Problem

The "strings" tile on the project dashboard (`ProjectTotals.tsx`, `redirectToBilling`) built `LINKS.ORGANIZATION_BILLING` with `preferredOrganization.id` where the route expects the organization **slug**, so organization owners on a billing-enabled instance were sent to `/organizations/<id>/billing`, a URL that doesn't resolve.

## Fix

One line: pass `preferredOrganization?.slug` instead of `?.id`, matching every other `ORGANIZATION_BILLING` caller (`PlanLimitPopoverCloud`, `CriticalUsageCircle`, `OrganizationBillingRedirect`, `BaseOrganizationSettingsView`).

The click handler stays guarded by `canGoToBilling` (`billing.enabled` && current user is org `OWNER`), so a preferred organization always exists when it runs; the `|| ''` fallback is left as-is.

Link target is kept as `ORGANIZATION_BILLING` (same as today). For reference, `BillingMenuItem` and the trial announcement link to `ORGANIZATION_SUBSCRIPTIONS` instead; changing the destination is a separate product decision, not part of this bug.

Salvaged from closed PR #3816, which fixed the same bug inside a larger refactor; the helper introduced there is intentionally not ported.

## Verification

- `npm run eslint -- --fix && npm run tsc` clean.
- No automated test added: the tile is only clickable with `billing.enabled` and an `OWNER` role, which the platform Cypress suite never runs with (billing disabled), and no billing e2e spec visits the project dashboard. Covering it would mean a new cross-repo billing e2e spec, out of proportion for a one-line link fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the billing redirect to use the organization’s preferred URL slug, improving navigation to the correct billing page.
  - Added a safe fallback when no organization slug is available.


dkrizan:
Hmm, I’m not sure what “URL that doesn’t resolve” means. I tested the link in the app, and it works, but that doesn’t necessarily mean the link creation is correct. It’s not. Approving.

dkrizan (approved):

PR #3924, opened 2026-09-17
fix: keep client-supplied event data from overriding PostHog attribution

Anty0:
Salvaged from closed PR #3816 (the reporter-side item only; the `organizationMember` flag and `OrganizationRoleService` dependency are deliberately not ported).

## Problem

`PostHogBusinessEventReporter.captureWithPostHog` built the PostHog properties as `serverAttribution + utmData + data + $set`. `data` is the free-form map from `BusinessEventReportRequest` (`POST /v2/public/business-events/report`), so any client — authenticated or anonymous — could overwrite `$groups`, `organizationId`, `organizationName` and `glossaryId`, i.e. attribute its events to another organization. Only `$set` was protected by merge order. Client keys could also set PostHog-reserved properties such as `$set_once`, and a `$`-prefixed event name (e.g. `$create_alias`) was forwarded as-is.

## Fix

- Client-supplied maps (`utmData`, `data`) are merged first, then server attribution, then `$set`, so server values always win — including a server `null`, which `filterValueNotNull` then drops, so an unresolved attribution cannot be forged either.
- Client keys that start with `$`, that PostHog ingestion reads out of `properties` (`distinct_id`, `token`), or that name a server attribution property are dropped before the merge.
- Events whose name starts with `$` are refused before any DB lookups (debug log; the condition is client-triggerable).

Not covered on purpose: `sdkType`/`sdkVersion` in the client map. Those legitimately arrive through the same `data` map from `BusinessEventPublisher.getDataWithSdkInfo`, so they cannot be filtered here without dropping real SDK attribution.

## Tests

- `PostHogBusinessEventReporterTest` (`:data`): client data/utm cannot overwrite resolved attribution; cannot forge unresolved attribution; reserved keys dropped; reserved event name refused.
- `BusinessEventControllerTest`: end-to-end through the public endpoint — forged `organizationId`/`organizationName`/`$set_once` in `data` do not reach PostHog.

## Summary by CodeRabbit

* **Bug Fixes**
  * Protected organization attribution and identification data from being overridden by client-provided event details.
  * Blocked reserved PostHog properties and event names from being reported.
  * Prevented forged attribution data from being included in business event reports.


backend/data/src/main/kotlin/io/tolgee/component/reporting/PostHogBusinessEventReporter.kt
```diff
@@ -65,7 +70,12 @@ class PostHogBusinessEventReporter(
         "organizationId" to data.organizationId,
         "organizationName" to data.organizationName,
         "glossaryId" to data.glossaryId,
-      ) + (data.utmData ?: emptyMap()) + (data.data ?: emptyMap()) + setEntry
+      )
+    val clientSupplied =
+      ((data.utmData ?: emptyMap()) + (data.data ?: emptyMap()))
+        .filterKeys { !it.startsWith(RESERVED_PREFIX) && it !in RESERVED_KEYS && it !in serverAttribution.keys }
```
dkrizan, line 76: I think this filter also runs on events from internal callers, not only the public endpoint. Billing's `TelemetryController` sets `$ip` on `SELF_HOSTED_TELEMETRY_REPORT` so PostHog can geolocate the self-hosted instance, and that key gets dropped here - so all self-hosted telemetry would be located at our server instead. Billing tests don't run in this CI, so nothing would catch it.

Wouldn't it be better to filter only the data from `BusinessEventReportRequest` in `BusinessEventPublisher`, before the event is built?

